PR #155, opened 2022-03-07
Introduce evaluation delay for rule groups

codesome:
This is to support https://github.com/grafana/mimir/issues/1267

With this change, the `DefaultEvaluationDelay` function in the manager options will fetch the global or per-tenant evaluation delay and the time adjustments need not be handled in Mimir anymore.


pstibrany:
Is there any plan to upstream this change?

pstibrany:
> Haven't though about it yet. I don't see any harm in that. Maybe we could after we upstream the out-of-order support, with which this will be more relevant to have in Prometheus.

It would simplify our merge of upstream into `mimir-prometheus`. (If Prometheus community accepts the PR in the first place.)

Issue #147, opened 2022-02-11
Query path: Implement the OOOHead satisfying the blockreader interface


jesusvazquez:
One of the main lines of work torwards achieving support for out of order in prometheus is actually implementing the storage of the samples in the head as well as the ability to query them.

These are the milestones we need to achieve:
- [x] Extend the series in the Head to have a list of out of order chunks.
- [x] Implement a `RangeHead` for out of order samples called `OOORangeHead`
  - [x] IndexReader implementation https://github.com/grafana/mimir-prometheus/pull/149
  - [x] ChunkReader implementation at https://github.com/grafana/mimir-prometheus/pull/163
  - [x] Unit testing many query scenarios to make the implementation more bulletproof
- [x] Support for querying OOO data from the head. https://github.com/grafana/mimir-prometheus/pull/222

jesusvazquez:
Small update here, still working on the ChunkReader implementation, I expect to have something working between this week and the beginning of the next one.

jesusvazquez:
Replying to https://github.com/grafana/mimir-prometheus/issues/147#issuecomment-1098018961 to keep track of the status for each case:
- Case 1 is being tested here https://github.com/grafana/mimir-prometheus/blob/f3f6711bfe6acb7e126b5e167ece3cb3516f15e2/tsdb/ooo_head_read_test.go#L135-L155
  - The unit test tries every permutation of the existing chunks so the unit test has all the combinations: increasing, descending order plus other combinations
-  Case 2: https://github.com/grafana/mimir-prometheus/blob/f3f6711bfe6acb7e126b5e167ece3cb3516f15e2/tsdb/ooo_head_read_test.go#L180-L200
- Case 3 & Case 4: https://github.com/grafana/mimir-prometheus/blob/f3f6711bfe6acb7e126b5e167ece3cb3516f15e2/tsdb/ooo_head_read_test.go#L201-L219
  - Case 4 is also covered thanks to all the permutations of the chunks. In the description Chunk 0 is the chunk with the latest sample, in one of the permutations it would be the head chunk validating Case 4.

Work missing for Case 5 and 6.

Before that I'll try to submit a PR covering the previous 4 cases for the Chunks method.

jesusvazquez:
As crazy as it sounds but I think the query path is ready :tada: :tada: I'm tentatively closing this since now we're ready to build a mimir version with our changes and test it out in dev.

PR #163, opened 2022-03-10
Implement OOOHeadChunkReader

jesusvazquez:
Relates to https://github.com/grafana/mimir-prometheus/issues/147

This PR implements the ChunkReader interface for the OOORangeHead and its called `OOOHeadChunkReader`. We're taking the structure from the [headChunkReader](https://github.com/grafana/mimir-prometheus/blob/62252d7f8a2fa0bd12681860b8c785c8005994d6/tsdb/head_read.go#L295) implementation and making the necessary changes for out of order.
- The interface has an important method called `Chunk()` and as we explained in https://github.com/grafana/mimir-prometheus/issues/147#issuecomment-1048064175 it needs to return a merged chunk from all chunks that overlap with it. Remember that in the OOOHead chunks may not be ordered in time.
- `Chunk()` makes use of a new function called `oooMergedChunk` which is the actual function that will take care of merging the chunks.
- We're combining all the overlapping chunks into a single struct called `mergedOOOChunks` which satisfies `chunkenc.Chunk` and provides an iterator to go through all the overlapping chunks samples in order.
- Unit tests added that use the ingestion path for out of order to get the samples in and then query the chunks back to make sure the results match the expectations.

### DIFF
--- a/tsdb/head_read.go
+++ b/tsdb/head_read.go
@@ -367,6 +367,45 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
 	return mc, true, nil
 }
 
+// ooochunk returns the chunk for the HeadChunkID from memory or by m-mapping it from the disk.
+// If garbageCollect is true, it means that the returned *memChunk
+// (and not the chunkenc.Chunk inside it) can be garbage collected after its usage.
+func (s *memSeries) ooochunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *memChunk, garbageCollect bool, err error) {
+	// ix represents the index of chunk in the s.mmappedChunks slice. The chunk id's are
+	// incremented by 1 when new chunk is created, hence (id - firstChunkID) gives the slice index.
+	// The max index for the s.mmappedChunks slice can be len(s.mmappedChunks)-1, hence if the ix
+	// is len(s.mmappedChunks), it represents the next chunk, which is the head chunk.
+	ix := int(id) - int(s.firstOOOChunkID)
+	if ix < 0 || ix > len(s.oooMmappedChunks) {
+		return nil, false, storage.ErrNotFound
+	}
+	if ix == len(s.oooMmappedChunks) {
+		// TODO(jesus.vazquez) The logic in this if was true for series in order
+		// but in ooo chunks, we might flush the head chunk even if its not full.
+		if s.oooHeadChunk == nil {
+			return nil, false, errors.New("invalid head chunk")
+		}
+		return s.oooHeadChunk, false, nil
+	}
+
+	// TODO(jesus.vazquez) Here we not only need to get the chunk matching the index id
+	// but actually a merged chunk between the matching chunk and all of those that overlap
+	// Let's wait for ganesh's confirmation
+	chk, err := cdm.Chunk(s.oooMmappedChunks[ix].ref)
+	if err != nil {
+		if _, ok := err.(*chunks.CorruptionErr); ok {
+			panic(err)
+		}
+		return nil, false, err
+	}
+	mc := s.memChunkPool.Get().(*memChunk) // TODO(jesus.vazquez) Do we need a memChunkPool for OOO Chunks?
+	mc.chunk = chk
+	mc.minTime = s.oooMmappedChunks[ix].minTime // TODO(jesus.vazquez) The minTime its not really from this chunk but the new merged chunk we create from this chunk
+	mc.maxTime = s.oooMmappedChunks[ix].maxTime // TODO(jesus.vazquez) The minTime its not really from this chunk but the new merged chunk we create from this chunk
+	return mc, true, nil
+}
+
+// TODO(jesus.vazquez) What is a safe chunk??
 type safeChunk struct {
 	chunkenc.Chunk
 	s               *memSeries
--- a/tsdb/head_read.go
+++ b/tsdb/head_read.go
@@ -367,45 +367,150 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
 	return mc, true, nil
 }
 
-// ooochunk returns the chunk for the HeadChunkID from memory or by m-mapping it from the disk.
-// If garbageCollect is true, it means that the returned *memChunk
-// (and not the chunkenc.Chunk inside it) can be garbage collected after its usage.
-func (s *memSeries) ooochunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *memChunk, garbageCollect bool, err error) {
-	// ix represents the index of chunk in the s.mmappedChunks slice. The chunk id's are
-	// incremented by 1 when new chunk is created, hence (id - firstChunkID) gives the slice index.
+// oooMergedChunk returns the requested chunk based on the given chunks.Meta
+// reference from memory or by m-mapping it from the disk. The returned chunk
+// might be a merge of all the overlapping chunks, if any, amongst all the
+// chunks in the OOOHead.
+func (s *memSeries) oooMergedChunk(meta chunks.Meta, cdm chunkDiskMapper, mint, maxt int64) (chunk *mergedOOOChunks, err error) {
+	_, cid := chunks.HeadChunkRef(meta.Ref).Unpack()
+
+	// ix represents the index of chunk in the s.mmappedChunks slice. The chunk meta's are
+	// incremented by 1 when new chunk is created, hence (meta - firstChunkID) gives the slice index.
 	// The max index for the s.mmappedChunks slice can be len(s.mmappedChunks)-1, hence if the ix
 	// is len(s.mmappedChunks), it represents the next chunk, which is the head chunk.
-	ix := int(id) - int(s.firstOOOChunkID)
+	ix := int(cid) - int(s.firstOOOChunkID)
 	if ix < 0 || ix > len(s.oooMmappedChunks) {
-		return nil, false, storage.ErrNotFound
-	}
-	if ix == len(s.oooMmappedChunks) {
-		// TODO(jesus.vazquez) The logic in this if was true for series in order
-		// but in ooo chunks, we might flush the head chunk even if its not full.
-		if s.oooHeadChunk == nil {
-			return nil, false, errors.New("invalid head chunk")
-		}
-		return s.oooHeadChunk, false, nil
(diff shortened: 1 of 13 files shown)


tsdb/head_read.go
```diff
@@ -363,6 +363,163 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
...
+		}
+	}
+
+	// Next we want to sort all the collected chunks by min time so we can find
+	// those that overlap and stop when we know the rest don't.
+	sort.Sort(byMinTime(tmpChks))
+
+	oc := &mergedOOOChunks{}
+	for _, c := range tmpChks {
+		// TODO(Jesus.vazquez) Note that chunks are sorted by mintime but the head could get an old enough sample and this condition would make us discard overlapping chunks
```
jesusvazquez, line 432: @codesome Please take a look at this comment :eyes:

tsdb/head_read.go
```diff
@@ -363,6 +363,163 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
...
+		if c.Ref == meta.Ref || len(oc.chunks) > 0 && c.MinTime <= oc.chunks[len(oc.chunks)-1].MaxTime {
+
+			if c.Ref == oooHeadRef {
+				xor, err := s.oooHeadChunk.chunk.ToXor()
+				if err != nil {
+					return nil, errors.Wrap(err, "failed to convert ooo head chunk to xor chunk")
+				}
+				c.Chunk = xor
+			} else {
+				// TODO(jesus.vazquez) Get rid of this loop
```
jesusvazquez, line 446: @codesome any ideas on how to do this better?

tsdb/head_read.go
jesusvazquez: I disabled this if after this testcase https://github.com/grafana/mimir-prometheus/blob/e69e42e49ef5450973903199d75c62474cb187f1/tsdb/ooo_head_read_test.go#L268-L304

Where the head had a sample previous to the first sample of an older overlapping chunk.

tsdb/head_read.go
```diff
@@ -363,6 +363,162 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
...
+		chunkRef := chunks.ChunkRef(chunks.NewHeadChunkRef(s.ref, s.oooHeadChunkID(i)))
+		// We can skip chunks that came in later than the last known OOOLastRef
+		if chunkRef > meta.OOOLastRef {
+			break
+		}
+
+		if chunkRef == meta.OOOLastRef {
+			tmpChks = append(tmpChks, chunks.Meta{
+				MinTime: meta.MinTime,
+				MaxTime: meta.MaxTime,
```
codesome, line 413: If my suggestion is valid, then it requires a test case in the unit test.
```suggestion
				MinTime: meta.OOOLastMinTime,
				MaxTime: meta.OOOLastMaxTime,
```

tsdb/head_read.go
```diff
@@ -363,6 +363,162 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
...
+				MinTime: s.oooMmappedChunks[i].minTime,
+				MaxTime: s.oooMmappedChunks[i].maxTime,
+				Ref:     chunkRef,
+			})
+		}
+	}
+
+	// Next we want to sort all the collected chunks by min time so we can find
+	// those that overlap and stop when we know the rest don't.
+	sort.Sort(byMinTime(tmpChks))
```
codesome, line 427: Just saw that we are only sorting by MinTime. Since these are ooo chunks, two chunks can have the same MinTime. So to make the sorting deterministic, we can have another condition on the Ref if the MinTime is same.

tsdb/head_read.go
codesome: That is not an issue because, in case of Head, which will be the last chunk in our case, will use the `OOOLastMinTime` and is not affected by the samples that come in later. Right?

tsdb/head_read.go
```diff
@@ -363,6 +363,162 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
...
+
+	oc := &mergedOOOChunks{}
+	for _, c := range tmpChks {
+		// TODO(Jesus.vazquez) Note that chunks are sorted by mintime but the head could get an old enough sample and this condition would make us discard overlapping chunks
+		// if c.Ref < meta.Ref {
+		// 	continue
+		// }
```
codesome, line 434: We can remove this comment. This condition is not valid because we rely on sorted mintime.

tsdb/head_read.go
```diff
@@ -363,6 +363,162 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
...
+
+		if c.Ref == meta.Ref || len(oc.chunks) > 0 && c.MinTime <= oc.chunks[len(oc.chunks)-1].MaxTime {
+
+			if c.Ref == oooHeadRef {
+				xor, err := s.oooHeadChunk.chunk.ToXor()
```
codesome, line 439: Can we add a TODO here saying: see if we can get away with not having to do ToXor() and instead use the copy of the underlying slice. And only do ToXor() if Bytes() is called on the merged chunk.

(We don't need to do it now)

codesome (commented):
Great work! Initial comments on the non test code. Mostly nits, but one potential bug. I will look at unit tests next.

tsdb/head_read.go
```diff
@@ -363,6 +363,162 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
...
+		if s.oooHeadChunk == nil {
+			return nil, errors.New("invalid ooo head chunk")
+		}
+	}
+
+	// We create a temporary slice of chunk metas to hold the information of all
+	// possible chunks that may overlap with the requested chunk.
+	tmpChks := make([]chunks.Meta, 0, len(s.oooMmappedChunks))
+
+	oooHeadRef := chunks.ChunkRef(chunks.NewHeadChunkRef(s.ref, s.oooHeadChunkID(len(s.oooMmappedChunks))))
```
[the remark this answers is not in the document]
codesome: Because we know that series ref is going to be the same. So we need not pack it with NewHeadChunkRef() everytime.

tsdb/head_read.go
codesome: How about we have `tmpChks` as a slice of
```go
struct{
    chunks.Meta
    ix int
}
```

codesome:
Here are more test cases that I think we need to cover. 5 and 6 being a little more involved.
---
1) Similar case of 2 chunks overlapping, but the chunks are sorted in descending w.r.t. mint
```
Query Interval [------------------------------------------------------------------------------------------]

Chunk 0                                                                  [--------]

Chunk 1                                                           [-------]

Chunk 2                                             [--------]

Chunk 3: Current Head                       [--------]
```

---
2) Merge of >2 chunks. You see a chunk overlaps with only one other chunks, but at the end all 3 overlap together.
```
Query Interval [------------------------------------------------------------------------------------------]

Chunk 0                                                   [--------]

Chunk 1                                      [-------]

Chunk 2: Current Head                              [--------]
```

---
3) Query interval partially overlaps some chunks. In this case, the result will still include samples of the entire chunk and not just the query interval at this level (they are filtered somewhere else up the stack, so for the sake of tests, we can include the entire samples of the chunk)
```
Query Interval                                 [-----------------]

Chunk 0                                                   [--------]

Chunk 1                                      [-------]

Chunk 2: Current Head                              [--------]
```

---
4) Query interval overlaps the current head
```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1: Current Head                                      [--------]
```
---

Here starts the more interesting bit: ingesting more samples to the Head _after_ calling Series()

5) Current head gets an old sample and new sample after the Series() call

During Series()
```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1: Current Head                                      [--------]
```
Hence Series() would return chunk(0) based on mint.

Now after Series() call, Head gets an old sample, a newer sample, *and also a sample that is within the old chunk range* and it becomes like

```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1: Current Head                       [----------------------------------]
```

So in the result, we should still get the samples from both chunks. _But_, for chunk(1), it must not include samples beyond the lastmint and lastmaxt as seen during Series() call since it could overlap with other chunks. But it includes the new samples that came in later but is within the [lastmint, lastmaxt]. I think this bit is not taken care yet.
Another thing that it tests is that, because we will request chunk(0) here, it should still get a merged chunk, which tests if we are taking care of sorting them based on the last ooo chunk's mint as seen during Series() call.

---
6) A little more interesting one. Here, after Series() call, we not only get an old and new sample like above case, but we also end up m-mapping that chunk and create a new head chunk that overlaps with chunk(0).

During Series()
```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1: Current Head                                      [--------]
```
Hence Series() would return chunk(0) based on mint.

Now after Series() call, Head gets an old sample, a new sample, also a sample that is within the old chunk range, and many more samples such that chunk(1) is mmapped and chunk(2) is born as shown that overlaps with chunk(0)

```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1:                                    [----------------------------------]

Chunk 2: Current Head                             [--------------]
```

So the result of this should be similar to the above case and only include chunk(0) and chunk(1) and no samples outside lastmint and lastmaxt from chunk(1) but include the new samples in chunk(1) within the range (here it tests the last chunk being a m-mapped chunk, unlike the head chunk above) and _must not_ include chunk(2) since it was after the last ooo chunk.

---

tsdb/head_read.go
jesusvazquez: Yup i think so too, then the first fix of this PR seems OK.

codesome:
I just merged https://github.com/grafana/mimir-prometheus/pull/190. You may want to sync this PR with out-of-order branch to get that fix.

jesusvazquez:
To unblock the compaction work, I'm going:
- Remove the commit that added the first scenario suggested by ganesh which is not yet solved.
- Merge this PR to get the chunks code on the main branch instead of a panic.
- Open a new PR where I work on giving support to the remaining scenarios.

PR #182, opened 2022-04-01
Replay OOO samples from the WAL

codesome:
This is a follow up of https://github.com/grafana/mimir-prometheus/pull/181 and based on that PR. Once that is merged, the diff of this PR will be more digestible.

The changes look intimidating, but what I have done is:

1. Copy `loadWal` into `loadOOOWal` and remove all that is not required (i.e. keeping only samples replay and nothing to do with m-map chunks here yet).
2. Copy the `walSubsetProcessor` struct and its methods into `oooWalSubsetProcessor` and modify `oooWalSubsetProcessor.processWALSamples()` to write to the ooo chunk instead or the in-order chunk.
3. Some handling to identify if the replay error was for old WAL replay or the ooo WAL replay hence repair the appropriate WAL in db.go.

Things still missing and TODO in a follow up PR:
1. Add markers in the WAL as to when the series m-mapped a ooo chunk.
2. Replay ooo m-map chunks during the replay (also requires identifying ooo m-map chunks, needs adding the metadata during m-mapping).
3. Use the m-map marker from above to discard the WAL samples to prevent duplicate m-map chunks.

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -686,14 +686,44 @@ func (h *Head) Init(minValidTime int64) error {
 		level.Info(h.logger).Log("msg", "WAL segment loaded", "segment", i, "maxSegment", endAt)
 		h.updateWALReplayStatusRead(i)
 	}
+	walReplayDuration := time.Since(walReplayStart)
 
-	walReplayDuration := time.Since(start)
-	h.metrics.walTotalReplayDuration.Set(walReplayDuration.Seconds())
+	// Replay OOO WAL.
+	oooWalReplayStart := time.Now()
+	startFrom, endAt, e = wal.Segments(h.oooWal.Dir())
+	if e != nil {
+		return errors.Wrap(e, "finding OOO WAL segments")
+	}
+	h.startWALReplayStatus(startFrom, endAt)
+
+	for i := startFrom; i <= endAt; i++ {
+		s, err := wal.OpenReadSegment(wal.SegmentName(h.oooWal.Dir(), i))
+		if err != nil {
+			return errors.Wrap(err, fmt.Sprintf("open OOO WAL segment: %d", i))
+		}
+
+		sr := wal.NewSegmentBufReader(s)
+		err = h.loadOOOWal(wal.NewReader(sr), multiRef)
+		if err := sr.Close(); err != nil {
+			level.Warn(h.logger).Log("msg", "Error while closing the ooo wal segments reader", "err", err)
+		}
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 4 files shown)


tsdb/head.go
```diff
@@ -686,14 +686,44 @@ func (h *Head) Init(minValidTime int64) error {
...
+			return err
+		}
+		level.Info(h.logger).Log("msg", "OOO WAL segment loaded", "segment", i, "maxSegment", endAt)
+		h.updateWALReplayStatusRead(i)
+	}
+
+	oooWalReplayDuration := time.Since(oooWalReplayStart)
+
+	walTotalReplayDuration := time.Since(start)
+	h.metrics.walTotalReplayDuration.Set(walTotalReplayDuration.Seconds())
```
jesusvazquez, line 720: Would it make sense to have a label in this metric to measure WAL duration, OOOWAL duration and then the entire total?

tsdb/head_wal.go
```diff
@@ -446,6 +446,205 @@ func (wp *walSubsetProcessor) waitUntilIdle() {
...
+
+	// The records are always replayed from the oldest to the newest.
+	for d := range decoded {
+		samples := d
+		// We split up the samples into chunks of 5000 samples or less.
+		// With O(300 * #cores) in-flight sample batches, large scrapes could otherwise
+		// cause thousands of very large in flight buffers occupying large amounts
+		// of unused memory.
+		for len(samples) > 0 {
+			m := 5000
```
jesusvazquez, line 525: I'm not sure if I understand this well enough but do you think that chunks of 5000 samples or less is good for the OOO Head? Knowing that the head will mmap the head chunk after just a few samples maybe we need to tweak this? Or maybe its fine as it is. WDYT?

jesusvazquez (approved):
LGTM Left a couple of comments with one suggestion and another for a better understanding of the concurrency when reading the samples while replaying the WAL.

tsdb/head.go
codesome: Luckily the metric name is `prometheus_tsdb_data_replay_duration_seconds`

tsdb/head_wal.go
codesome: This is less to do with the size of a chunk and more to do with memory management as the comments say. The "chunk" in the comments is confusing. It means a part of max 5000 samples to be sent via the channel and not the number of samples in a chunk. Will update the comment to say "part".

tsdb/head.go
codesome: I missed to answer the exact question: the end user only cares about how much time it took in total to start TSDB. So the metric for total replay is enough. For debugging further, we have it in the logs.

jesusvazquez: Makes sense, saw the log message below :+1:

PR #184, opened 2022-04-04
Replay ooo m-map chunks on startup

codesome:
This PR replays OOO m-map chunks on startup.

To do that, we need to identify ooo m-map chunks in the first place. For that I am introducing a `EncOOOXOR`, which basically sets the first bit of the chunk encoding to 1. `EncOOOXOR` is to be only used by the Head block to identify OOO chunks. It is only used for `Head <-> ChunkDiskMapper` interactions right now and ideally does not require to be used elsewhere.

### DIFF
--- a/tsdb/chunks/old_head_chunks.go
+++ b/tsdb/chunks/old_head_chunks.go
@@ -473,7 +473,7 @@ func (cdm *OldChunkDiskMapper) Chunk(ref ChunkDiskMapperRef) (chunkenc.Chunk, er
 // and runs the provided function with information about each chunk. It returns on the first error encountered.
 // NOTE: This method needs to be called at least once after creating ChunkDiskMapper
 // to set the maxt of all the file.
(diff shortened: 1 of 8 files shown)


tsdb/chunks/old_head_chunks.go
```diff
@@ -536,7 +536,10 @@ func (cdm *OldChunkDiskMapper) IterateAllChunks(f func(seriesRef HeadSeriesRef,
 				break
 			}
 
-			idx += ChunkEncodingSize // Skip encoding.
+			// Encoding.
+			chkEnc := chunkenc.Encoding(mmapFile.byteSlice.Range(idx, idx+ChunkEncodingSize)[0])
```
jesusvazquez, line 540: Why `[0]`?

jesusvazquez (approved):

tsdb/chunks/old_head_chunks.go
codesome: ChunkEncodingSize is 1 byte, hence the result is a slice of 1 byte, and we only need that byte. Hence [0].

PR #190, opened 2022-04-05
Add overlapping chunks in decreasing order test

jesusvazquez:
We thought about a few more cases that we think the Series() method
should be able to support. https://github.com/grafana/mimir-prometheus/pull/163#issuecomment-1088588894

This particular commit brings one where 4 chunks overlap in pairs in
decreasing order (the first case mentioned in the comment above)

I think we need to review the assumption we made about leaving the head chunk
max time set to maxint again because if the head chunk contains the oldest
samples in the interval then our algorithm fails.

In this commit I'm setting the maxtime to the headchunk's existing maxtime
instead.

### DIFF
--- a/tsdb/ooo_head_read.go
+++ b/tsdb/ooo_head_read.go
@@ -68,7 +68,7 @@ func (oh *OOOHeadIndexReader) Series(ref storage.SeriesRef, lbls *labels.Labels,
 
 		tmpChks = append(tmpChks, chunks.Meta{
 			MinTime:        s.oooHeadChunk.minTime,
-			MaxTime:        math.MaxInt64, // Set the head chunks as open (being appended to).
+			MaxTime:        s.oooHeadChunk.maxTime,
 			Ref:            lastChunkRef,
 			OOOLastRef:     lastChunkRef,
 			OOOLastMinTime: lastMinT,
--- a/tsdb/ooo_head_read.go
+++ b/tsdb/ooo_head_read.go
@@ -59,49 +59,41 @@ func (oh *OOOHeadIndexReader) Series(ref storage.SeriesRef, lbls *labels.Labels,
 	var lastChunkRef chunks.ChunkRef
 	lastMinT, lastMaxT := int64(math.MaxInt64), int64(math.MaxInt64)
 
-	// We first collect the oooHeadChunk, if any. And if it exists we set value
-	// for the markers defined above.
-	if s.oooHeadChunk != nil && s.oooHeadChunk.OverlapsClosedInterval(oh.mint, oh.maxt) {
-		lastMinT = s.oooHeadChunk.minTime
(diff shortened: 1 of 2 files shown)


tsdb/ooo_head_read.go
```diff
@@ -68,7 +68,7 @@ func (oh *OOOHeadIndexReader) Series(ref storage.SeriesRef, lbls *labels.Labels,
 
 		tmpChks = append(tmpChks, chunks.Meta{
 			MinTime:        s.oooHeadChunk.minTime,
-			MaxTime:        math.MaxInt64, // Set the head chunks as open (being appended to).
+			MaxTime:        s.oooHeadChunk.maxTime,
```
jesusvazquez, line 71: If this was set to maxTime then in this part of the code no other chunks would be taken into account because the condition `if c.MinTime > maxTime` would never be met cc @codesome 

Can you think of any issues where setting it to `s.oooHeadChunk.maxTime` becomes an issue? Its true that we can newer samples while we're reading the chunks but to me this doesnt sound like a big problem. Remember that we're already keeping a query consistenty that discards new chunks.
https://github.com/grafana/mimir-prometheus/blob/88b756a18f9b7cd6f20936e0eee085bec135ac92/tsdb/ooo_head_read.go#L121-L131

Dieterbe: i'm also not a fan of this "unusual exception" for the head chunk.

imagine this scenario:
```
HEAD -----
chunk0       ----
chunk1             ----
chunk2                   ---
```

if we "inflate" the headChunk.maxTime , i think it will cause all chunks to get merged, which is unnessary.

jesusvazquez (commented):
Good job simplifying the Series() function. I also like the new declaration of chunkIntervals :+1:

jesusvazquez:
Relates to https://github.com/grafana/mimir-prometheus/issues/147 since its fixing bugs in the query path.

tsdb/ooo_head_read_test.go
```diff
@@ -23,9 +46,9 @@ func TestOOOHeadIndexReader_Series(t *testing.T) {
 		name                string
 		queryMinT           int64
 		queryMaxT           int64
-		inputChunkIntervals []struct{ mint, maxt int64 }
+		inputChunkIntervals []chunkInterval
 		expSeriesError      bool
```
codesome, line 50: We are not using this anywhere

codesome (approved):
LGTM with nits. I am fixing my own comments. Nice work on the permutations and finding the bug!

tsdb/ooo_head_read.go
Dieterbe: for the record, we resolved this in meeting with jesus, ganesh and myself.

> Dieter: Do we agree that we dont need to set head chunk maxtime to max int?
> Yes

PR #196, opened 2022-04-11
Discard unknown chunk encodings

jesusvazquez:
This is a previous step towards releasing support for out of order chunks.

Memory mapped out of order chunks have a different encoding where the first bit is used to identified them as such. This decision would make it hard to rollback to a previous version of the TSDB once any OOO chunks have been memory mapped as it would lead to corruption errors when loading them into memory.

For this reason want to "discarding unknown encoding" chunks rather than failing. This would guarantee a safe rollback in case its necessary.

### DIFF
--- a/tsdb/chunks/head_chunks_test.go
+++ b/tsdb/chunks/head_chunks_test.go
@@ -158,7 +158,7 @@ func TestChunkDiskMapper_WriteUnsupportedChunk_Chunk_IterateChunks(t *testing.T)
 		require.NoError(t, hrw.Close())
 	}()
 
-	ucSeriesRef, ucChkRef, ucMint, ucMaxt, uchunk := createUnsupportedChunk(t, 0, hrw)
+	ucSeriesRef, ucChkRef, ucMint, ucMaxt, uchunk := createUnsupportedChunk(t, 0, hrw, nil)
 
 	// Checking on-disk bytes for the first file.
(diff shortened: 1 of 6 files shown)


tsdb/chunks/head_chunks_test.go
```diff
@@ -492,3 +543,43 @@ func createChunk(t *testing.T, idx int, hrw *ChunkDiskMapper) (seriesRef HeadSer
 	<-awaitCb
 	return
 }
+
+func createUnsupportedChunk(t *testing.T, idx int, hrw *ChunkDiskMapper, hrwOld *OldChunkDiskMapper) (seriesRef HeadSeriesRef, chunkRef ChunkDiskMapperRef, mint, maxt int64, chunk chunkenc.Chunk) {
```
codesome, line 547:
```suggestion
func writeUnsupportedChunk(t *testing.T, idx int, hrw *ChunkDiskMapper, hrwOld *OldChunkDiskMapper) (seriesRef HeadSeriesRef, chunkRef ChunkDiskMapperRef, mint, maxt int64, chunk chunkenc.Chunk) {
```

codesome (approved):
LGTM with a non-blocking nit

tsdb/chunks/head_chunks_test.go
jesusvazquez: Addressed in https://github.com/grafana/mimir-prometheus/pull/196/commits/805adaea704ab9e480b4e4d7cfdcc41f7c0ada9d

PR #197, opened 2022-04-11
Add m-map markers in the OOO WBL and avoid duplicate m-map chunks

codesome:
This PR adds the m-map markers in the WBL and uses it to avoid creating duplicate m-map chunks while replaying the WBL.

Whenever it sees a m-marker, it first verifies that we have the m-map chunk on the disk that the marker is referring to, and clears out the in-memory ooo chunk of that series so that it does not get m-mapped.

There are potential edge cases if the ooo chunk size was changed between restarts, but it is not being handled yet and is a TODO in the code. It is fine to not fix it right now - it wont crash anything, some duplicates will be present which will be eventually cleaned up.

### DIFF
--- a/tsdb/head_append.go
+++ b/tsdb/head_append.go
@@ -491,7 +491,7 @@ func (a *headAppender) Commit() (err error) {
 	}()
 	collectOOORecords := func() {
 		if a.head.oooWbl == nil {
-			// WAL is not enabled. So no need to collect.
+			// WBL is not enabled. So no need to collect.
 			oooWblSamples = nil
 			oooMmapMarkers = nil
 			return
--- a/tsdb/head_wal.go
+++ b/tsdb/head_wal.go
@@ -691,7 +691,6 @@ func (wp *oooWalSubsetProcessor) processWALSamples(h *Head) (unknownRefs uint64)
 				unknownRefs++
 				continue
 			}
-			// TODO(codesome): Ignore samples that are already m-mapped. Needs a m-map marker in the WAL.
 			if _, chunkCreated, _ := ms.insert(s.T, s.V, h.chunkDiskMapper); chunkCreated {
 				h.metrics.chunksCreated.Inc()
 				h.metrics.chunks.Inc()
--- a/tsdb/record/record.go
+++ b/tsdb/record/record.go
@@ -42,7 +42,7 @@ const (
 	Tombstones Type = 3
 	// Exemplars is used to match WAL records of type Exemplars.
 	Exemplars Type = 4
-	// MmapMarkers is used to match WAL records of type MmapMarkers.
+	// MmapMarkers is used to match OOO WBL records of type MmapMarkers.
 	MmapMarkers Type = 5
 )
 


tsdb/head_append.go
```diff
@@ -490,12 +475,50 @@ func (a *headAppender) Commit() (err error) {
...
+		enc              record.Encoder
+	)
+	defer func() {
+		for i := range oooRecords {
+			a.head.putBytesBuffer(oooRecords[i][:0])
+		}
+	}()
+	collectOOORecords := func() {
+		if a.head.oooWbl == nil {
+			// WAL is not enabled. So no need to collect.
```
jesusvazquez, line 494:
```suggestion
			// WBL is not enabled. So no need to collect.
```

tsdb/head_append.go
```diff
@@ -490,12 +475,50 @@ func (a *headAppender) Commit() (err error) {
...
+		if oooMmapMarkers != nil {
+			markers := make([]record.RefMmapMarker, 0, len(oooMmapMarkers))
+			for ref, mmapRef := range oooMmapMarkers {
+				markers = append(markers, record.RefMmapMarker{
+					Ref:     ref,
+					MmapRef: mmapRef,
+				})
+			}
+			r := enc.MmapMarkers(markers, a.head.getBytesBuffer())
+			oooRecords = append(oooRecords, r)
```
jesusvazquez, line 510: Are you sure about this? Coulnd't these oooRecords be added twice, once here and then again on the next if?

tsdb/head_wal.go
```diff
@@ -637,7 +692,7 @@ func (wp *oooWalSubsetProcessor) processWALSamples(h *Head) (unknownRefs uint64)
 				continue
 			}
 			// TODO(codesome): Ignore samples that are already m-mapped. Needs a m-map marker in the WAL.
```
jesusvazquez, line 694: Do you need to remove this comment?

tsdb/head_test.go
```diff
@@ -3343,16 +3343,11 @@ func TestOOOMmapReplay(t *testing.T) {
 	copy(expMmapChunks, ms.oooMmappedChunks)
 
 	// Restart head.
-
 	require.NoError(t, h.Close())
 
-	// Remove ooo_wal to not create duplicate ooo-chunks.
-	// TODO(codesome): Remove this test hack once we have m-map markers in the WAL.
-	require.NoError(t, os.RemoveAll(filepath.Join(dir, "ooo_wal")))
```
codesome, line 3351: This part tests that no duplicate chunks are created

tsdb/head_wal.go
```diff
@@ -649,6 +704,21 @@ func (wp *oooWalSubsetProcessor) processWALSamples(h *Head) (unknownRefs uint64)
 	return unknownRefs
 }
 
+func (wp *oooWalSubsetProcessor) waitUntilIdle() {
```
codesome, line 707: This function is a copy paste from `walSubsetProcessor`

tsdb/record/record.go
```diff
@@ -42,6 +42,8 @@ const (
 	Tombstones Type = 3
 	// Exemplars is used to match WAL records of type Exemplars.
 	Exemplars Type = 4
+	// MmapMarkers is used to match WAL records of type MmapMarkers.
```
jesusvazquez, line 45: Should we say they are explicit for WBL records?

jesusvazquez (approved):
Left a few comments but looks good.

It's a bit hard for me to to understand 100% whats happening here but I get the idea and I can use this as reference to come back when I have questions.

tsdb/head_append.go
jesusvazquez: No sorry its clear now

PR #212, opened 2022-04-15
Covering more scenarios in out of order chunkreader

jesusvazquez:
Following with the comment https://github.com/grafana/mimir-prometheus/issues/147#issuecomment-1098018961 and the fantastic work done by @Dieterbe  in https://github.com/grafana/mimir-prometheus/pull/190, this PR aims at bringing a few more unit tests to the out of order chunkreader implementation.

I'm thinking we could have this PR to bring tests for cases 1 to 4 (See https://github.com/grafana/mimir-prometheus/issues/147#issuecomment-1098018961 for more context). And then open a new one for cases 5 and 6 which are probably going to require some changes.

Also, note that the tests for Series() and Chunk() are slightly different since Chunk() requires actually appending samples to then make sure that the returned merged chunk does have the expected content.

### DIFF
--- a/tsdb/ooo_head_read_test.go
+++ b/tsdb/ooo_head_read_test.go
@@ -514,6 +514,65 @@ func TestOOOHeadChunkReader_Chunk(t *testing.T) {
 			},
 		},
 		{
+			name:                 "Two windows of overlapping chunks in descending order get properly converged",
+			queryMinT:            minutes(0),
+			queryMaxT:            minutes(100),
+			firstInOrderSampleAt: minutes(120),
+			dbOpts:               opts,
+			inputSamples: tsdbutil.SampleSlice{
+				// Chunk 0
+				sample{t: minutes(40), v: float64(0)},
+				sample{t: minutes(42), v: float64(0)},
+				sample{t: minutes(44), v: float64(0)},
+				sample{t: minutes(46), v: float64(0)},
+				sample{t: minutes(50), v: float64(0)},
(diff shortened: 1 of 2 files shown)


codesome (approved):
LGTM. We can try to adopt the permutation logic of `TestOOOHeadIndexReader_Series` for the chunks tests as well. (If permutation cannot work well for all cases, we can make it opt-in per test case as well).

tsdb/ooo_head_read_test.go
```diff
@@ -582,6 +637,94 @@ func TestOOOHeadChunkReader_Chunk(t *testing.T) {
...
+					sample{t: minutes(35), v: float64(1)},
+					sample{t: minutes(42), v: float64(1)},
+					sample{t: minutes(50), v: float64(2)},
+				},
+			},
+		},
+		{
+			name:                 "Query interval partially overlaps with a triplet of chunks but still returns a single merged chunk",
+			queryMinT:            minutes(12),
+			queryMaxT:            minutes(48),
```
codesome, line 687: I think case 3 is missing, where query interval partially overlaps with a m-mapped chunks. Making this 33 can take care of case 3 and 4 together.

jesusvazquez: Thanks! You're right, addressed here https://github.com/grafana/mimir-prometheus/pull/212/commits/60d025a9622d6de4f4246f4682723f185c0e5486

PR #215, opened 2022-04-18
Bring main changes into out of order

jesusvazquez:
`out-of-order` catch up with the changes in `main`.


tsdb/head.go
```diff
@@ -736,15 +742,16 @@ func (h *Head) loadMmappedChunks(refSeries map[chunks.HeadSeriesRef]*memSeries)
...
 		lastRef = chunkRef
-		if !isOOO && maxt < h.minValidTime.Load() {
+		if encoding != chunkenc.EncOOOXOR && maxt < h.minValidTime.Load() {
 			return nil
 		}
+
 		ms, ok := refSeries[seriesRef]
 
-		if isOOO {
+		if encoding == chunkenc.EncOOOXOR {
```
codesome, line 754: Let's use `chunkenc.IsOutOfOrderChunk(encoding)` here. We might have more encoding in the future of TSDB.

codesome (commented):
Looking at https://github.com/grafana/mimir-prometheus/pull/215/commits/6f713e4185940c385c53d6d4573bf58682beb14d, I would suggest you to bring back ignoring of unknown chunk encodings. 

The removal of tests in `tsdb/chunks/...` is fine, but we might want to bring back the test from `head_test.go. And instead of the `encoding != chunkenc.EncXOR` check, we can introduce a function in `chunkenc` package saying `IsValidEncoding`. `EncUnsupportedXOR` could make some other bit 1 which makes it different from OOOXOR.

jesusvazquez:
Closing since @codesome already took care of this.

PR #216, opened 2022-04-18
add option to use the new chunk disk mapper from upstream

replay:
This removes the `OldChunkDiskMapper` which we added a few months ago.

At the time we added it because we needed a way to disable the chunk write queue and in upstream this wasn't possible. In the meantime upstream has been modified in such a way that the chunk write queue can also be disabled there by passing a size of `0`, so we should make `mimir-prometheus` match upstream again.

Edit:

I updated this PR, the current description is now this one: https://github.com/grafana/mimir-prometheus/pull/216#issuecomment-1103095848


pracucci:
I prefer to keep the `OldChunkDiskMapper` until we've migrated our prod cells to the new one (progressively). Reason why `OldChunkDiskMapper` back in the days (even if it was already supported to disable the queue-based writer with a value 0) was to have a safer way to progressively rollout to the queue-based one

jesusvazquez (commented):
When I brought prometheus/main changes into mimir-prometheus I had to add this to a test so that the behaviour would be the same

https://github.com/grafana/mimir-prometheus/blob/51f26e0cba007a7cdfc32f1ba62ce9694a90dda3/tsdb/head_test.go#L1621

That would probably conflict in future merges so lets not forget to leave it the same as in prometheus/main once we're done with this PR.

replay:
> I prefer to keep the OldChunkDiskMapper until we've migrated our prod cells to the new one (progressively). Reason why OldChunkDiskMapper back in the days (even if it was already supported to disable the queue-based writer with a value 0) was to have a safer way to progressively rollout to the queue-based one

I updated this PR, now the only thing which it does is to add an option to use the new chunk disk mapper. 
Note that this new option is independent of the chunk write queue size, that way we can first (slowly) roll out the new chunk disk mapper without activating the queue, if this works fine then we can remove the old chunk disk mapper. The primary goal of this is to make `mimir-prometheus` as similar to upstream as possible, the goal is not to roll out the chunk write queue again. In a later step we can consider re-activating the chunk write queue.
The default value of the new option is to continue using the old chunk disk mapper.

tsdb/head.go
```diff
@@ -154,6 +154,9 @@ type HeadOptions struct {
 	EnableMemorySnapshotOnShutdown bool
 
 	IsolationDisabled bool
+
+	// Temporary flag which we use to select whether we want to use the new or the old chunk disk mapper.
```
pstibrany, line 158: Can we clarify old/new terminology a bit?

```suggestion
	// Temporary flag which we use to select whether to use the new (asynchronous, used in upstream
	// Prometheus) or the old (legacy, synchronous) chunk disk mapper.
```

I understand that default value is chosen such that Mimir defaults to old (proven) mapper, but I find it little strange that mimir-prometheus defaults to "old" chunk mapper, which is no longer available in upstream Prometheus. If someone used mimir-prometheus instead of prometheus with no special settings, he would get different behaviour.

pstibrany (approved):
LGTM (no need to block on my comment)

tsdb/head.go
replay: > If someone used mimir-prometheus instead of prometheus with no special settings, he would get different behaviour.

That's true, but this is also the case already right now (before this PR). With this PR we basically just introduce to option for a user to switch to the same behavior as upstream if they want to. The plan is to roll this out in our production environments and then slowly enable the new chunk disk mapper everywhere, once it's been enabled everywhere and there haven't been any issues for a while we can then remove the old one completely. 
So TL;DR this choice of the default is to not change the behavior until the new chunk disk mapper has been proven.

replay: Regarding your suggestion to change that comment, that part is actually not really accurate `(asynchronous, used in upstream Prometheus)` because the new chunk disk mapper also works synchronously by default, only if the queue size gets set to a value above `0` then it switches to async mode.

How about this:

```
// Temporary flag which we use to select whether to use the new (used in upstream
// Prometheus) or the old (legacy) chunk disk mapper
```

replay: I updated the comment: 55cbbafe3

pstibrany: Thanks.

PR #218, opened 2022-04-19
OOOChunkReader returns consistent queries after head keeps expanding

jesusvazquez:
Alright! This PR brings query consistency in the case that the head keeps receiving samples in between Series() and Chunk() calls.

This came up from the scenarios @codesome thought about in https://github.com/grafana/mimir-prometheus/issues/147#issuecomment-1098018961 In particular case 5 and case 6.

<details>
<summary>Case 5 and 6 detailed:</summary>

Here starts the more interesting bit: ingesting more samples to the Head _after_ calling Series()

5) Current head gets an old sample and new sample after the Series() call

During Series()
```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1: Current Head                                      [--------]
```
Hence Series() would return chunk(0) based on mint.

Now after Series() call, Head gets an old sample, a newer sample, *and also a sample that is within the old chunk range* and it becomes like

```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1: Current Head                       [----------------------------------]
```

So in the result, we should still get the samples from both chunks. _But_, for chunk(1), it must not include samples beyond the lastmint and lastmaxt as seen during Series() call since it could overlap with other chunks. But it includes the new samples that came in later but is within the [lastmint, lastmaxt]. I think this bit is not taken care yet.
Another thing that it tests is that, because we will request chunk(0) here, it should still get a merged chunk, which tests if we are taking care of sorting them based on the last ooo chunk's mint as seen during Series() call.

---
6) A little more interesting one. Here, after Series() call, we not only get an old and new sample like above case, but we also end up m-mapping that chunk and create a new head chunk that overlaps with chunk(0).

During Series()
```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1: Current Head                                      [--------]
```
Hence Series() would return chunk(0) based on mint.

Now after Series() call, Head gets an old sample, a new sample, also a sample that is within the old chunk range, and many more samples such that chunk(1) is mmapped and chunk(2) is born as shown that overlaps with chunk(0)

```
Query Interval                                 [-----------------]

Chunk 0                                             [--------]

Chunk 1:                                    [----------------------------------]

Chunk 2: Current Head                             [--------------]
```

So the result of this should be similar to the above case and only include chunk(0) and chunk(1) and no samples outside lastmint and lastmaxt from chunk(1) but include the new samples in chunk(1) within the range (here it tests the last chunk being a m-mapped chunk, unlike the head chunk above) and _must not_ include chunk(2) since it was after the last ooo chunk.
</details>

## What this PR does

This PR makes a few changes to the logic in ChunkReader and
oooMergedChunk.

1. If the memory mapped chunk we're iterating was the previous head, it
   will be appended with the OOOLastMinT and OOOLastMaxT markers.
2. When chunks are loaded from disk, if they are the previously known
   head, the chunk will be loaded and then wrapped into a boundedChunk
   so that when its iterated, only samples within the OOOLastMinT and
   OOOLastMaxT markers are returned.
3. Two new structs:
  - boundedChunk wraps a chunk and overrides Bytes() and Iterator()
  - boundedIterator wraps an Iterator and overrides the methods to
    return only truthful answers for samples within bounds.
4. Unit test cases for scenarios 5 and 6

### DIFF
--- a/tsdb/chunkenc/ooo.go
+++ b/tsdb/chunkenc/ooo.go
@@ -4,49 +4,49 @@ import (
 	"sort"
 )
 
-type sample struct {
-	t int64
-	v float64
+type Sample struct {
+	T int64
+	V float64
 }
 
-// OOOChunk maintains samples in time-ascending order.
+// OOOChunk maintains Samples in time-ascending order.
 // Inserts for timestamps already seen, are dropped.
 // Samples are stored uncompressed to allow easy sorting.
 // Perhaps we can be more efficient later.
 type OOOChunk struct {
-	samples []sample
+	Samples []Sample
 }
 
 func NewOOOChunk(capacity int) *OOOChunk {
-	return &OOOChunk{samples: make([]sample, 0, capacity)}
+	return &OOOChunk{Samples: make([]Sample, 0, capacity)}
 }
 
-// Insert inserts the sample such that order is maintained.
+// Insert inserts the Sample such that order is maintained.
 // Returns false if insert was not possible due to the same timestamp already existing.
 func (o *OOOChunk) Insert(t int64, v float64) bool {
-	// find index of sample we should replace
-	i := sort.Search(len(o.samples), func(i int) bool { return o.samples[i].t >= t })
+	// find index of Sample we should replace
+	i := sort.Search(len(o.Samples), func(i int) bool { return o.Samples[i].T >= t })
 
-	if i >= len(o.samples) {
+	if i >= len(o.Samples) {
 		// none found. append it at the end
-		o.samples = append(o.samples, sample{t, v})
+		o.Samples = append(o.Samples, Sample{t, v})
 		return true
 	}
 
-	if o.samples[i].t == t {
+	if o.Samples[i].T == t {
 		return false
 	}
 
-	// expand length by 1 to make room. use a zero sample, we will overwrite it anyway
-	o.samples = append(o.samples, sample{})
-	copy(o.samples[i+1:], o.samples[i:])
-	o.samples[i] = sample{t, v}
+	// expand length by 1 to make room. use a zero Sample, we will overwrite it anyway
+	o.Samples = append(o.Samples, Sample{})
+	copy(o.Samples[i+1:], o.Samples[i:])
+	o.Samples[i] = Sample{t, v}
 
 	return true
 }
 
 func (o *OOOChunk) NumSamples() int {
-	return len(o.samples)
+	return len(o.Samples)
 }
 
 func (o *OOOChunk) ToXor() (*XORChunk, error) {
@@ -55,8 +55,8 @@ func (o *OOOChunk) ToXor() (*XORChunk, error) {
 	if err != nil {
 		return nil, err
 	}
-	for _, s := range o.samples {
-		app.Append(s.t, s.v)
+	for _, s := range o.Samples {
+		app.Append(s.T, s.V)
 	}
 	return x, nil
 }
@@ -67,7 +67,7 @@ func (o *OOOChunk) ToXor() (*XORChunk, error) {
 // func (c *OOOChunk) NumSamples() int {
 // func (c *OOOChunk) Compact() {
 
-// func (it *oooIterator) Seek(t int64) bool {
+// func (it *oooIterator) Seek(T int64) bool {
 // func (it *oooIterator) At() (int64, float64) {
 // func (it *oooIterator) Err() error {
 // func (it *oooIterator) Next() bool {
--- a/tsdb/chunkenc/ooo.go
+++ b/tsdb/chunkenc/ooo.go
@@ -9,7 +9,7 @@ type sample struct {
 	v float64
 }
 
-// OOOChunk maintains Samples in time-ascending order.
+// OOOChunk maintains samples in time-ascending order.
 // Inserts for timestamps already seen, are dropped.
 // Samples are stored uncompressed to allow easy sorting.
 // Perhaps we can be more efficient later.
--- a/tsdb/chunkenc/ooo.go
+++ b/tsdb/chunkenc/ooo.go
@@ -4,9 +4,9 @@ import (
 	"sort"
 )
 
-type Sample struct {
-	T int64
-	V float64
+type sample struct {
+	t int64
+	v float64
 }
 
 // OOOChunk maintains Samples in time-ascending order.
@@ -14,39 +14,39 @@ type Sample struct {
 // Samples are stored uncompressed to allow easy sorting.
 // Perhaps we can be more efficient later.
 type OOOChunk struct {
-	Samples []Sample
+	samples []sample
 }
 
 func NewOOOChunk(capacity int) *OOOChunk {
-	return &OOOChunk{Samples: make([]Sample, 0, capacity)}
+	return &OOOChunk{samples: make([]sample, 0, capacity)}
 }
 
-// Insert inserts the Sample such that order is maintained.
+// Insert inserts the sample such that order is maintained.
 // Returns false if insert was not possible due to the same timestamp already existing.
 func (o *OOOChunk) Insert(t int64, v float64) bool {
-	// find index of Sample we should replace
-	i := sort.Search(len(o.Samples), func(i int) bool { return o.Samples[i].T >= t })
+	// find index of sample we should replace
+	i := sort.Search(len(o.samples), func(i int) bool { return o.samples[i].t >= t })
 
-	if i >= len(o.Samples) {
+	if i >= len(o.samples) {
 		// none found. append it at the end
-		o.Samples = append(o.Samples, Sample{t, v})
+		o.samples = append(o.samples, sample{t, v})
 		return true
 	}
 
-	if o.Samples[i].T == t {
+	if o.samples[i].t == t {
 		return false
 	}
 
-	// expand length by 1 to make room. use a zero Sample, we will overwrite it anyway
-	o.Samples = append(o.Samples, Sample{})
-	copy(o.Samples[i+1:], o.Samples[i:])
-	o.Samples[i] = Sample{t, v}
+	// expand length by 1 to make room. use a zero sample, we will overwrite it anyway
+	o.samples = append(o.samples, sample{})
+	copy(o.samples[i+1:], o.samples[i:])
+	o.samples[i] = sample{t, v}
 
 	return true
 }
 
 func (o *OOOChunk) NumSamples() int {
-	return len(o.Samples)
+	return len(o.samples)
 }
 
 func (o *OOOChunk) ToXor() (*XORChunk, error) {
@@ -55,19 +55,26 @@ func (o *OOOChunk) ToXor() (*XORChunk, error) {
 	if err != nil {
 		return nil, err
 	}
-	for _, s := range o.Samples {
-		app.Append(s.T, s.V)
+	for _, s := range o.samples {
+		app.Append(s.t, s.v)
 	}
 	return x, nil
 }
 
-// TODO: to support querying, implement Iterator
-
-// func (c *OOOChunk) Iterator(it Iterator) Iterator {
-// func (c *OOOChunk) NumSamples() int {
-// func (c *OOOChunk) Compact() {
-
-// func (it *oooIterator) Seek(T int64) bool {
-// func (it *oooIterator) At() (int64, float64) {
-// func (it *oooIterator) Err() error {
-// func (it *oooIterator) Next() bool {
+func (o *OOOChunk) ToXorBetweenTimestamps(mint, maxt int64) (*XORChunk, error) {
+	x := NewXORChunk()
+	app, err := x.Appender()
+	if err != nil {
+		return nil, err
+	}
+	for _, s := range o.samples {
+		if s.t < mint {
+			continue
+		}
+		if s.t > maxt {
+			break
+		}
+		app.Append(s.t, s.v)
+	}
+	return x, nil
+}
--- a/tsdb/head_read.go
+++ b/tsdb/head_read.go
@@ -459,7 +459,8 @@ func (s *memSeries) oooMergedChunk(meta chunks.Meta, cdm chunkDiskMapper, mint, 
 					}
 				} else {
 					// We need to remove samples that are outside of the markers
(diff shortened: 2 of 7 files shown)


tsdb/chunkenc/ooo.go
```diff
@@ -4,49 +4,49 @@ import (
 	"sort"
 )
 
-type sample struct {
-	t int64
-	v float64
+type Sample struct {
+	T int64
+	V float64
 }
```
jesusvazquez, line 10: I'm exporting everything here to have the samples accessible from the ooo head. Maybe a better way to do this is to have an interator.

tsdb/head_read.go
```diff
@@ -437,9 +437,31 @@ func (s *memSeries) oooMergedChunk(meta chunks.Meta, cdm chunkDiskMapper, mint,
...
 		if c.meta.Ref == meta.Ref || len(mc.chunks) > 0 && c.meta.MinTime <= mc.chunks[len(mc.chunks)-1].MaxTime {
 			if c.meta.Ref == oooHeadRef {
-				xor, err := s.oooHeadChunk.chunk.ToXor() // TODO(jesus.vazquez) (This is an optimization idea that has no priority and might not be that useful) See if we could use a copy of the underlying slice. That would leave the more expensive ToXor() function only for the usecase where Bytes() is called.
-				if err != nil {
-					return nil, errors.Wrap(err, "failed to convert ooo head chunk to xor chunk")
+				var xor *chunkenc.XORChunk
+				// If head chunk min and max time match the meta OOO markers
+				// that means that the chunk has not expanded so we can append
+				// it as it is.
+				if s.oooHeadChunk.minTime == meta.OOOLastMinTime && s.oooHeadChunk.maxTime == meta.OOOLastMaxTime {
```
jesusvazquez, line 444: This was a happy idea and made case 5 work https://github.com/grafana/mimir-prometheus/issues/147#issuecomment-1098018961 but its not enough for case 6.

tsdb/head_read.go
```diff
@@ -437,9 +437,31 @@ func (s *memSeries) oooMergedChunk(meta chunks.Meta, cdm chunkDiskMapper, mint,
...
+					for _, sample := range s.oooHeadChunk.chunk.Samples {
+						if sample.T < meta.OOOLastMinTime {
+							continue
+						}
+						if sample.T > meta.OOOLastMaxTime {
+							break
+						}
+						oooChunk.Insert(sample.T, sample.V)
+					}
+					xor, err = oooChunk.ToXor()
```
codesome, line 461: I will be more efficient (memory and CPU) if `oooChunk` here was a XOR chunk to start with and we create XOR chunk directly.

jesusvazquez: Addressed in https://github.com/grafana/mimir-prometheus/pull/218/commits/645411c8a9ff40ac673b712de2edfc8d34c7fa19

tsdb/head_read.go
```diff
@@ -450,7 +480,16 @@ func (s *memSeries) oooMergedChunk(meta chunks.Meta, cdm chunkDiskMapper, mint,
...
 					return nil, err
 				}
-				c.meta.Chunk = chk
+				if c.meta.Ref == meta.OOOLastRef &&
+					(c.origMinT != meta.OOOLastMinTime || c.origMaxT != meta.OOOLastMaxTime) {
+					// The head expanded and was memory mapped so now we need to
+					// wrap the chunk within a chunk that doesnt allows us to iterate
+					// through samples out of the OOOLastMinT and OOOLastMaxT
+					// markers.
+					c.meta.Chunk = boundedChunk{chk, meta.OOOLastMinTime, meta.OOOLastMaxTime}
```
jesusvazquez, line 489: This is what makes case 6 work https://github.com/grafana/mimir-prometheus/issues/147#issuecomment-1098018961

tsdb/head_read.go
```diff
@@ -511,6 +550,80 @@ func (o mergedOOOChunks) NumSamples() int {
...
+func (b boundedIterator) Seek(t int64) bool {
+	if t < b.minT || t > b.maxT {
+		return false
+	}
+	return b.Iterator.Seek(t)
+}
+
+func (b boundedIterator) At() (int64, float64) {
+	return b.Iterator.At()
+}
```
jesusvazquez, line 620: I could probably remove this

tsdb/head_read.go
```diff
@@ -511,6 +550,80 @@ func (o mergedOOOChunks) NumSamples() int {
...
+
+func (b boundedIterator) Err() error {
+	// TODO(jesus.vazquez) should this return an error if our Seek() method returned false for a sample outside bounds?
```
codesome, line 623: No

tsdb/head_read.go
```diff
@@ -511,6 +550,80 @@ func (o mergedOOOChunks) NumSamples() int {
...
+	for b.Iterator.Next() {
+		t, _ := b.Iterator.At()
+		if t < b.minT {
+			continue
+		} else if t > b.maxT {
+			return false
+		}
+		break
+	}
+	return true
```
codesome, line 608: Reason: If `b.Iterator.Next()` returned `false`, then we must be returning `false` from that point onwards.
```suggestion
		return true
	}
	return false
```

tsdb/head_read.go
```diff
@@ -511,6 +550,80 @@ func (o mergedOOOChunks) NumSamples() int {
...
+}
+
+func (b boundedIterator) Seek(t int64) bool {
+	if t < b.minT || t > b.maxT {
+		return false
+	}
+	return b.Iterator.Seek(t)
+}
```
codesome, line 616: Seek is like a pointer in the iterator that only moves forward. If seek is called for `t`, it seeks until it finds a sample that is `>= t`

So when `t < minT`, the first Seek'ed sample must be after minT, so we seek for minT instead and make sure the seeked sample if before maxT.

Next if `t > maxT`, it means that it permanently moves forward the pointer there, so any subsequent Seek() or Next() call must return false. Hence we move forward the pointer.

```suggestion
func (b boundedIterator) Seek(t int64) bool {
	if t < b.minT {
		// We must seek at least up to b.minT if it is asked for something before that.
		ok := b.Iterator.Seek(b.minT)
		if !ok {
			return false
		}
		t, _ := b.Iterator.At()
		return t <= b.maxT 
	}
	if t > b.maxT {
		// We seek anyway so that the subsequent Next() calls will also return false.
		b.Iterator.Seek(t)
		return false
	}
	return b.Iterator.Seek(t)
}
```

codesome (commented):
Few comments on the implementation, I am yet to review the unit test. We should probably add unit tests for `boundedIterator`.

tsdb/ooo_head_read_test.go
```diff
@@ -768,9 +768,183 @@ func TestOOOHeadChunkReader_Chunk(t *testing.T) {
...
+				sample{t: minutes(10), v: float64(1)},
+				sample{t: minutes(32), v: float64(1)},
+				sample{t: minutes(50), v: float64(1)},
+				// Chunk 1 gets mmapped and Chunk 2, the new head is born
+				sample{t: minutes(25), v: float64(2)},
+				sample{t: minutes(31), v: float64(2)},
+			},
+			expChunkError: false,
+			// ts (in minutes)         0       10       20       30       40       50       60       70       80       90       100
+			// Query Interval                       [----------------------]
```
codesome, line 875:
```suggestion
			// ts (in minutes)         0       10       20       30       40       50       60       70       80       90       100
			// Query Interval          [-------------------------------------------------------------------------------------]
```

tsdb/ooo_head_read_test.go
```diff
@@ -768,9 +768,183 @@ func TestOOOHeadChunkReader_Chunk(t *testing.T) {
...
+				// Chunk 1 Head
+				sample{t: minutes(25), v: float64(1)},
+				sample{t: minutes(35), v: float64(1)},
+			},
+			samplesAfterSeriesCall: tsdbutil.SampleSlice{
+				sample{t: minutes(10), v: float64(1)},
+				sample{t: minutes(32), v: float64(1)},
+				sample{t: minutes(50), v: float64(1)},
+				// Chunk 1 gets mmapped and Chunk 2, the new head is born
+				sample{t: minutes(25), v: float64(2)},
```
codesome, line 870: Nice choice of test sample!

tsdb/ooo_head_read_test.go
```diff
@@ -768,9 +768,183 @@ func TestOOOHeadChunkReader_Chunk(t *testing.T) {
...
+		require.NoError(t, err)
+		return ref
+	}
+
+	tests := []struct {
+		name                   string
+		queryMinT              int64
+		queryMaxT              int64
+		firstInOrderSampleAt   int64
+		dbOpts                 *Options
```
codesome, line 799: Do you need this here since you are using the same everywhere?

tsdb/ooo_head_read_test.go
```diff
@@ -768,9 +768,183 @@ func TestOOOHeadChunkReader_Chunk(t *testing.T) {
...
+	opts.OOOAllowance = 120 * time.Minute.Milliseconds()
+
+	s1 := labels.FromStrings("l", "v1")
+	minutes := func(m int64) int64 { return m * time.Minute.Milliseconds() }
+
+	appendSample := func(app storage.Appender, l labels.Labels, timestamp int64, value float64) storage.SeriesRef {
+		ref, err := app.Append(0, l, timestamp, value)
+		require.NoError(t, err)
+		return ref
+	}
```
codesome, line 792: Except `minutes()`, maybe all can move closer to where it is used, i.e. just before the start of the loop (see my other comment about `opts`)

tsdb/ooo_head_read_test.go
```diff
@@ -768,9 +768,183 @@ func TestOOOHeadChunkReader_Chunk(t *testing.T) {
...
+			db := newTestDBWithOpts(t, tc.dbOpts)
+
+			app := db.Appender(context.Background())
+			s1Ref := appendSample(app, s1, tc.firstInOrderSampleAt, float64(tc.firstInOrderSampleAt/1*time.Minute.Milliseconds()))
+			require.NoError(t, app.Commit())
+
+			// OOO few samples for s1.
+			app = db.Appender(context.Background())
+			for _, s := range tc.initialSamples {
+				appendSample(app, s1, s.T(), s.V())
```
codesome, line 908: I did not see this before in earlier PRs, but appendSample is technically just a 2 line function. We can inline it here and below instead.

codesome (commented):
Nice work on the tests! I only have some nits here

tsdb/chunkenc/ooo.go
jesusvazquez: I'm unexporting everything and creating a ToXorBetweenTimestamps() method instead.

tsdb/ooo_head_read_test.go
jesusvazquez: I did refactor `opts` :+1:  minutes() and appendSample() are right before the test cases and the loops start.

tsdb/ooo_head_read_test.go
jesusvazquez: I'd rather keep it as it is. I also like inline functions but I don't like to repeat them.

jesusvazquez:
@codesome Added unit tests for boundedChunk please re-review.

tsdb/head_read_test.go
```diff
@@ -0,0 +1,104 @@
...
+		inputMaxT  int64
+		expBytes   []byte
+	}{
+		{
+			name:       "if there are no samples only expect first two bytes",
+			inputChunk: newTestChunk(0),
+			expBytes:   []byte{0x0, 0x0},
+		},
+		{
+			name:       "if there are no bounds only first sample is returned",
```
codesome, line 26: This is actually setting the bounds [0.0], so the description needs to be updated

tsdb/head_read_test.go
```diff
@@ -0,0 +1,104 @@
...
+			inputChunk: newTestChunk(10),
+			expBytes:   []byte{0x0, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
+		},
+		{
+			name:       "if there are bounds only bytes from samples included are returned",
+			inputChunk: newTestChunk(10),
+			inputMinT:  1,
+			inputMaxT:  8,
+			expBytes:   []byte{0x0, 0x8, 0x2, 0x3f, 0xf0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0xc2, 0x5f, 0xff, 0x6c, 0x6, 0xd6, 0x16, 0xda, 0xd, 0xb0, 0x2d, 0x2d, 0x42, 0x78},
```
codesome, line 35: Mentioning bytes like this does not give much confidence on if it is right. So it would be best to build the expect chunk and compare bytes of that.

tsdb/head_read_test.go
```diff
@@ -0,0 +1,104 @@
...
+		inputMaxT  int64
+		expSamples []sample
+	}{
+		{
+			name:       "if there are no samples it returns nothing",
+			inputChunk: newTestChunk(0),
+			expSamples: nil,
+		},
+		{
+			name:       "if there are no bounds set only sample at ts 0 is returned",
```
codesome, line 60: same thing about description here

tsdb/head_read_test.go
```diff
@@ -0,0 +1,104 @@
...
+			inputChunk: newTestChunk(1),
+			expSamples: []sample{
+				{0, 0},
+			},
+		},
+		{
+			name:       "if there are bounds set only samples within them are returned",
```
codesome, line 67: we should also test the case where mint is actual mint while maxt is less than actual maxt, and the other way round

tsdb/ooo_head_read_test.go
codesome: My intention was keeping these after the `tests` declaration, because have to scroll a lot to see what appendSample does.

codesome (commented):
We also need tests for boundedChunk's Seek(). I am already fixing them and will push shortly.

codesome (approved):
:tada:

PR #220, opened 2022-04-20
Add compaction for out of order samples in the Head

codesome:
This adds compaction to generate multiple blocks out of OOO Head as per [this design doc](https://docs.google.com/document/d/1Kppm7qL9C-BJB1j6yb6-9ObG3AbdZnFUBYPNNWwDBYM/edit#heading=h.f3vr8krap9ta).

Since our aim is to support a relatively smaller OOOAllowance, the approach used here is to iterate over all possible block ranges for OOO Head and do compaction one at a time. For example if OOO Head had samples from 1.5h to 4.5h. It will try compaction for these 3 ranges one after another - [0h, 2h), [2h, 4h), [4h, 6) - while querying for all OOO series during each iteration.

But this step is done only once for all blocks together: 

At the start:
1. Cut a new WBL file and record the last WBL file seen at that time.
2. M-map all the active ooo chunks and keep track of the last m-map chunk ref seen.

<then the writing of blocks happen as per above>

After blocks have been written:
1. OOO m-map chunks are garbage collected for all ooo m-map chunks that are before the last m-map chunks seen above. Empty series are also garbage collected here.
2. WBL files before the above seen file number are deleted.

### DIFF
--- a/tsdb/compact.go
+++ b/tsdb/compact.go
@@ -472,6 +472,17 @@ func (c *LeveledCompactor) compact(dest string, dirs []string, open []*Block, sh
 		uids = append(uids, b.meta.ULID.String())
 	}
 
+	// TODO: create outBlocks := [][]shardedBlock instead.
+	// That way, outBlocks[x] will still be label based sharding, while outBlocks[x][y] will be time based for out of order samples.
+	// For compactions that is not for out of order samples, len(outBlocks[x]) will always be 1.
+
+	// During ingestion of samples we can identify which ooo blocks will exists so that
+	// we dont have to prefill symbols and etc for the blocks that will be empty.
+	// With this, len(outBlocks[x]) will still be the same for all x so that we can pick blocks easily.
+	// Just that, only some of the outBlocks[x][y] will be valid and populated based on preexisting knowledge of
+	// which blocks to expect.
+	// In case we see a sample that is not present in the estimated block ranges, we will create them on flight.
+
 	outBlocks := make([]shardedBlock, shardCount)
 	outBlocksTime := ulid.Now() // Make all out blocks share the same timestamp in the ULID.
 	for ix := range outBlocks {
@@ -958,6 +969,20 @@ func (c *LeveledCompactor) populateBlock(blocks []BlockReader, minT, maxT int64,
 		}
 
 		debugOutOfOrderChunks(chks, c.logger)
+
+		// TODO(codesome):
+		// Split the chunks into different blocks.
+		// And then call blockWriters for them all.
+
+		//splitChunks := func(chks []chunks.Meta) []chunks.Meta {
+		//	// Splits chunks into pieces such that it does not cross block boundaries.
+		//  // Also merge smaller chunks of same block if possible.
+		//}
+		//chks = splitChunks(chks)
+
+		//selectBlock := func(lbls labels.Labels, chkMint, chkMaxt, blockMint, blockMaxt, blockRange int64) {
+		//	// Based on the chunk range and series labels, decide which block does the chunk belong to.
+		//}
 
 		obIx := uint64(0)
 		if len(outBlocks) > 1 {
--- a/tsdb/compact.go
+++ b/tsdb/compact.go
@@ -76,7 +76,8 @@ type Compactor interface {
 	//  * Returns empty ulid.ULID{}.
 	Compact(dest string, dirs []string, open []*Block) (ulid.ULID, error)
 
-	// TODO: write comments.
+	// CompactOOO creates a new block per possible block range in the compactor's directory from the OOO Head given.
+	// Each ULID in the result corresponds to a block in a unique time range.
 	CompactOOO(dest string, oooHead *OOOCompactionHead) (result []ulid.ULID, err error)
 }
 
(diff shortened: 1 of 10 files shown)


tsdb/compact.go
```diff
@@ -541,6 +545,130 @@ func (c *LeveledCompactor) compact(dest string, dirs []string, open []*Block, sh
 	return nil, errs.Err()
 }
 
+// CompactOOOWithSplitting splits the input OOO Head into shardCount number of output blocks
+// per possible block range, and returns slice of block IDs. In result[i][j],
+// 'j' corresponds to the shard index while 'i' corresponds to a single time range of blocks.
```
jesusvazquez, line 550:
```suggestion
// 'i' corresponds to a single time range of blocks while 'j' corresponds to the shard index.
```

I'd expect to read the index keys in order

tsdb/compact.go
```diff
@@ -541,6 +545,130 @@ func (c *LeveledCompactor) compact(dest string, dirs []string, open []*Block, sh
...
+	// During ingestion of samples we can identify which ooo blocks will exists so that
+	// we dont have to prefill symbols and etc for the blocks that will be empty.
+	// With this, len(outBlocks[x]) will still be the same for all x so that we can pick blocks easily.
+	// Just that, only some of the outBlocks[x][y] will be valid and populated based on preexisting knowledge of
+	// which blocks to expect.
+	// In case we see a sample that is not present in the estimated block ranges, we will create them on flight.
+	outBlocks := make([][]shardedBlock, 0)
+	outBlocksTime := ulid.Now() // Make all out blocks share the same timestamp in the ULID.
+	blockSize := oooHead.ChunkRange()
+	oooHeadMint, oooHeadMaxt := oooHead.MinTime(), oooHead.MaxTime()
```
jesusvazquez, line 593: Shouldnt these be oooHead.MinOOOTime()? I'm adding them in https://github.com/grafana/mimir-prometheus/pull/222

If that is the case could you add a TODO here?

tsdb/head.go
```diff
@@ -1135,6 +1138,41 @@ func (h *Head) truncateWAL(mint int64) error {
...
+				// The actual min time is in the appendable window.
+				// So we set the mint to the appendableMinValidTime.
+				h.minTime.Store(appendableMinValidTime)
+				h.minValidTime.Store(appendableMinValidTime)
+			}
+		}
+
+		// Truncate the chunk m-mapper.
+		if err := h.chunkDiskMapper.Truncate(minMmapFile); err != nil {
+			return errors.Wrap(err, "truncate by file no chunks.HeadReadWriter in truncateOOO")
```
jesusvazquez, line 1169: I don't understand this error message, without context I see that Truncate() may have failed but why `no chunks.HeadReadWriter in truncateOOO`??

tsdb/ooo_head_read.go
```diff
@@ -212,3 +221,190 @@ func (cr OOOHeadChunkReader) Chunk(meta chunks.Meta) (chunkenc.Chunk, error) {
...
+
+func (ch *OOOCompactionHead) Tombstones() (tombstones.Reader, error) {
+	return tombstones.NewMemTombstones(), nil
+}
+
+func (ch *OOOCompactionHead) Meta() BlockMeta {
+	var id [16]byte
+	copy(id[:], "copy(id[:], \"ooo_compact_head\")")
+	return BlockMeta{
+		MinTime: ch.mint,
```
jesusvazquez, line 320: Shouldnt this be ch.MinOOOTime() I'm adding them in https://github.com/grafana/mimir-prometheus/pull/222

If so can you please add a TODO here and in the line below?

tsdb/ooo_head_read.go
```diff
@@ -212,3 +221,190 @@ func (cr OOOHeadChunkReader) Chunk(meta chunks.Meta) (chunkenc.Chunk, error) {
...
+// obtained from this only looks at the m-map chunks within the given time ranges while not looking
+// beyond the ch.lastMmapRef.
+// Only the method of BlockReader interface are valid for the cloned OOOCompactionHead.
+func (ch *OOOCompactionHead) CloneForTimeRange(mint, maxt int64) *OOOCompactionHead {
+	return &OOOCompactionHead{
+		oooIR:       NewOOOHeadIndexReader(ch.oooIR.head, mint, maxt),
+		lastMmapRef: ch.lastMmapRef,
+		postings:    ch.postings,
+		chunkRange:  ch.chunkRange,
+		mint:        ch.mint,
```
jesusvazquez, line 339: Shouldnt this be ch.MinOOOTime() I'm adding them in https://github.com/grafana/mimir-prometheus/pull/222

If so can you please add a TODO here and in the line below?

tsdb/ooo_head_read.go
```diff
@@ -212,3 +221,190 @@ func (cr OOOHeadChunkReader) Chunk(meta chunks.Meta) (chunkenc.Chunk, error) {
...
+		maxt:        ch.maxt,
+	}
+}
+
+func (ch *OOOCompactionHead) Size() int64                            { return 0 }
+func (ch *OOOCompactionHead) MinTime() int64                         { return ch.mint }
```
jesusvazquez, line 345: Shouldnt this be ch.MinOOOTime() I'm adding them in https://github.com/grafana/mimir-prometheus/pull/222

If so can you please add a TODO here and in the line below?

jesusvazquez (commented):
I think I got most of it :tada: 

I see that in the NewOOOCompactionHead we're still missing to mmap all in memory chunks and track the last mmapped chunk. Also that we're missing postings. 

Can I assume those are 3 blockers to merge this?

codesome:
>I see that in the NewOOOCompactionHead we're still missing to mmap all in memory chunks and track the last mmapped chunk.

If you are talking about that TODO comment, then those are actually old comments that need cleanup. We are m-mapping the chunks and tracking the last one if you look at the code. I will remove that comment.

> Also that we're missing postings.

That is fine for compaction. Since we are anyway going through all the series, we already get the ooo postings efficiently. We need ooo postings separately for queries mainly. So for compaction it is not a blocker.

tsdb/compact.go
codesome: oooHead is actually OOOCompactionHead, which calculates the actual ooo mint and ooo maxt (it is done in NewOOOCompactionHead). So this is correct.

tsdb/ooo_head_read.go
codesome: oooHead is actually OOOCompactionHead, which calculates the actual ooo mint and ooo maxt (it is done in NewOOOCompactionHead). So this is correct.

tsdb/ooo_head_read.go
codesome: oooHead is actually OOOCompactionHead, which calculates the actual ooo mint and ooo maxt (it is done in NewOOOCompactionHead). So this is correct.

tsdb/ooo_head_read.go
codesome: oooHead is actually OOOCompactionHead, which calculates the actual ooo mint and ooo maxt (it is done in NewOOOCompactionHead). So this is correct.

tsdb/head.go
codesome: It should actually read as "file number", will make it clearer :)

jesusvazquez (approved):
LGTM!

tsdb/compact.go
codesome: Thanks, will update this

PR #222, opened 2022-04-20
Update db Querier to read from the ooo range head

jesusvazquez:
The way the querier reads from the tsdb is that it iterates all the existing blocks that overlap with the query interval (mint,maxt). And blocks can be 1) persistent blocks 2) in order head implementing block reader and now 3) out of order head implementing block reader :tada: 

What this PR does:
- Extends the db querier to also read form the out of order range head.
- Implement noop tombstone reader for the ooo range head. As stated in the design doc, we're not supporting tombstones for out of order.
- Add to the head a tracker for minOOOTime and maxOOOTime so that we know how much does the out of order head expand in time. 
  - Note: We also need to make sure that this markers are properly tracked and updated during the ingestion of new ooo samples and compaction.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -1657,7 +1657,7 @@ func (db *DB) Querier(_ context.Context, mint, maxt int64) (storage.Querier, err
 	}
 
 	var outOfOrderHeadQuerier storage.Querier
-	if maxt >= db.head.minOOOTime.Load() {
+	if overlapsClosedInterval(mint, maxt, db.head.MinOOOTime(), db.head.MaxOOOTime()) {
 		rh := NewOOORangeHead(db.head, mint, maxt)
 		var err error
 		outOfOrderHeadQuerier, err = NewBlockQuerier(rh, mint, maxt)
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -1664,25 +1664,6 @@ func (db *DB) Querier(_ context.Context, mint, maxt int64) (storage.Querier, err
 		if err != nil {
 			return nil, errors.Wrapf(err, "open block querier for ooo head %s", rh)
 		}
-
-		// TODO(jesus.vazquez) I am not sure the truncation process affects the ooo head. Let's review this.
-		// Getting the querier above registers itself in the queue that the truncation waits on.
-		// So if the querier is currently not colliding with any truncation, we can continue to use it and still
-		// won't run into a race later since any truncation that comes after will wait on this querier if it overlaps.
-		shouldClose, getNew, newMint := db.head.IsQuerierCollidingWithTruncation(mint, maxt)
-		if shouldClose {
-			if err := outOfOrderHeadQuerier.Close(); err != nil {
-				return nil, errors.Wrapf(err, "closing ooo head block querier %s", rh)
-			}
-			outOfOrderHeadQuerier = nil
-		}
-		if getNew {
-			rh := NewOOORangeHead(db.head, newMint, maxt)
-			outOfOrderHeadQuerier, err = NewBlockQuerier(rh, newMint, maxt)
-			if err != nil {
-				return nil, errors.Wrapf(err, "open block querier for ooo head while getting new querier %s", rh)
-			}
-		}
 	}
 
 	blockQueriers := make([]storage.Querier, 0, len(blocks))
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -1657,7 +1657,6 @@ func (db *DB) Querier(_ context.Context, mint, maxt int64) (storage.Querier, err
 	}
 
 	var outOfOrderHeadQuerier storage.Querier
-	db.head.MinTime()
 	if maxt >= db.head.minOOOTime.Load() {
 		rh := NewOOORangeHead(db.head, mint, maxt)
 		var err error
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -1628,13 +1628,13 @@ func (db *DB) Querier(_ context.Context, mint, maxt int64) (storage.Querier, err
 			blocks = append(blocks, b)
 		}
 	}
-	var headQuerier storage.Querier
+	var inOrderHeadQuerier storage.Querier
 	if maxt >= db.head.MinTime() {
 		rh := NewRangeHead(db.head, mint, maxt)
 		var err error
-		headQuerier, err = NewBlockQuerier(rh, mint, maxt)
+		inOrderHeadQuerier, err = NewBlockQuerier(rh, mint, maxt)
 		if err != nil {
-			return nil, errors.Wrapf(err, "open querier for head %s", rh)
+			return nil, errors.Wrapf(err, "open block querier for head %s", rh)
 		}
 
 		// Getting the querier above registers itself in the queue that the truncation waits on.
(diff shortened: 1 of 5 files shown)


tsdb/db.go
```diff
@@ -1628,30 +1628,59 @@ func (db *DB) Querier(_ context.Context, mint, maxt int64) (storage.Querier, err
...
+		outOfOrderHeadQuerier, err = NewBlockQuerier(rh, mint, maxt)
+		if err != nil {
+			return nil, errors.Wrapf(err, "open block querier for ooo head %s", rh)
+		}
+
+		// TODO(jesus.vazquez) I am not sure the truncation process affects the ooo head. Let's review this.
+		// Getting the querier above registers itself in the queue that the truncation waits on.
+		// So if the querier is currently not colliding with any truncation, we can continue to use it and still
+		// won't run into a race later since any truncation that comes after will wait on this querier if it overlaps.
+		shouldClose, getNew, newMint := db.head.IsQuerierCollidingWithTruncation(mint, maxt)
```
jesusvazquez, line 1672: I'm not sure we need this here @codesome

tsdb/head_append.go
```diff
@@ -536,6 +536,7 @@ func (a *headAppender) Commit() (err error) {
 			if delta <= series.oooAllowance {
 				// ... and the delta is within the OOO tolerance
 				var mmapRef chunks.ChunkDiskMapperRef
+				// TODO (jesus.vazquez) Should we track here new values for head ooo mint and maxt?
```
jesusvazquez, line 539: @codesome where should we update the new values for h.minOOOTime and h.maxOOOTime? I'm thinking ingestion and compaction but I could be wrong.

tsdb/ooo_head.go
```diff
@@ -33,17 +36,17 @@ func (oh *OOORangeHead) Chunks() (ChunkReader, error) {
 }
 
 func (oh *OOORangeHead) Tombstones() (tombstones.Reader, error) {
-	// TODO(jesus.vazquez) We still need to clarify what to do with tombstones
-	// here and in the design doc.
-	panic("implement me")
+	// As stated in the design doc https://docs.google.com/document/d/1Kppm7qL9C-BJB1j6yb6-9ObG3AbdZnFUBYPNNWwDBYM/edit?usp=sharing
+	// Tombstones are not supported for out of order metrics.
+	return noopTombstoneReader{}, nil
```
jesusvazquez, line 41: @codesome Please review this implementation, I could also return an error saying its not implemented but decided to return a noop tombstonereader instead. That way the caller is simply getting nothing so it can assume there are no ooo tombstones.

Please correct me if I'm wrong.

tsdb/head_append.go
jesusvazquez: This is done.

tsdb/db.go
jesusvazquez: We removed this.

tsdb/ooo_head.go
jesusvazquez: This is done

tsdb/db_test.go
```diff
@@ -3803,3 +3803,80 @@ func TestOOOCompaction(t *testing.T) {
...
+	opts.OOOCapMax = 30
+	opts.OOOAllowance = 24 * time.Hour.Milliseconds()
+	opts.AllowOverlappingBlocks = true
+
+	series1 := labels.FromStrings("foo", "bar1")
+
+	minutes := func(m int64) int64 { return m * time.Minute.Milliseconds() }
+	addSample := func(db *DB, fromMins, toMins int64) {
+		app := db.Appender(context.Background())
+		for min := fromMins; min <= toMins; min += time.Minute.Milliseconds() {
```
codesome, line 3819: I think you meant this?
```suggestion
		for min := fromMins*time.Minute.Milliseconds(); min <= toMins; min += time.Minute.Milliseconds() {
```

tsdb/db_test.go
```diff
@@ -3803,3 +3803,80 @@ func TestOOOCompaction(t *testing.T) {
...
+			db := openTestDB(t, opts, nil)
+			db.DisableCompactions()
+			defer func() {
+				require.NoError(t, db.Close())
+			}()
+			// Add in-order samples.
+			addSample(db, tc.inOrderMinT, tc.inOrderMaxT)
+
+			// Add out-of-order samples.
+			addSample(db, tc.oooMinT, tc.oooMaxT)
```
codesome, line 3865: I just had a thought that, similar to `samplesAppended` metric, we could have an additional metric `oooSamplesAppended` which will not only let us monitor the ooo ingestion but also cross check in the tests if samples are going in ooo Head. Not a blocker for this PR, but nice to add anyway if you have time.

tsdb/db_test.go
```diff
@@ -3803,3 +3803,80 @@ func TestOOOCompaction(t *testing.T) {
...
+
+			seriesSet := query(t, querier, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar1"))
+
+			require.NotNil(t, seriesSet[series1.String()])
+			expT := tc.queryMinT
+			for _, sample := range seriesSet[series1.String()] {
+				require.Equal(t, expT, sample.T())
+				require.Equal(t, float64(expT), sample.V())
+				expT += minutes(1)
+			}
```
codesome, line 3879: This only checks if the samples present in the result are correct. It does not check if any samples are missing from the result. A better way to test this might be to build the expected result map and just compare seriesSet with it.

tsdb/ooo_head.go
```diff
@@ -33,17 +35,17 @@ func (oh *OOORangeHead) Chunks() (ChunkReader, error) {
...
 }
 
 func (oh *OOORangeHead) Meta() BlockMeta {
 	var id [16]byte
 	copy(id[:], "____ooo_head____")
 	return BlockMeta{
-		MinTime: oh.head.MinTime(), // TODO(ganesh) We might want to track in the head whats the mint and maxt for out of order samples
-		MaxTime: oh.head.MaxTime(), // TODO(ganesh) We might want to track in the head whats the mint and maxt for out of order samples
+		MinTime: oh.head.MinOOOTime(),
+		MaxTime: oh.head.MaxOOOTime(),
```
codesome, line 48: Since this is a range head, it should be oh.mint and oh.maxt

codesome (commented):
3 blocking comments and 1 non blocking (about the new metric)

tsdb/db_test.go
jesusvazquez: I'm assuming fromMins and toMins are already minutes in milliseconds thats why I increase min by an entire minute in milliseconds on every iteration.

codesome: Oh right, I missed that you are using `minutes()` everywhere.

tsdb/db_test.go
```diff
@@ -3803,3 +3803,93 @@ func TestOOOCompaction(t *testing.T) {
...
+	series1 := labels.FromStrings("foo", "bar1")
+
+	minutes := func(m int64) int64 { return m * time.Minute.Milliseconds() }
+	addSample := func(db *DB, fromMins, toMins, queryMinT, queryMaxT int64, expSamples []tsdbutil.Sample) ([]tsdbutil.Sample, int) {
+		app := db.Appender(context.Background())
+		totalAppended := 0
+		for min := fromMins; min <= toMins; min += time.Minute.Milliseconds() {
+			_, err := app.Append(0, series1, min, float64(min))
+			if min >= queryMinT && min <= queryMaxT {
+				expSamples = appendSorted(expSamples, sample{t: min, v: float64(min)})
```
codesome, line 3823: More efficient way is to just build the samples and right before comparing with actual samples just sort it _once_. But since this is just a small test, no need for another commit.

codesome (approved):
Great work!

PR #224, opened 2022-04-26
Extend chunkquerier to read from the ooo head

jesusvazquez:
Mimir uses ChunkQuerier to read from the TSDB. In https://github.com/grafana/mimir-prometheus/pull/222 we forgot to extend ChunkQuerier to also read form the out of order head. Both PRs are pretty similar.

This one also adds unit tests.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -1701,11 +1701,11 @@ func (db *DB) ChunkQuerier(_ context.Context, mint, maxt int64) (storage.ChunkQu
 			blocks = append(blocks, b)
 		}
 	}
-	var headQuerier storage.ChunkQuerier
+	var inOrderHeadQuerier storage.ChunkQuerier
 	if maxt >= db.head.MinTime() {
 		rh := NewRangeHead(db.head, mint, maxt)
 		var err error
-		headQuerier, err = NewBlockChunkQuerier(rh, mint, maxt)
+		inOrderHeadQuerier, err = NewBlockChunkQuerier(rh, mint, maxt)
 		if err != nil {
 			return nil, errors.Wrapf(err, "open querier for head %s", rh)
 		}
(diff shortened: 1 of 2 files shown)


tsdb/db.go
```diff
@@ -1715,20 +1715,30 @@ func (db *DB) ChunkQuerier(_ context.Context, mint, maxt int64) (storage.ChunkQu
...
 		}
 	}
 
+	var outOfOrderHeadQuerier storage.ChunkQuerier
+	if overlapsClosedInterval(mint, maxt, db.head.MinOOOTime(), db.head.MaxOOOTime()) {
+		rh := NewOOORangeHead(db.head, mint, maxt)
+		var err error
+		outOfOrderHeadQuerier, err = NewBlockChunkQuerier(rh, mint, maxt)
+		if err != nil {
+			return nil, errors.Wrapf(err, "open block querier for ooo head %s", rh)
```
codesome, line 1738:
```suggestion
			return nil, errors.Wrapf(err, "open block chunk querier for ooo head %s", rh)
```

codesome (commented):
This PR as-is is looking good. I think you are waiting to write some unit tests for this?

tsdb/db_test.go
```diff
@@ -3886,10 +3890,95 @@ func TestOOOQuery(t *testing.T) {
...
+				return expSamples[i].T() < expSamples[j].T()
+			})
+
+			querier, err := db.ChunkQuerier(context.TODO(), tc.queryMinT, tc.queryMaxT)
+			require.NoError(t, err)
+			defer querier.Close()
+
+			chks := queryChunks(t, querier, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar1"))
+			require.NotNil(t, chks[series1.String()])
+			require.GreaterOrEqual(t, float64(oooSamples), prom_testutil.ToFloat64(db.head.metrics.outOfOrderSamplesAppended), "number of ooo appended samples mismatch")
```
codesome, line 3972: Why is it `>=` and not `===`?

tsdb/db_test.go
```diff
@@ -3886,10 +3890,95 @@ func TestOOOQuery(t *testing.T) {
...
+			sort.Slice(expSamples, func(i, j int) bool {
+				return expSamples[i].T() < expSamples[j].T()
+			})
+
+			querier, err := db.ChunkQuerier(context.TODO(), tc.queryMinT, tc.queryMaxT)
+			require.NoError(t, err)
+			defer querier.Close()
+
+			chks := queryChunks(t, querier, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar1"))
+			require.NotNil(t, chks[series1.String()])
```
codesome, line 3971: We should also check that len(chks) is 1. Same in Test_Querier_OOOQuery for seriesSet.

codesome (approved):
LGTM with a small comment

tsdb/db_test.go
jesusvazquez: You're right, fixed :+1:

PR #225, opened 2022-04-27
Set AllowOverlappingBlocks to true if ooo allowance is configured

jesusvazquez:
Relates to https://github.com/grafana/mimir-ooo/issues/15

When out of order is enabled, compaction can potentially create overlapping blocks with the in order blocks. Because of this we want to make sure that Prometheus's option `AllowOverlappingBlocks` is always true if OOOAllowance is set.

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -314,7 +314,7 @@ func main() {
 	serverOnlyFlag(a, "storage.tsdb.wal-compression", "Compress the tsdb WAL.").
 		Hidden().Default("true").BoolVar(&cfg.tsdb.WALCompression)
 
-	serverOnlyFlag(a, "storage.tsdb.ooo-allowance", "Allow upto this much out-of-order.  Supported units: h, m, s").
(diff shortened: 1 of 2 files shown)


cmd/prometheus/main.go
```diff
@@ -314,7 +314,7 @@ func main() {
 	serverOnlyFlag(a, "storage.tsdb.wal-compression", "Compress the tsdb WAL.").
 		Hidden().Default("true").BoolVar(&cfg.tsdb.WALCompression)
 
-	serverOnlyFlag(a, "storage.tsdb.ooo-allowance", "Allow upto this much out-of-order.  Supported units: h, m, s").
+	serverOnlyFlag(a, "storage.tsdb.ooo-allowance", "Allow samples to be this old for out-of-order.  Supported units: h, m, s. Also if the value is greater than AllowOverlappingBlocks's will be overrided and set to true since out-of-order head compaction can potentially create overlapping blocks.").
```
codesome, line 317: We should mention the flag name since this is user facing. And the data overlaps even before compaction.
```suggestion
	serverOnlyFlag(a, "storage.tsdb.ooo-allowance", "Allow samples to be this old for out-of-order.  Supported units: h, m, s. If the value is non-zero, then storage.tsdb.allow-overlapping-blocks will be set to true since out-of-order data can potentially overlap with other data.").
```

codesome (approved):

PR #226, opened 2022-04-28
Bunch of OOO Head enhancements and fixes

codesome:
This PR brings in multiple enhancements/fixes

1. Setting in-order mint/maxt of the Head properly.
2. Append ooo samples when no series exist for it in the Head. We take Head's maxt for the OOOAllowance check in that case.
3. Fix some m-map marker stuff. It was a TODO in the comment.

Unit tests exist for all of the fixes.

It will be easier to review 1 commit at a time instead of PR as a whole.

### DIFF
--- a/tsdb/head_append.go
+++ b/tsdb/head_append.go
@@ -479,6 +479,8 @@ func (a *headAppender) Commit() (err error) {
 		total            = len(a.samples)
 		oooTotal         = 0
 		oob, ooo, tooOld int   // out of bounds, out of order, too old
+		inOrderMint      int64 = math.MaxInt64
+		inOrderMaxt      int64 = math.MinInt64
 		ooomint          int64 = math.MaxInt64
 		ooomaxt          int64 = math.MinInt64
 		oooWblSamples    []record.RefSample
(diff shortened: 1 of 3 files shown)


tsdb/head_append.go
```diff
@@ -314,35 +318,49 @@ func (a *headAppender) Append(ref storage.SeriesRef, lset labels.Labels, t int64
...
-	if c == nil {
-		return 0, nil
+// If the returned boolean is true, then the sample belongs to the out of order chunk.
+func (s *memSeries) appendable(t int64, v float64, headMaxt, minValidTime int64) (bool, int64, error) {
+	msMaxt := s.maxTime()
+	if msMaxt == math.MinInt64 {
+		// The series has no sample and was freshly created.
+		if t >= minValidTime {
+			// We can append it in the in-order chunk.
+			return false, 0, nil
```
jesusvazquez, line 328: Just saw the comment above but still it was hard for me to understand this false statement. Maybe the boolean return parameter could be `appendableOutOfOrder`.

jesusvazquez (commented):
LGTM

tsdb/head_append.go
codesome: Will name it `isOutOfOrder`

jesusvazquez (approved):

PR #233, opened 2022-05-05
clarify OOO metrics and comments

Dieterbe:
In headAppender.Commit(), we always report any delta into the OOO
histogram, regardless of whether the sample was accepted or rejected,
and regardless of whether OOO is enabled or not.

In headAppender.Append(), we only reported the delta for OOO samples and
when OOO support was disabled.

The comprehensive behavior seems more useful, so switch to that.

+ cleanup exact overwrite case, don't increment the wrong metric

### DIFF
--- a/tsdb/head_append.go
+++ b/tsdb/head_append.go
@@ -287,19 +287,23 @@ func (a *headAppender) Append(ref storage.SeriesRef, lset labels.Labels, t int64
 	s.Lock()
 	// TODO: if we definitely know at this point that the sample is ooo, then optimise
 	// to skip that sample from the WAL and write only in the WBL.
-	if _, delta, err := s.appendable(t, v, a.headMaxt, a.minValidTime); err != nil {
-		s.Unlock()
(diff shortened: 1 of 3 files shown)


tsdb/head_append.go
```diff
@@ -597,29 +606,36 @@ func (a *headAppender) Commit() (err error) {
...
+				// sample timestamp is beyond any previously ingested timestamp
+				if s.T < inOrderMint { // TODO(ganesh): dieter thinks this never applies and can be removed because we know we're in order.
 					inOrderMint = s.T
 				}
 				if s.T > inOrderMaxt {
 					inOrderMaxt = s.T
 				}
 			} else {
-				total--
-				ooo++
```
Dieterbe, line 622: we were incrementing the wrong counter here.
as i was troubleshooting some issues with cortex-load-generator i confirmed that everytime we are in this section, it's always because the sample was duplicate wrt the 'highest ts' sample seen so far for that series.

PR #234, opened 2022-05-17
When removing start/end text op in regexp, don't break the contract.

colega:
We can't remove the only Sub from regexp, since the contract for some operations says that there's at least one Sub, like OpStar or OpPlus.

In order to convert a single-sub element into a no-op, we change the operation to OpEmptyString. You can see the same trick being used by standard regexp code when simplifying `x{0}`: https://github.com/golang/go/blob/cf7ec0fa098a46c3b75cc3d625f5d7528fe6e984/src/regexp/syntax/simplify.go#L41-L46

Relates to https://github.com/grafana/mimir/issues/1866

### DIFF
--- a/model/labels/regexp.go
+++ b/model/labels/regexp.go
@@ -163,6 +163,9 @@ func clearBeginEndText(re *syntax.Regexp) {
 	}
 	if len(re.Sub) == 1 {
 		if re.Sub[0].Op == syntax.OpBeginText || re.Sub[0].Op == syntax.OpEndText {
+			// We need to remove this element. Since it's the only one, we convert into a matcher of an empty string.
+			// OpEmptyMatch is regexp's nop operator.
+			re.Op = syntax.OpEmptyMatch
 			re.Sub = nil
 			return
 		}
--- a/model/labels/matcher_test.go
+++ b/model/labels/matcher_test.go
@@ -86,6 +86,16 @@ func TestMatcher(t *testing.T) {
 			value:   "foo-bar",
 			match:   false,
 		},
+		{
+			matcher: mustNewMatcher(t, MatchRegexp, "bar^+"),
+			value:   "foo-bar",
+			match:   false,
+		},
+		{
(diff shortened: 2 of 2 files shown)


56quarters (approved):
LGTM, nice fix!

model/labels/regexp.go
```diff
@@ -163,6 +163,9 @@ func clearBeginEndText(re *syntax.Regexp) {
 	}
 	if len(re.Sub) == 1 {
 		if re.Sub[0].Op == syntax.OpBeginText || re.Sub[0].Op == syntax.OpEndText {
+			// We need to remove this element. Since it's the only one, we convert into a matcher of an empty string.
+			// OpEmptyMatch is regexp's nop operator.
+			re.Op = syntax.OpEmptyMatch
```
pracucci, line 168: I think **theoretically**, the same issue could happen below, if `re.Sub` contains `[OpBeginText, OpEndText]` and after the cleanup we end up with `re.Sub = []` but practically I haven't found a way to trigger that case yet.

pracucci (approved):
Makes sense. See my non-comment tho :)

model/labels/regexp.go
colega: We could build `(^|$)` to hit that path, however I understand that it won't be an issue.

It's an issue here because some operators expect _exactly_ one sub:
```
	OpStar                         // matches Sub[0] zero or more times
	OpPlus                         // matches Sub[0] one or more times
	OpQuest                        // matches Sub[0] zero or one times
	OpRepeat                       // matches Sub[0] at least Min times, at most Max (Max == -1 is no limit)
```

However, if we have more than one sub, then it's none of these, and there are no operators matching _at least 1 sub_ or similar, actually this path (if I understand correctly) has to be one of these operators:

```
	OpConcat                       // matches concatenation of Subs
	OpAlternate                    // matches alternation of Subs
```

Which are (should be?) perfectly fine having an empty subs set.

PR #235, opened 2022-05-23
Add an option to enable overlapping compaction separately with overlapping queries

codesome:
This is so that Mimir ingesters don't compact overlapping blocks and let the compactor handle it


codesome:
I will port this PR in Mimir before merging this to test if disabling overlapping compaction fixes the query bug

cmd/prometheus/main.go
```diff
@@ -1535,7 +1535,8 @@ func (opts tsdbOptions) ToTSDBOptions() tsdb.Options {
 		RetentionDuration:              int64(time.Duration(opts.RetentionDuration) / time.Millisecond),
 		MaxBytes:                       int64(opts.MaxBytes),
 		NoLockfile:                     opts.NoLockfile,
-		AllowOverlappingBlocks:         opts.AllowOverlappingBlocks,
+		AllowOverlappingQueries:        opts.AllowOverlappingBlocks,
```
jesusvazquez, line 1538:
```suggestion
		AllowOverlappingQueries:        opts.AllowOverlappingQueries,
```

jesusvazquez: Am I missing something here or the initialization should be changed like in my suggestion?

jesusvazquez (changes requested):
I believe there is a typo in the main.go file, please verify. Other than that LGTM

cmd/prometheus/main.go
codesome: The user facing flag does not change, because it's a breaking change. The end user does not need this granular controls. They either enable it both, or disable it both. The option is only for TSDB for the library users to use. So it's fine here.

Issue #237, opened 2022-05-26
Write comprehensive unit tests for out of order ingestion


codesome:
- [x] Appending samples out of order and out of bound and working properly. With proper on-disk artifacts.
- [x] Querying of ooo samples working properly.
	- [x] Querying only ooo
	- [x] Querying ooo and in-order
	- [x] Query including difference scenerios of possible chunk arrangement in the head block
- [x] Compaction happening for blocks properly with multiple blocks output where required
	- [x] M-map files are cleared properly
	- [x] WBL cleared properly
	- [x] Querying works properly after compaction
		- [x] With level 1 blocks before overlapping compaction
		- [x] After overlapping compaction
	- [x] OOO Compaction runs right after normal Head's compaction and does not trigger if normal head is not compacted
- [x] When compaction fails multiple times, m-map files and wbl is still cleared properly only after a successful compaction
- [x] WBL/M-map replay happening properly
	- [x] When there is only WBL
		- [x] With m-map markers as well
	- [x] When there is only M-map
	- [x] When WBL and M-map both exist with and without m-map markers
	- [x] WBL+M-map replay with mmap markers when the ooomax size changes between replay.
		- [x] Size increases
		- [x] Size decreases
	- [x] Corruption in WBL and replay
	- [x] Corruption in M-map and replay
- [x] No OOO artifacts are created when OOOAllowance is 0
	- [x] Including, no in-memory ooo chunk is created when ooo sample is discarded in this case
	- [x] Old metrics are incremented and not the new ones

PR #239, opened 2022-05-27
Add TestOOOWALAndMmapReplay for test various restart scenarios

codesome:
Takes care of following cases from https://github.com/grafana/mimir-prometheus/issues/237

- [ ] When there is only WBL
	- [ ] With m-map markers as well
- [ ] When there is only M-map
- [ ] When WBL and M-map both exist with and without m-map markers
- [ ] WBL+M-map replay with mmap markers when the ooomax size changes between replay.
	- [ ] Size increases
	- [ ] Size decreases


jesusvazquez (approved):
Really nice :+1: I think I got it all. I'll leave only one suggestion!

tsdb/db_test.go
```diff
@@ -3997,3 +3997,175 @@ func Test_ChunkQuerier_OOOQuery(t *testing.T) {
...
+	require.NoError(t, fileutil.CopyDirs(mmapDir, originalMmapDir))
+	resetWBLToOriginal := func() {
+		require.NoError(t, os.RemoveAll(wblDir))
+		require.NoError(t, fileutil.CopyDirs(originalWblDir, wblDir))
+	}
+	resetMmapToOriginal := func() {
+		require.NoError(t, os.RemoveAll(mmapDir))
+		require.NoError(t, fileutil.CopyDirs(originalMmapDir, mmapDir))
+	}
+	{ // 1. Restart DB with both WBL and M-map files for ooo data.
```
jesusvazquez, line 4091: Instead of these anonymous functions, perhaps we could do the usual testing function here and use the description in the comment as the test name?

PR #240, opened 2022-06-01
Improve test coverage for out-of-order head's compaction

codesome:
Covers the following points in https://github.com/grafana/mimir-prometheus/issues/237

* M-map files are cleared properly
* Querying works properly after compaction
  * With level 1 blocks before overlapping compaction
  * After overlapping compaction
* OOO Compaction runs right after normal Head's compaction and does not trigger if normal head is not compacted


jesusvazquez (approved):
:woman_dancing:

PR #244, opened 2022-06-02
Add TestOOOCompactionFailure to test compaction failure

codesome:
Covers this point from https://github.com/grafana/mimir-prometheus/issues/237
*  When compaction fails multiple times, m-map files and wbl is still cleared properly only after a successful compaction

The test makes sure that a failed compaction does not leave the
OOO Head corrupted. And eventually when a compaction works, the
blocks are created properly and the m-map files, WBL files and
the OOO Head are cleared properly.


jesusvazquez (approved):
wow okey that was a very hidden one

PR #247, opened 2022-06-03
Job queue

pstibrany:
This PR reimplements `chan chunkWriteJob` with custom buffered queue that should use less memory, because it doesn't preallocate entire buffer for maximum queue size at once. Instead it allocates individual "segments" with smaller size.

As elements are added to the queue, they fill individual segments. When elements are removed from the queue (and segments), empty segments can be thrown away. This doesn't change memory usage of the queue when it's full, but should decrease its memory footprint when it's empty (queue will keep max 1 segment in such case).

(Going to test the change in our dev cluster)

### DIFF
--- a/tsdb/chunks/chunk_write_queue.go
+++ b/tsdb/chunks/chunk_write_queue.go
@@ -29,6 +29,9 @@ const (
 
 	// Minimum interval between shrinking of chunkWriteQueue.chunkRefMap.
 	chunkRefMapMinShrinkInterval = 10 * time.Minute
+
+	// With chunkWriteJob being 64 bytes, this will use ~512 KiB for empty queue.
+	chunkQueueSegmentSize = 8192
 )
 
 type chunkWriteJob struct {
@@ -45,7 +48,7 @@ type chunkWriteJob struct {
 // Chunks that shall be written get added to the queue, which is consumed asynchronously.
 // Adding jobs to the queue is non-blocking as long as the queue isn't full.
 type chunkWriteQueue struct {
-	jobs chan chunkWriteJob
+	jobs *writeJobQueue
 
 	chunkRefMapMtx        sync.RWMutex
 	chunkRefMap           map[ChunkDiskMapperRef]chunkenc.Chunk
@@ -84,7 +87,7 @@ func newChunkWriteQueue(reg prometheus.Registerer, size int, writeChunk writeChu
 	)
 
 	q := &chunkWriteQueue{
-		jobs:                  make(chan chunkWriteJob, size),
+		jobs:                  newWriteJobQueue(size, chunkQueueSegmentSize),
 		chunkRefMap:           make(map[ChunkDiskMapperRef]chunkenc.Chunk),
 		chunkRefMapLastShrink: time.Now(),
 		writeChunk:            writeChunk,
@@ -108,7 +111,12 @@ func (c *chunkWriteQueue) start() {
 	go func() {
 		defer c.workerWg.Done()
 
-		for job := range c.jobs {
+		for {
+			job, ok := c.jobs.pop()
+			if !ok {
+				return
+			}
+
 			c.processJob(job)
 		}
 	}()
@@ -191,7 +199,14 @@ func (c *chunkWriteQueue) addJob(job chunkWriteJob) (err error) {
 	}
 	c.chunkRefMapMtx.Unlock()
 
-	c.jobs <- job
+	ok := c.jobs.push(job)
+	if !ok {
+		c.chunkRefMapMtx.Lock()
+		delete(c.chunkRefMap, job.ref)
+		c.chunkRefMapMtx.Unlock()
+
+		return errors.New("queue is closed")
+	}
 
 	return nil
 }
@@ -218,7 +233,7 @@ func (c *chunkWriteQueue) stop() {
 
 	c.isRunning = false
 
-	close(c.jobs)
+	c.jobs.close()
 
 	c.workerWg.Wait()
 }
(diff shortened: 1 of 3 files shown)


tsdb/chunks/queue.go
```diff
@@ -0,0 +1,132 @@
...
+// of equal size (segmentSize). When segment is not used anymore, reference to it are removed,
+// so it can be treated as a garbage.
+type writeJobQueue struct {
+	maxSize     int
+	segmentSize int
+
+	mtx         sync.Mutex            // protects all following variables
+	pushed      *sync.Cond            // signalled when something is pushed into the queue
+	popped      *sync.Cond            // signalled when element is popped from the queue
+	first, last *writeJobQueueSegment // pointer to first and last segment, if any
```
replay, line 16: that's interesting, i haven't seen this syntax before where you define to properties in one line like this. It makes sense here, but if we use that here why not also for `pushed` and `popped`?

tsdb/chunks/queue.go
```diff
@@ -0,0 +1,132 @@
...
+			if q.first == nil {
+				q.first = q.last
+			}
+		}
+
+		q.last.queue = append(q.last.queue, job)
+		q.size++
+		q.pushed.Signal()
+		return true
+	}
```
replay, line 88: Does all of this need to be inside the loop, wouldn't something like this be sufficient:

```suggestion
	if q.closed {
		return false
	}

	for {
		if q.size >= q.maxSize {
			// wait until queue has more space or is closed
			q.popped.Wait()
			continue
		}
	}

	// cap(q.last.queue)-len(q.last.queue) is free space remaining in the q.last.queue.
	if q.last == nil || cap(q.last.queue)-len(q.last.queue) == 0 {
		prevLast := q.last
		q.last = &writeJobQueueSegment{
			queue: make([]chunkWriteJob, 0, q.segmentSize),
		}

		if prevLast != nil {
			prevLast.next = q.last
		}
		if q.first == nil {
			q.first = q.last
		}
	}

	q.last.queue = append(q.last.queue, job)
	q.size++
	q.pushed.Signal()
	return true
```

tsdb/chunks/queue.go
pstibrany: Sure, let's do that. I will merge their comments.

tsdb/chunks/queue.go
pstibrany: You're right!

pstibrany:
@replay Thanks for your review. I've addressed your comments. I've also modified implementation to avoid using `len` and `cap` and use two indexes instead (index of next read for `pop`, and next write for `push`), I think it makes code easier to read and understand.

I would appreciate if you had a chance to look at the PR again.

tsdb/chunks/queue.go
```diff
@@ -0,0 +1,132 @@
...
+	defer q.mtx.Unlock()
+
+	for {
+		if q.closed {
+			return false
+		}
+
+		if q.size >= q.maxSize {
+			// wait until queue has more space or is closed
+			q.popped.Wait()
```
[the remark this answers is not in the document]
replay: Oh right, I need to read the docs of `cond.Wait` again 🤦‍♂️

tsdb/chunks/queue.go
```diff
@@ -0,0 +1,132 @@
...
+	closed      bool                  // after closing the queue, nothing can be pushed to it
+}
+
+type writeJobQueueSegment struct {
+	queue []chunkWriteJob
+	next  *writeJobQueueSegment
+}
+
+func newWriteJobQueue(maxSize, segmentSize int) *writeJobQueue {
+	if maxSize <= 0 || segmentSize <= 0 {
```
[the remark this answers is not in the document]
pstibrany:
```
┌────────────────────────────────────────────┐
│                                            │
│                  segment                   │
│                                            │
└────────────────────────────────────────────┘
      ▲                              ▲        
      │                              │        
      │                              │        
┌──────────┐                   ┌──────────┐   
│ nextRead │                   │nextWrite │   
└──────────┘                   └──────────┘   
```

Let's imagine that both `nextRead` and `nextWrite` move towards right. Adding new element to the queue = moving `nextWrite` by one element. Popping element from the queue = moving `nextRead` by one element. Once `nextRead` reaches the end, we drop the segment.

When size == 0, `nextRead` will be equal to `nextWrite`, but another `push` will use another element (`nextWrite`).

If queue is still used (pushed to / popped from), eventually segment will be used fully, even if it's much bigger than `maxSize`.

tsdb/chunks/queue.go
```diff
@@ -0,0 +1,126 @@
...
+	defer q.mtx.Unlock()
+
+	// wait until something is pushed to the queue, or queue is closed.
+	for q.size == 0 {
+		if q.closed {
+			return chunkWriteJob{}, false
+		}
+
+		q.pushed.Wait()
+	}
```
replay, line 103: Do I understand it correctly that if there are elements in the queue and then the queue gets closed, these elements that are already in it can still be popped until the queue is empty?

What happens if `pop()` is called on an empty queue and blocks on `q.pushed.Wait()`, but then the queue gets closed and `q.pushed.Wait()` unblocks? Wouldn't that panic on `:105` because `q.first` would be `nil`?

replay: Ah right, never mind me :) thanks!

replay (approved):
This looks great, thanks!

pstibrany:
Thanks for your review @replay!

tsdb/chunks/chunk_write_queue.go
```diff
@@ -230,7 +251,7 @@ func (c *chunkWriteQueue) queueIsEmpty() bool {
 func (c *chunkWriteQueue) queueIsFull() bool {
 	// When the queue is full and blocked on the writer the chunkRefMap has one more job than the cap of the jobCh
 	// because one job is currently being processed and blocked in the writer.
```
pracucci, line 253: This comment should be updated too.

pracucci: I think comment in `queueSize()` needs to be updated too.

pstibrany: Oh, thanks for catching this. Will send new PR.

PR #251, opened 2022-06-09
Add an option to enable overlapping compaction separately with overlapping queries

codesome:
This PR bring https://github.com/grafana/mimir-prometheus/pull/235 into main.

With this change, we can selectively allow overlapping blocks to exists and be able to query form them, while not doing any compaction on them. This will let Mimir ingesters to not error on overlapping blocks and not do the merging of blocks in the ingesters so that compactor can take care of compaction later.


jesusvazquez (approved):
Carefully reviewed the PR and looks good to me. 

Special attention at the changes in db.go where now if `AllowOverlappingQueries` is true the tsdb won't fail to open if it find overlapping blocks.

pstibrany (approved):
LGTM from what I understand.

pstibrany:
We want to enable querying of overlapping blocks (due to out-of-order ingestion), but not compaction, to not confuse Mimir's shipper. This PR makes it possible.

PR #267, opened 2022-06-17
WaitForIdle only sleep within select

jesusvazquez:
Before everytime within the loop we would sleep for 10 microseconds, now
we're only sleeping if output has drained but there are still input
samples.

@ywwg came up with this idea in https://raintank-corp.slack.com/archives/C02ME6ZS2NQ/p1655390202442689?thread_ts=1655372958.069319&cid=C02ME6ZS2NQ

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -722,7 +722,12 @@ func (h *Head) Init(minValidTime int64) error {
 		}
 		h.startWALReplayStatus(startFrom, endAt)
 
+		replayStart := time.Now()
+		segmentCount := 0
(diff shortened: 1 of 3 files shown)


jesusvazquez:
Leaving this one as draft since we want to properly test and get this merged first https://github.com/prometheus/prometheus/pull/10878

ywwg:
seems like this is probably not needed, so I guess let's close it :)

codesome:
We can use a similar technique to https://github.com/prometheus/prometheus/pull/10973 to avoid that sleep

PR #268, opened 2022-06-20
Add out-of-order sample support in TSDB

codesome:
This is a placeholder PR to see how can we break down into smaller PRs.

**DO NOT MERGE**

cc @jesusvazquez

PR #269, opened 2022-06-21
Add out-of-order sample support to the TSDB

codesome:
This implementation is based on this design doc:
https://docs.google.com/document/d/1Kppm7qL9C-BJB1j6yb6-9ObG3AbdZnFUBYPNNWwDBYM/edit?usp=sharing

This commit adds support to accept out-of-order ("OOO") sample into the TSDB
up to a configurable time allowance. If OOO is enabled, overlapping querying
are automatically enabled.

### DIFF
--- a/tsdb/head_read.go
+++ b/tsdb/head_read.go
@@ -212,9 +212,18 @@ func (h *headIndexReader) Series(ref storage.SeriesRef, lbls *labels.Labels, chk
 	return nil
 }
 
-// headChunkID returns the HeadChunkID corresponding to .mmappedChunks[pos]
+// headChunkID returns the HeadChunkID referred to by the given position.
+// * 0 <= pos < len(s.mmappedChunks) refer to s.mmappedChunks[pos]
+// * pos == len(s.mmappedChunks) refers to s.headChunk
 func (s *memSeries) headChunkID(pos int) chunks.HeadChunkID {
 	return chunks.HeadChunkID(pos) + s.firstChunkID
+}
+
+// oooHeadChunkID returns the HeadChunkID referred to by the given position.
+// * 0 <= pos < len(s.oooMmappedChunks) refer to s.oooMmappedChunks[pos]
+// * pos == len(s.oooMmappedChunks) refers to s.oooHeadChunk
+func (s *memSeries) oooHeadChunkID(pos int) chunks.HeadChunkID {
+	return chunks.HeadChunkID(pos) + s.firstOOOChunkID
 }
 
 // LabelValueFor returns label value for the given label name in the series referred to by ID.
@@ -287,8 +296,8 @@ func (h *headChunkReader) Close() error {
 }
 
 // Chunk returns the chunk for the reference number.
-func (h *headChunkReader) Chunk(ref chunks.ChunkRef) (chunkenc.Chunk, error) {
-	sid, cid := chunks.HeadChunkRef(ref).Unpack()
+func (h *headChunkReader) Chunk(meta chunks.Meta) (chunkenc.Chunk, error) {
+	sid, cid := chunks.HeadChunkRef(meta.Ref).Unpack()
 
 	s := h.head.series.getByID(sid)
 	// This means that the series has been garbage collected.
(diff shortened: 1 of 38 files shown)


codesome:
This replaces https://github.com/grafana/mimir-prometheus/pull/268. Somehow having a single commit while manually adding co-authors removes the CLA requirement from the co-authors. (See 268, CLA fails because Dieter hasn't signed, but it does not fail here).

codesome:
We can consider merging this after https://github.com/grafana/mimir-prometheus/pull/270/ is merged and synced in this PR.

jesusvazquez:
Lets get the OOOAllowance and the OOO WBL dir rename into this PR.

tsdb/head_read.go
```diff
@@ -358,6 +367,259 @@ func (s *memSeries) chunk(id chunks.HeadChunkID, cdm chunkDiskMapper) (chunk *me
...
+
+func (o mergedOOOChunks) NumSamples() int {
+	samples := 0
+	for _, c := range o.chunks {
+		samples += c.Chunk.NumSamples()
+	}
+	return samples
+}
+
+// Compact TODO(jesus.vazquez) Clarify why this method wont be called
```
jesusvazquez, line 544: Maybe we can remove this

tsdb/index/postings.go
```diff
@@ -49,8 +49,12 @@ var ensureOrderBatchPool = sync.Pool{
 // EnsureOrder() must be called once before any reads are done. This allows for quick
 // unordered batch fills on startup.
 type MemPostings struct {
-	mtx     sync.RWMutex
-	m       map[string]map[string][]storage.SeriesRef
+	mtx sync.RWMutex
+	m   map[string]map[string][]storage.SeriesRef
+	// TODO We should have a mapping of which series have out of order metrics
```
jesusvazquez, line 54: I don't think we need this anymore?

tsdb/index/postings.go
```diff
@@ -216,6 +220,8 @@ func (p *MemPostings) Get(name, value string) Postings {
 	return newListPostings(lp...)
 }
 
+// TODO GetOOOPostings( returns newOOOListPostings )
```
jesusvazquez, line 223: Remove this too

tsdb/ooo_head.go
```diff
@@ -0,0 +1,76 @@
...
+		Stats: BlockStats{
+			NumSeries: oh.head.NumSeries(),
+		},
+	}
+}
+
+// Size returns 0 because the space taken by the out of order samples is taken
+// into account by the RangeHead size.
+func (oh *OOORangeHead) Size() int64 {
+	// TODO(jesus.vazquez) Find what's the appropriate value here
```
jesusvazquez, line 59: We never solved this, i think its also not called. WDYT @codesome

jesusvazquez (commented):
I did take a look at some of the TODO comments that we can get rid of. Can you take a look at the comments @codesome?

tsdb/ooo_head.go
codesome: Oh, that makes me realise that we are not including the size of WBL anywhere. I will fix that.

codesome:
Opened https://github.com/grafana/mimir-prometheus/pull/273 for TODOs

jesusvazquez (approved):
Preemptively approving knowing that you will rebase this PR with the changes from https://github.com/grafana/mimir-prometheus/pull/273

Let's goooo :tada: :tada: :tada: :tada:

nwpuCfy:
> This PR changes `ChunkReader` interface in a way that's incompatible with usage in Thanos and Mimir. We could fix Mimir, but unless you plan to vendor `mimir-prometheus` in Thanos, it needs to be reverted :(

Did this mean TSDB that supports out of order data is not compatible with thanos? So I can't use it in cortex v1.11 that supports obs storage?

codesome:
@nwpuCfy that comment was about code interface and not the data format. Thanos has switched to the new interface I think. And they use prometheus/prometheus, and not grafana/mimir-prometheus.

jesusvazquez:
@nwpuCfy As @codesome pointed out, thanos is now vendoring prometheus/prometheus were we ported the changes you see in this PR so there should not be any problems :+1:  Use prometheus/prometheus as your reference for the official state of out-of-order support.

nwpuCfy:
> @nwpuCfy As @codesome pointed out, thanos is now vendoring prometheus/prometheus were we ported the changes you see in this PR so there should not be any problems 👍 Use prometheus/prometheus as your reference for the official state of out-of-order support.
ok, thanks for reply

nwpuCfy:
> @nwpuCfy that comment was about code interface and not the data format. Thanos has switched to the new interface I think. And they use prometheus/prometheus, and not grafana/mimir-prometheus.

thanks for reply

PR #272, opened 2022-06-22
Remove the hack in config for OutOfOrderAllowance

codesome:
Removes the confusion over how the .Milliseconds() in the TSDB is used for OutOfOrderAllowance


jesusvazquez (approved):
LGTM find this less confusing than the user having to understand the unit he was translating the duration from. Less load for the user.

PR #275, opened 2022-06-24
Update MinOOOTime and MaxOOOTime properly after restart

codesome:
We were not updating the MinOOOTime and MaxOOOTime after a replay. The tests did not catch it before because the default values made the querier pick the ooo head. But if any sample comes in, it is possible it can skip ooo head in valid cases. Now in the tests I am explicitly checking for those values to be correct.


jesusvazquez (approved):
LGTM smart fix

PR #276, opened 2022-06-24
Rename OutOfOrderAllowance to OutOfOrderTimeWindow

jesusvazquez:
After review Allowance is perhaps a bit misleading so we've decided to
replace it with a more common term like TimeWindow.


codesome (commented):
Nice! Only 3 places remaining
```
db_test.go:5128:		// Allowance to 0, hence disabled.
db_test.go:5160:		// Allowance to 0.
head.go:221:	// Allowance can be set on runtime. So the capMin and capMax should be valid
```

codesome (approved):

PR #277, opened 2022-06-27
Avoid gaps in in-order data after restart with out-of-order enabled

codesome:
Fixes https://github.com/grafana/mimir/issues/2238

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -842,7 +842,7 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
 	// We do not consider blocks created from out-of-order samples for Head's minValidTime
 	// since minValidTime is only for the in-order data and we do not want to discard unnecessary
 	// samples from the Head.
-	inOrderMaxTime, ok := db.inOrderBlocksMaxTime(opts.MinBlockDuration)
+	inOrderMaxTime, ok := db.inOrderBlocksMaxTime()
 	if ok {
 		minValidTime = inOrderMaxTime
 	}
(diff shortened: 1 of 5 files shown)


tsdb/db.go
```diff
@@ -1636,6 +1639,30 @@ func (db *DB) Blocks() []*Block {
...
+		}
+	}
+	if !hasOOO {
+		// Temporary patch. To be removed by mid July 2022.
+		// Before this patch, blocks did not have "out_of_order" in their meta, so we cannot
+		// say which block has the out_of_order data. In that case the out-of-order block can be
+		// up to 2 block ranges ahead of the latest in-order block.
+		// Note: if hasOOO was true, it means the latest block has the new meta and is taken care in inOrderBlocksMaxTime().
+		maxt -= 2 * minBlockDuration
+	}
```
jesusvazquez, line 1662: Nice addition :+1:

tsdb/db.go
```diff
@@ -1636,6 +1639,30 @@ func (db *DB) Blocks() []*Block {
...
+			maxt = b.meta.MaxTime
+		}
+	}
+	if !hasOOO {
+		// Temporary patch. To be removed by mid July 2022.
+		// Before this patch, blocks did not have "out_of_order" in their meta, so we cannot
+		// say which block has the out_of_order data. In that case the out-of-order block can be
+		// up to 2 block ranges ahead of the latest in-order block.
+		// Note: if hasOOO was true, it means the latest block has the new meta and is taken care in inOrderBlocksMaxTime().
+		maxt -= 2 * minBlockDuration
```
pstibrany, line 1661: If there was no block, this return `math.MinInt64 - 2* minBlockDuration`, which will underflow and become very large positive number.

How does this logic (subtracting `2 * minBlockDuration`) work with `Truncate`?

codesome: There is a `ok` variable to take care of it :) I can use it here to avoid confusion

PR #283, opened 2022-07-04
Remove temporary patch for out-of-order

codesome:
This is no longer required


jesusvazquez (approved):
LGTM! We also need to remove the rename of ooo_wbl to just wbl. Want to do it in this PR or I do it on a different one?

codesome:
We can do it in this PR. On it.

Issue #284, opened 2022-07-04
Simplify the concept of `OutOfOrderTimeWindow`


codesome:
## Problem

Currently, here is how the `OutOfOrderTimeWindow` is interpreted when there is a matching time series or not.

Taking example of `OutOfOrderTimeWindow`

**Case 1**: *If there is a series in the Head block for the sample*, whose max time is let's say `SeriesMaxT`, then we accept out-of-order samples for this series that has timestamp greater than or equal to `SeriesMaxT-OutOfOrderTimeWindow`, no other logic.

**Case 2**: *If a sample does not have any series in the Head block*, we don't have a `SeriesMaxT` to compare with. So we take the Head's maxt (call it `HeadMaxT`). So the ingestable window for the "out-of-order" sample becomes `HeadMaxT-OutOfOrderTimeWindow`. There is another caveat here, more like 2 sub-cases

  * Case 2.1: If the sample timestamp was after `HeadMaxT-1h`, it is still ingestable as an in-order sample. And it goes in the in-order Head. Hence if `OutOfOrderTimeWindow` was less than 1h, it technically has no use if the series does not exist.
  * Case 2.1: If the sample timestamp was before `HeadMaxT-1h`, it cannot go in the in-order Head. Only then the `HeadMaxT-OutOfOrderTimeWindow` is applicable, and only makes sense if `OutOfOrderTimeWindow` was greater than 1h.

Now here is one example of irregularity/confusion with having these 2 cases. 

`OutOfOrderTimeWindow=30m`. Let's say currently `HeadMaxT=10h`. And there is an incoming sample with `T=8h40m`, and the sample has no series in the Head.

Now this sample will be rejected because it doesn't fit in both `T >= HeadMaxT-1h` and `T >= HeadMaxT-OutOfOrderTimeWindow` according to the case 2 above.

BUT, if by chance there was a sample for that series that came right before with `T2=9h1m`, then it would go in the in-order Head and make `SeriesMaxT=9h1m`. And since the case 1 applies now, the earlier sample of `T=8h40m` is suddenly acceptable.

## Proposal

An updated version of **case 2**.

If the series does not exist, then it is same as case 2 above. If the series exist, and if the sample is out of order, then we check if the sample timestamp is after `HeadMaxT-OutOfOrderTimeWindow`, hence removing `SeriesMaxT` completely out of the picture.

[...]

PR #285, opened 2022-07-07
Simplify how OutOfOrderTimeWindow works

codesome:
This PR tries to simplify the logic around working of the `OutOfOrderTimeWindow`.

Current logic: If there is a series, time window is w.r.t. series' maxt. If the series does not exist, then it is w.r.t. the head's maxt. Problems with this is described in more detail in https://github.com/grafana/mimir-prometheus/issues/284

New solution: out of order time window now will always be taken w.r.t. the head's maxt. Making the logic as "if the sample cannot go in as in-order sample, then it should be within OutOfOrderTimeWindow from TSDB's maxt to inside out-of-order chunk".

This simple of logic of pinning it to a single time (i.e. head's maxt) will also make work with caches easier.

(This PR also removes 2 flags that are not required at Prometheus level)

### DIFF
--- a/cmd/prometheus/main.go
+++ b/cmd/prometheus/main.go
@@ -314,12 +314,6 @@ func main() {
 	serverOnlyFlag(a, "storage.tsdb.wal-compression", "Compress the tsdb WAL.").
 		Hidden().Default("true").BoolVar(&cfg.tsdb.WALCompression)
 
-	serverOnlyFlag(a, "storage.tsdb.out-of-order-cap-min", "Minimum capacity for out of order chunks (in samples. between 0 and 255.)").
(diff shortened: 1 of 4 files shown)


codesome:
cc @bboreham @pracucci @pstibrany if you have any opinions against this change

codesome:
It will be easier to read the change if you first look at how it works for in-order in the upstream ([here](https://github.com/prometheus/prometheus/blob/27559e7b4fb3dc9853927757fe1640b219e1832b/tsdb/head_append.go#L303-L322)). In-order logic is almost same.

jesusvazquez (approved):
These changes look good to me. I think it simplifies the issue and now we only need to remember `[headmaxt-oootimewindow, headmaxt]`

PR #287, opened 2022-07-13
Update upstream prometheus

pstibrany:
Merge upstream Prometheus from https://github.com/prometheus/prometheus/commit/44fcf876caad9a5d28b92a728d2c98c34015d377 into mimir-prometheus.

TO BE MERGED WITH MERGE COMMIT.


pracucci (approved):
I reviewed all changes except `web/ui`. The errors saga continues. I think that's the only change that could impact Mimir, in particular if we use `errors.Cause()` on any error returned by Prometheus vendored code.

PR #297, opened 2022-07-14
Merge prometheus/sparsehistogram into mimir-prometheus/sparsehistogram

LeviHarrison:
We're still working everything out, this PR is just here to make commenting easier.


model/textparse/openmetricsparse_test.go
```diff
@@ -237,9 +237,7 @@ foo_total 17.0 1520879607.789 # {xx="yy"} 5`
 			p.Metric(&res)
 			found := p.Exemplar(&e)
 			require.Equal(t, exp[i].m, string(m))
-			if e.HasTs {
```
codesome, line 240: Any reason this `if` check was removed?

LeviHarrison: Change came from the sparsehistogram branch (https://github.com/prometheus/prometheus/pull/9069), maybe @beorn7 has an answer?

codesome:
I merged the PR (not squash) to unblock since CI was green. I will do a post merge review (it is gonna take some time to verify all, it's a huge PR)

model/textparse/openmetricsparse_test.go
beorn7: I don't remember, but this looks like it was a bug previously. Why would I ask if the exemplar has a timestamp and then test the one returned by the `Series` call?

PR #309, opened 2022-08-02
Update minOOOTime after truncating Head

codesome:


tsdb/head.go
```diff
@@ -1234,31 +1211,48 @@ func (h *Head) truncateOOO(lastWBLFile int, minOOOMmapRef chunks.ChunkDiskMapper
...
+			h.minValidTime.Store(appendableMinValidTime)
 		}
 	}
+	if headMaxt-h.opts.OutOfOrderTimeWindow.Load() < minOOOTime {
+		// The allowed OOO window is lower than the min OOO time seen during GC.
+		// So it is possible that some OOO sample was inserted that was less that minOOOTime.
+		// So we play safe and set it to the min that was possible.
+		minOOOTime = headMaxt - h.opts.OutOfOrderTimeWindow.Load()
+	}
+	h.minOOOTime.Store(minOOOTime)
```
codesome, line 1249: This is the only new thing in this function. Everything else is the same. I took it out in its own function because we were doing the same thing in 2 places and this logic is a bit fragile to duplicate and regress.

jesusvazquez (approved):

PR #311, opened 2022-08-08
Remove old chunk mapper.

pstibrany:
This PR removes `OldChunkDiskMapper`


pracucci (approved):

pstibrany:
@codesome  I've removed old chunk mapper in this PR, but now `TestOOOCompaction` and `TestOOOCompactionFailure` are failing. It turns out that old chunk mapper was used by default instead of new chunk mapper (which is also in upstream Prometheus).

Would you please take a look at the failing tests? There seem to be small difference between old and new chunk mapper. I'll try to look at it as well, but you may have better idea about the differences between old and new chunk mapper (perhaps just the fact that new one is async is enough change to cause this).

pstibrany:
> @codesome I've removed old chunk mapper in this PR, but now `TestOOOCompaction` and `TestOOOCompactionFailure` are failing. It turns out that old chunk mapper was used by default instead of new chunk mapper (which is also in upstream Prometheus).
> 
> Would you please take a look at the failing tests? There seem to be small difference between old and new chunk mapper. I'll try to look at it as well, but you may have better idea about the differences between old and new chunk mapper (perhaps just the fact that new one is async is enough change to cause this).
> 
> Failures: https://github.com/grafana/mimir-prometheus/runs/7728655169?check_suite_focus=true

Fix is likely available in https://github.com/prometheus/prometheus/pull/11075.

codesome:
https://github.com/prometheus/prometheus/pull/11075 is same as the one in mimir-prometheus, and tests might fail for async queue there as well I think. In Prometheus the queue is disabled by default I believe. I will take a look at those tests in coming days.

codesome:
The last commit should fix the tests. I will merge on green (minus the unrelated lint failure)

pstibrany:
Thanks @codesome, much appreciated! I can take care of updating mimir-prometheus in Mimir as a next step.

codesome:
Another test failure because of some sorting issue. On it.

PR #312, opened 2022-08-09
Prevent panic with ApplyConfig

codesome:
We now replay WBL even if OOO is disabled, in which case WBL might be already initialised. This was not accounted for in ApplyConfig.


jesusvazquez (approved):
Good catch

codesome:
Merging without fixing lint. Lint failure is on an unrelated file.

PR #314, opened 2022-08-10
Fix OOO Head LabelValues and PostingsForMatchers

jesusvazquez:
OOOHeadIndexReader was using the headIndexReader PostingsForMatchers()
and LabelValues() implementation which lead to a very subtle bug that
led to wrong query results.

headIndexReader LabelValues() implementation checks if the query
timerange overlaps with the head maxt and mint and if it doesnt it
returns an empty list of values. Since this code was also used by the
ooo head it led to wrong results that we were not able to see in tests
because our queries where always from MinInt64 to MaxInt64. This commit
also adds a new test that performs multiple time range queries to make
sure this never happens again.

### DIFF
--- a/tsdb/db_test.go
+++ b/tsdb/db_test.go
@@ -4196,7 +4196,7 @@ func TestOOOAppendAndQuery(t *testing.T) {
 	s2 := labels.FromStrings("foo", "bar2")
 
 	minutes := func(m int64) int64 { return m * time.Minute.Milliseconds() }
-	expSamples := make(map[string][]tsdbutil.Sample)
+	appendedSamples := make(map[string][]tsdbutil.Sample)
 	totalSamples := 0
 	addSample := func(lbls labels.Labels, fromMins, toMins int64, faceError bool) {
 		app := db.Appender(context.Background())
@@ -4209,7 +4209,7 @@ func TestOOOAppendAndQuery(t *testing.T) {
 				require.Error(t, err)
 			} else {
 				require.NoError(t, err)
-				expSamples[key] = append(expSamples[key], sample{t: min, v: val})
+				appendedSamples[key] = append(appendedSamples[key], sample{t: min, v: val})
 				totalSamples++
 			}
 		}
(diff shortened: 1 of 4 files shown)


tsdb/db_test.go
```diff
@@ -4290,6 +4290,106 @@ func TestOOOAppendAndQuery(t *testing.T) {
...
+	require.Equal(t, float64(2), prom_testutil.ToFloat64(db.head.metrics.chunksCreated))
+
+	// Some ooo samples.
+	addSample(s1, 250, 260, false) // 250, 251, 252, 253, ... 265
+	addSample(s2, 255, 265, false)
+	verifyOOOMinMaxTimes(250, 265)
+
+	testQuery(250, 265) // Test query only for the ouf of order data only
+	testQuery(290, 300) // Test query only for the in order data only
+	testQuery(250, 300) // Test query for both
```
colega, line 4390: Can we make this a subtest of the previous test? To avoid all the repetitions or something. Not a blocker.

tsdb/ooo_head_read.go
```diff
@@ -134,6 +134,27 @@ func (oh *OOOHeadIndexReader) series(ref storage.SeriesRef, lbls *labels.Labels,
...
+// PostingsForMatchers needs to be overridden so that the right IndexReader
+// implementation gets passed down to the PostingsForMatchers call.
+func (oh *OOOHeadIndexReader) PostingsForMatchers(concurrent bool, ms ...*labels.Matcher) (index.Postings, error) {
+	return oh.head.pfmc.PostingsForMatchers(oh, concurrent, ms...)
+}
+
+// LabelValues needs to be overridden from the headIndexReader implementation due
+// to the check that happens at the beginning where we make sure that the query
+// interval overlaps with the head minooot and maxooot.
+func (oh *OOOHeadIndexReader) LabelValues(name string, matchers ...*labels.Matcher) ([]string, error) {
```
colega, line 146: Since we know this was the issue, I think we need to have an explicit test for this:

```
func TestOOOHeadIndexReader_LabelValues() {
    // create head
    // push OOO series
    // get OOOHead's index reader without overlapping head
    // call LabelValues and see
}
```

colega (approved):
The fix LGTM, but I'd add a test for LabelValues.

tsdb/ooo_head_read_test.go
```diff
@@ -357,6 +358,52 @@ func TestOOOHeadIndexReader_Series(t *testing.T) {
...
+	}, 90, 1)
+	require.NoError(t, err)
+	_, err = app.Append(0, labels.Labels{
+		{Name: "foo", Value: fmt.Sprintf("bar2")},
+	}, 90, 2)
+	require.NoError(t, err)
+
+	require.NoError(t, app.Commit())
+
+	oh := NewOOOHeadIndexReader(head, math.MinInt64, math.MaxInt64)
```
colega, line 390: You need to create an `OOOHeadIndexReader` that doesn't overlap with the Head range in order to test the issue you're fixing, right? I think this test would actually pass without the fix.

tsdb/ooo_head_read_test.go
```diff
@@ -357,6 +358,52 @@ func TestOOOHeadIndexReader_Series(t *testing.T) {
...
+	matchers := []*labels.Matcher{labels.MustNewMatcher(labels.MatchEqual, "foo", "bar1")}
+	values, err := oh.LabelValues("foo", matchers...)
```
colega, line 392: Please test also without matchers, as that takes a different code path.

tsdb/ooo_head_read_test.go
jesusvazquez: Yes sorry it was not complete, i was planning to do different queries next

PR #315, opened 2022-08-10
Gofumpt all files to fix ci lint

jesusvazquez:
CI is failing on main because so files are not gofumpt'd.

### DIFF
--- a/discovery/xds/kuma_mads.pb.go
+++ b/discovery/xds/kuma_mads.pb.go
@@ -23,13 +23,14 @@ package xds
 
 import (
 	context "context"
+	reflect "reflect"
+	sync "sync"
+
 	v3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
 	_ "github.com/envoyproxy/protoc-gen-validate/validate"
 	_ "google.golang.org/genproto/googleapis/api/annotations"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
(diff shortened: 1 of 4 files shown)


colega:
I think we should avoid ad-hoc linter fixes here, and bring them from upstream prometheus, to have less changes later.

Is this also happening in upstream? If not, what's different here?

jesusvazquez:
@colega I think in this case both repositories have diverged a lot. Prometheus has moved to use circle ci and the workfow `test.yml` which is the one we're using here does not exist.
- mimir-prometheus workflows https://github.com/grafana/mimir-prometheus/tree/main/.github/workflows
- prometheus workflows https://github.com/prometheus/prometheus/tree/main/.github/workflows

---

The reason why CI is failing is that

In test.yml workflow we're using golangci-lint v2 https://github.com/grafana/mimir-prometheus/blob/5db6963dea6c03cfd4c7e3ef102877af69c6f77c/.github/workflows/test.yml#L13-L14


Whereas in golangci-lint.yml workflow we use v3 https://github.com/grafana/mimir-prometheus/blob/d713e39865be0bd950404d1d0bb90a60ab4804e7/.github/workflows/golangci-lint.yml#L27-L30

The last workflow is present on both repos so I think the workaround for now is to remove the redundant `Lint` job in test.yml and then maybe decide what we do with CI at @codesome 's https://github.com/grafana/mimir-prometheus/pull/310/files

jesusvazquez:
Closing in favor of https://github.com/grafana/mimir-prometheus/pull/316

PR #316, opened 2022-08-10
Remove redundant CI Lint job

jesusvazquez:
This repository has a custom `test.yml` github workflow https://github.com/grafana/mimir-prometheus/blob/628f66d19afcef69b38fb6fc7b259e60a0ad1c31/.github/workflows/test.yml#L1 that has a redundant lint stage since there is a job from upstream that already does this https://github.com/grafana/mimir-prometheus/blob/d713e39865be0bd950404d1d0bb90a60ab4804e7/.github/workflows/golangci-lint.yml#L27-L30

So this PR is getting rid of the obsolete and redudant Lint job in test.yml

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,14 +5,6 @@ on:
   pull_request:
 
 jobs:
-  lint:
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Checkout Repo
-        uses: actions/checkout@v2
-      - name: Lint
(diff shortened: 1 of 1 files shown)


colega (approved):
Redundant and outdated! It's v2, and we're on v3, probably it's just running an outdated version not compatible with go 1.18.

[I see](https://github.com/prometheus/prometheus/blob/main/.github/workflows/golangci-lint.yml) the upstream has the one that you're leaving, so LGTM

PR #353, opened 2022-11-08
Use interned strings for memSeries labelsets

colega:
Instead of using different copies of same strings again and again, this uses a map in the Head to store references to known strings.

When the head is garbage-collected, we rebuild the map of strings that are kept in it.

This is one implementation of the Option 1 from @bboreham's document [Prometheus Labels data structure](https://docs.google.com/document/d/1fXrPI2kr-jUvkF9JkkA7aPk7UQYTG9Yvk6A-fSinqEU/edit).

It is not the most optimal, but it should be a fairly important improvement. At least it's worth testing and setting a bar for further improvements.


bboreham:
Can you check the performance?
Eg on reading the WAL.

colega:
BTW, note that most slowdown is shown when there's only 50 samples per series, of course, as then the ratio of time consumed creating series vs time consumed writing samples is less favorable for us: the lock contention on the interning while creating series is more likely, thus the 30% increase. For series with 480 samples the difference is much smaller (3-5%)

PR #355, opened 2022-11-22
Sync with upstream for native histograms

zenador:
~Using CI/CD to test sync changes~


codesome:
To be noted that this should not go into Mimir until we have closed https://github.com/prometheus/prometheus/issues/11585 and https://github.com/prometheus/prometheus/issues/11602 (but we can merge it in the meantime)

krajorama (approved):
Syntax ok. I've compared the mimir additions before and after this PR and everything is accounted for.

krajorama:
> To be noted that this should not go into Mimir until we have closed [prometheus/prometheus#11585](https://github.com/prometheus/prometheus/issues/11585) and [prometheus/prometheus#11602](https://github.com/prometheus/prometheus/issues/11602) (but we can merge it in the meantime)

Not to mention that we have to test the hell out of mimir after this is vendored.

PR #369, opened 2022-11-28
tsdb: make sharding function a parameter

bboreham:
Instead of relying on `labels.Hash()`, which may change, have the caller pass in a shard function if required.
    
For most purposes `tsdb.Options.ShardFunc` is used, but the compactor may be created independently so `NewLeveledCompactorWithChunkSize` also takes a shard function parameter.
    
Regular Prometheus, which does not use block sharding, will have this parameter as nil.

### DIFF
--- a/tsdb/block.go
+++ b/tsdb/block.go
@@ -322,11 +322,11 @@ type Block struct {
 // OpenBlock opens the block in the directory. It can be passed a chunk pool, which is used
 // to instantiate chunk structs.
 func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, err error) {
-	return OpenBlockWithCache(logger, dir, pool, nil, nil)
+	return OpenBlockWithOptions(logger, dir, pool, nil, nil)
 }
 
-// OpenBlockWithCache is like OpenBlock but allows to pass a cache provider.
-func OpenBlockWithCache(logger log.Logger, dir string, pool chunkenc.Pool, cache index.ReaderCacheProvider, shardFunc func(l labels.Labels) uint64) (pb *Block, err error) {
+// OpenBlockWithOptions is like OpenBlock but allows to pass a cache provider and sharding function.
+func OpenBlockWithOptions(logger log.Logger, dir string, pool chunkenc.Pool, cache index.ReaderCacheProvider, shardFunc func(l labels.Labels) uint64) (pb *Block, err error) {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
@@ -347,7 +347,7 @@ func OpenBlockWithCache(logger log.Logger, dir string, pool chunkenc.Pool, cache
 	}
 	closers = append(closers, cr)
 
-	indexReader, err := index.NewFileReaderWithCache(filepath.Join(dir, indexFilename), cache, shardFunc)
+	indexReader, err := index.NewFileReaderWithOptions(filepath.Join(dir, indexFilename), cache, shardFunc)
 	if err != nil {
 		return nil, err
 	}
--- a/tsdb/block.go
+++ b/tsdb/block.go
@@ -87,7 +87,7 @@ type IndexReader interface {
 
 	// ShardedPostings returns a postings list filtered by the provided shardIndex
 	// out of shardCount. For a given posting, its shard MUST be computed hashing
-	// the series labels mod shardCount (eg. `labels.Hash() % shardCount == shardIndex`).
+	// the series labels mod shardCount, using a hash function which is consistent over time.
 	ShardedPostings(p index.Postings, shardIndex, shardCount uint64) index.Postings
 
 	// Series populates the given builder and chunk metas for the series identified
@@ -322,11 +322,11 @@ type Block struct {
 // OpenBlock opens the block in the directory. It can be passed a chunk pool, which is used
 // to instantiate chunk structs.
 func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, err error) {
-	return OpenBlockWithCache(logger, dir, pool, nil)
+	return OpenBlockWithCache(logger, dir, pool, nil, nil)
 }
 
 // OpenBlockWithCache is like OpenBlock but allows to pass a cache provider.
-func OpenBlockWithCache(logger log.Logger, dir string, pool chunkenc.Pool, cache index.ReaderCacheProvider) (pb *Block, err error) {
+func OpenBlockWithCache(logger log.Logger, dir string, pool chunkenc.Pool, cache index.ReaderCacheProvider, shardFunc func(l labels.Labels) uint64) (pb *Block, err error) {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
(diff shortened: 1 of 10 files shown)


tsdb/compact.go
```diff
@@ -74,11 +74,11 @@ type Compactor interface {
 	//  * No block is written.
 	//  * The source dirs are marked Deletable.
 	//  * Returns empty ulid.ULID{}.
-	Compact(dest string, dirs []string, open []*Block) (ulid.ULID, error)
+	Compact(dest string, dirs []string, open []*Block, shard func(labels.Labels) uint64) (ulid.ULID, error)
```
pstibrany, line 77: Instead of breaking existing API, shall we add hashing function as a configuration option to `LeveledCompactor`, and keep interface unchanged?

bboreham: This particular change was not actually required; I took it out.  But I left it in `CompactWithSplitting`.

I think I would ask why `shardCount` is not a configuration option by the same argument.

pstibrany: IIRC `LeveledCompactor` is typically created once and reused between compactions. Hashing function doesn't need to change between compactions, while `shardCount` is specific for the individual compaction.

If we're only changing `CompactWithSplitting`, I don't mind having it as parameter either. Passing `nil` to function from `Compact` also verifies that sharding function is not used when shardCount=1.

aknuds1 (approved):
Looks good AFAICT. Awaiting @pstibrany's opinion :)

tsdb/compact.go
```diff
@@ -1159,7 +1159,10 @@ func (c *LeveledCompactor) populateSymbols(sets []storage.ChunkSeriesSet, outBlo
 
 		s := seriesSet.At()
 
-		obIx := s.Labels().Hash() % uint64(len(outBlocks))
+		var obIx uint64
+		if len(outBlocks) > 1 {
+			obIx = shard(s.Labels()) % uint64(len(outBlocks))
```
pstibrany, line 1164: we save a call to `shard` here, making non-split (eg. head) compaction bit faster. nice find.

pstibrany (approved):
lgtm, thanks.

bboreham:
I'm trying out the alternative @pstibrany suggested, storing the shard func in LeveledCompactor, at 8db1b3f1f3287517e641e3a61fa284fa4ede8ef5

pracucci:
> I'm trying out the alternative @pstibrany suggested, storing the shard func in LeveledCompactor, at [8db1b3f](https://github.com/grafana/mimir-prometheus/commit/8db1b3f1f3287517e641e3a61fa284fa4ede8ef5)

I personally like it more.

bboreham:
I have updated with a lot more places that needed to be changed, and rebased.

`cmd/compact/main.go` is using `labels.Hash` but I don't know if that sharding functionality belongs in a command in the Prometheus repo.

bboreham:
I have cherry-picked the commit from https://github.com/prometheus/prometheus/pull/11674 to see if that cures the CI failures.

tsdb/block.go
```diff
@@ -322,11 +322,11 @@ type Block struct {
 // OpenBlock opens the block in the directory. It can be passed a chunk pool, which is used
 // to instantiate chunk structs.
 func OpenBlock(logger log.Logger, dir string, pool chunkenc.Pool) (pb *Block, err error) {
-	return OpenBlockWithCache(logger, dir, pool, nil)
+	return OpenBlockWithCache(logger, dir, pool, nil, nil)
```
pracucci, line 325: Wouldn't be better if we pass `func(l labels.Labels) uint64 { return l.Hash() }` instead of `nil` as default behaviour?

tsdb/compact.go
```diff
@@ -157,10 +158,10 @@ func newCompactorMetrics(r prometheus.Registerer) *compactorMetrics {
 
 // NewLeveledCompactor returns a LeveledCompactor.
 func NewLeveledCompactor(ctx context.Context, r prometheus.Registerer, l log.Logger, ranges []int64, pool chunkenc.Pool, mergeFunc storage.VerticalChunkSeriesMergeFunc, enableOverlappingCompaction bool) (*LeveledCompactor, error) {
-	return NewLeveledCompactorWithChunkSize(ctx, r, l, ranges, pool, chunks.DefaultChunkSegmentSize, mergeFunc, enableOverlappingCompaction)
+	return NewLeveledCompactorWithChunkSize(ctx, r, l, ranges, pool, chunks.DefaultChunkSegmentSize, mergeFunc, enableOverlappingCompaction, nil)
```
pracucci, line 161: Same here. Have you considered it?

tsdb/index/index.go
```diff
@@ -1127,11 +1129,11 @@ func NewReaderWithCache(b ByteSlice, cacheProvider ReaderCacheProvider) (*Reader
 
 // NewFileReader returns a new index reader against the given index file.
 func NewFileReader(path string) (*Reader, error) {
-	return NewFileReaderWithCache(path, nil)
+	return NewFileReaderWithCache(path, nil, nil)
```
pracucci, line 1132: _Same here_

pracucci (approved):
Overall LGTM. I have of couple of high level comments:
1. In some exported functions, we initialise the `shardFunc` to `nil`. I'm wondering if you considered using `func(l labels.Labels) uint64 { return l.Hash() }` as default (looks a better approach to me)
2. The functions `NewFileReaderWithCache()` and `NewLeveledCompactorWithChunkSize()` were named like this because they allowed to customise the cache and chunk size respectively. Now they also take the `shardFunc`. I'm wondering if we should rename them to `...WithOptions()` to be more generic.

tsdb/block.go
bboreham: I wanted `nil`, so it would crash on any code path that needs a shard function.  In other words checking if I had missed something. 
Arguably this is less appropriate for released software.

bboreham: To strengthen this, I forgot to set `ShardFunc` in one code path today, and a test immediately crashed.
If it uses `l.Hash()` when the underlying Labels implementation changes, series will silently move to different shards and we have to hope that tests pick that up.

bboreham:
> I'm wondering if we should rename them to ...WithOptions() to be more generic.

I did this.

pstibrany:
It may be too late for my comment, but: after some more thoughts I think we should go with function parameter to `CompactWithSplitting` method.

My initial suggestion to use configuration option was a reaction to seeing `Compact` function signature changed. However if we only change `CompactWithSplitting`, this is not a problem.

Why is changing signature better (in my opinion)?
1. it is a signal to the reader that label hash function is only required when splitting, but not normal compaction.
2. changing signature and breaking `CompactWithSplitting` clients is actually a good thing here. If we use config option, we need to rely on running the code to find unset label hashing function (eg. unit tests in Mimir). By changing signature instead we get compilation errors, ie. much faster detection of places where the function is missing.

In retrospect, I am sorry for my earlier suggestion to use config option. I think it made this PR much bigger than it needs to be, changing all kinds of "support" code unnecessarily. If it's too late to change the direction of the PR (or you disagree), let's keep it as-is now.

pstibrany:
One more reason to keep this PR small: I'd like to try to upstream splitting compactor to Prometheus, as there seem to be some interest from Thanos and Cortex to use it too. By having the diff as small as possible, I think our chances would increase.

bboreham:
Changing from a member of `LeveledCompactor` to a function parameter is a bigger change - [see](https://github.com/grafana/mimir-prometheus/compare/shard-func-2).
Many more methods need to pass it through - `populateBlock`, `populateSymbols`, `compactOOO`, `Write`, `write`,  ...
(I haven't tried that change with Mimir yet - it might need to be even bigger.)

Maybe there should be a broader redesign so that not so many methods are touched by sharding

PR #374, opened 2022-12-02
Hackathon: raw_increase

colega:
Introduces a new function `raw_increase(matrix, scalar millis)` that calculates the raw incrase ("raw" as opposed to the extrapolated increase that `increase()` and `rate()` calculate) in the window provided by the scalar millis param. It can make use of a bigger matrix range if there are not enough values in the window, to find the previous
value.

It accounts for the counter resets.

If there are at least 2 samples in the window, it doesn't look back past the window.

### DIFF
--- a/promql/functions.go
+++ b/promql/functions.go
@@ -211,6 +211,50 @@ func histogramRate(points []Point, isCounter bool) *histogram.FloatHistogram {
 	return h.Compact(0)
 }
 
+func funcRawIncrease(vals []parser.Value, args parser.Expressions, enh *EvalNodeHelper) Vector {
+	var (
+		samples     = vals[0].(Matrix)[0]
+		window      = int64(vals[1].(Vector)[0].V)
+		resultValue float64
+	)
+
+	// Make sure that the requested
+	if ms := args[0].(*parser.MatrixSelector); window > ms.Range.Milliseconds() {
+		panic(fmt.Sprintf("invalid window for raw_increase, can't be bigger than matrix selector range: window=%dms > range=%dms", window, ms.Range.Milliseconds()))
+	}
+
+	// No sense in trying to compute a rate without at least two points. Drop this Vector element.
+	if len(samples.Points) < 2 {
+		return enh.Out
+	}
+
+	if samples.Points[0].H != nil {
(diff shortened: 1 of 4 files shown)


promql/functions.go
```diff
@@ -211,6 +211,50 @@ func histogramRate(points []Point, isCounter bool) *histogram.FloatHistogram {
...
 }
 
+func funcRawIncrease(vals []parser.Value, args parser.Expressions, enh *EvalNodeHelper) Vector {
+	var (
+		samples     = vals[0].(Matrix)[0]
+		window      = int64(vals[1].(Vector)[0].V)
+		resultValue float64
+	)
+
+	// Make sure that the requested
```
replay, line 221: that's incomplete

colega: Fixed.

promql/functions.go
```diff
@@ -211,6 +211,50 @@ func histogramRate(points []Point, isCounter bool) *histogram.FloatHistogram {
...
+		panic(fmt.Sprintf("invalid window for raw_increase, can't be bigger than matrix selector range: window=%dms > range=%dms", window, ms.Range.Milliseconds()))
+	}
+
+	// No sense in trying to compute a rate without at least two points. Drop this Vector element.
+	if len(samples.Points) < 2 {
+		return enh.Out
+	}
+
+	if samples.Points[0].H != nil {
+		// TODO (oleg): implement raw increase for histograms.
```
replay, line 232: is that for native histograms?

promql/testdata/functions.test
```diff
@@ -1037,3 +1037,31 @@ eval instant at 5m log10(exp_root_log - 20)
 	{l="y"} -Inf
 
 clear
+
+# Testing raw_increase().
+load 30s
+	http_requests{path="/foo"}	0+10x6 10 20 30 40
+	http_requests{path="/bar"}	0+10x6 10 20 30 10
+	http_requests{path="/baz"}	0+10x6 10 20  _ 10
+	http_requests{path="/boo"}	0+10x6  _  _  _ 10
```
replay, line 1046: would it be worth it to add a test with multiple resets?

PR #378, opened 2022-12-28
Allow forcing usage of PostingsForMatchersCache

colega:
When out-of-order is enabled, queries go through both Head and OOOHead, and they both execute the same PostingsForMatchers call, as memSeries are shared for both.

In some cases these calls can be heavy, and also frequent. We can deduplicate those calls by using the PostingsForMatchers cache that we already use for query sharding.

The usage of this cache can skip a newly appended series in the results for the duration of the ttl.


codesome (approved):

pracucci:
@colega Can you add a CHANGELOG entry, please?

PR #395, opened 2023-01-16
Merge Jan 16 from Upstream

fayzal-g:
Merging up to upstream https://github.com/prometheus/prometheus/commit/cb2be6e62ff80ba84b639e4251044214e185886f


bboreham (approved):
LGTM.  

I believe this contains a breaking change to the WAL disk format for sparse histograms, so we might get some disruption when rolling out in any environment receiving those.

PR #400, opened 2023-01-19
Option to align rule group's evaluation time to interval

pstibrany:
This PR adds option to rule groups to align their evaluation time on evaluation interval.

This can be useful for example redundancy when two rulers are evaluating same rules -- if they both generate samples with same timestamp, only one sample will be written. But if one of rulers fails, the other one can keep on writing data.

(Our use case is different)

### DIFF
--- a/rules/fixtures/rules_with_alignment.yaml
+++ b/rules/fixtures/rules_with_alignment.yaml
@@ -1,0 +1,20 @@
+groups:
+  - name: aligned
+    align_execution_time_on_interval: true
+    interval: 5m
+    rules:
+      - record: job:http_requests:rate5m
+        expr: sum by (job)(rate(http_requests_total[5m]))
+
+  - name: unaligned_default
+    interval: 5m
+    rules:
+      - record: job:http_requests:rate5m
+        expr: sum by (job)(rate(http_requests_total[5m]))
+
+  - name: unaligned_explicit
+    interval: 5m
+    align_execution_time_on_interval: false
+    rules:
+      - record: job:http_requests:rate5m
+        expr: sum by (job)(rate(http_requests_total[5m]))
--- a/rules/fixtures/rules_with_alignment.yaml
+++ b/rules/fixtures/rules_with_alignment.yaml
@@ -6,6 +6,13 @@ groups:
       - record: job:http_requests:rate5m
         expr: sum by (job)(rate(http_requests_total[5m]))
 
+  - name: aligned_with_crazy_interval
+    align_execution_time_on_interval: true
+    interval: 1m27s
+    rules:
+      - record: job:http_requests:rate5m
+        expr: sum by (job)(rate(http_requests_total[5m]))
+
   - name: unaligned_default
     interval: 5m
     rules:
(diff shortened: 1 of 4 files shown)


rules/manager_test.go
```diff
@@ -919,6 +919,67 @@ func TestUpdateSetsSourceTenants(t *testing.T) {
...
+	ruleManager.start()
+	defer ruleManager.Stop()
+
+	rgs, errs := rulefmt.ParseFile("fixtures/rules_with_alignment.yaml")
+	require.Empty(t, errs, "file parsing failures")
+
+	tmpFile, err := os.CreateTemp("", "rules.test.*.yaml")
+	require.NoError(t, err)
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
```
replay, line 949:
```suggestion
	t.Cleanup(func() {
		os.Remove(tmpFile.Name())
		tmpFile.Close()
	})
```

rules/fixtures/rules_with_alignment.yaml
```diff
@@ -0,0 +1,20 @@
+groups:
+  - name: aligned
+    align_execution_time_on_interval: true
+    interval: 5m
```
replay, line 4: would it make sense to add a test case with a weird `interval`, like for example `1m27s`?

replay:
Doesn't `Group.CopyState()` need to be updated to copy the new `alignExecutionTimeOnInterval ` property?

replay (approved):
Looks great, thanks!

rules/manager_test.go
```diff
@@ -919,6 +919,67 @@ func TestUpdateSetsSourceTenants(t *testing.T) {
...
+			if lg.name == groupName {
+				g = lg
+				break
+			}
+		}
+		require.NotNil(t, g, "group not found: %s", groupName)
+
+		// In very rare cases (when g.hash() % g.interval == 0) alignment cannot be checked, because aligned and unaligned eval timestamps
+		// would be the same. Chance of this happening is tiny.
+		require.NotZero(t, g.hash()%uint64(g.interval), "alignment cannot be checked")
```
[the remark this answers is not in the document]
pstibrany: Yes it is. Perhaps we can call `t.Skip()` in that case.

pstibrany:
> Doesn't `Group.CopyState()` need to be updated to copy the new `alignExecutionTimeOnInterval ` property?

I don't think that's right. `CopyState` is used when rule group was updated. In that case old version of the group is stopped, and before new group is started, state from old group is copied over to new group. New group already has the correct aligned option set (it could even be different from old group!).

rules/fixtures/rules_with_alignment.yaml
pstibrany: Added in https://github.com/grafana/mimir-prometheus/pull/400/commits/039b2a08a49f4c38b3f24b30be2215ee168563d8.

rules/manager_test.go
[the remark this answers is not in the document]
pstibrany: Improved comment and added skipping in https://github.com/grafana/mimir-prometheus/pull/400/commits/039b2a08a49f4c38b3f24b30be2215ee168563d8.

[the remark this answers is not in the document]
replay: Great thanks, using `t.Skip()` sounds like the right choice (I didn't know about this before).

rules/manager_test.go
pstibrany: I will keep this as-is, as other tests use the same pattern.

PR #402, opened 2023-01-23
Expose `Matcher.Prefix()`

colega:
Sometimes label matchers know that they match values with a specific prefix. This information can be valuable in some downstream storage implementations.


colega:
I was wondering if we should expose a non-empty prefix value on the equality matchers, as that's a specific case of prefix matching (entire value is a prefix), but I decided to not to, just like `SetMatches()` doesn't return a single-value slice when it's called on the equality matchers.

codesome (approved):

PR #404, opened 2023-01-26
Fixup labels usage

bboreham:
Code should not be assuming `labels.Labels` is a slice, because at some point that could change.

### DIFF
--- a/promql/engine_test.go
+++ b/promql/engine_test.go
@@ -3114,17 +3114,11 @@ func TestRangeQuery(t *testing.T) {
 			Result: Matrix{
 				Series{
 					Points: []Point{{V: 1, T: 0}, {V: 3, T: 60000}, {V: 5, T: 120000}},
-					Metric: labels.Labels{
-						labels.Label{Name: "__name__", Value: "bar"},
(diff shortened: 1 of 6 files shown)


pstibrany (approved):

pracucci:
Shouldn't these changes be done in the upstream Prometheus and then get vendored here? 🤔

bboreham:
These are changes to code that doesn't exist upstream, e.g. to do with sharding.

pracucci:
Aha! I didn't realize from a quick look. LGTM :)

bboreham:
CI is timing out queries again - #407 should fix.

bboreham:
Rebased and added a few more fixes for code added since January.

pstibrany (approved):
Changes look fine. I was hoping that `go test -tags stringlabels ./...` would work, but there were compilation problems:
```
# github.com/prometheus/prometheus/model/labels
model/labels/sharding.go:12:20: cannot range over ls (variable of type Labels)
model/labels/sharding.go:17:22: invalid operation: cannot slice ls (variable of type Labels)
```

colega (approved):

PR #411, opened 2023-01-30
Hardcode the labels stable hash function instead of taking it as an option

pracucci:
This PR is a follow up of https://github.com/grafana/mimir-prometheus/pull/369. Instead of taking the sharding function as an option, we always use a "stable" hashing function which is guaranteed to not change over time.

### DIFF
--- a/model/labels/sharding.go
+++ b/model/labels/sharding.go
@@ -1,0 +1,32 @@
+package labels
+
+import (
+	"github.com/cespare/xxhash/v2"
+)
+
+// StableHash is a labels hashing implementation which is guaranteed to not change over time.
+// This function should be used whenever labels hashing backward compatibility must be guaranteed.
+func StableHash(ls Labels) uint64 {
+	// Use xxhash.Sum64(b) for fast path as it's faster.
+	b := make([]byte, 0, 1024)
+	for i, v := range ls {
+		if len(b)+len(v.Name)+len(v.Value)+2 >= cap(b) {
+			// If labels entry is 1KB+ do not allocate whole entry.
+			h := xxhash.New()
+			_, _ = h.Write(b)
+			for _, v := range ls[i:] {
+				_, _ = h.WriteString(v.Name)
+				_, _ = h.Write(seps)
+				_, _ = h.WriteString(v.Value)
+				_, _ = h.Write(seps)
+			}
+			return h.Sum64()
+		}
+
+		b = append(b, v.Name...)
+		b = append(b, seps[0])
+		b = append(b, v.Value...)
+		b = append(b, seps[0])
+	}
+	return xxhash.Sum64(b)
+}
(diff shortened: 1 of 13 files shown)


tsdb/head.go
```diff
@@ -1568,12 +1566,8 @@ func (h *Head) getOrCreate(hash uint64, lset labels.Labels) (*memSeries, bool, e
 }
 
 func (h *Head) getOrCreateWithID(id chunks.HeadSeriesRef, hash uint64, lset labels.Labels) (*memSeries, bool, error) {
-	shardHash := hash
-	if h.opts.ShardFunc != nil {
-		shardHash = h.opts.ShardFunc(lset)
-	}
 	s, created, err := h.series.getOrSet(hash, lset, func() *memSeries {
-		return newMemSeries(lset, id, shardHash, h.opts.ChunkEndTimeVariance, h.opts.IsolationDisabled)
+		return newMemSeries(lset, id, labels.StableHash(lset), h.opts.ChunkEndTimeVariance, h.opts.IsolationDisabled)
```
pracucci, line 1570: Note to reviewers: this callback is not called within a lock, so should be fine moving the hashing here.

tsdb/head.go
```diff
@@ -1792,7 +1786,7 @@ func (s *stripeSeries) getOrSet(hash uint64, lset labels.Labels, createSeries fu
 	// rather it gives a 'hint' whether to create a series or not.
 	preCreationErr := s.seriesLifecycleCallback.PreCreation(lset)
 
-	// Create the series, unless the PreCreation() callback as failed.
+	// Create the series, unless the PreCreation() callback has failed.
```
bboreham, line 1789: Best to do this in a separate PR upstream then merge to here.

pracucci:
> As mentioned in Mimir's PR, we'd need to make sure that there's a test ensuring that we're using this specific hashing algorightm.

Yes, I'm wondering on it, but it's a set of tests we'll have in Mimir.

model/labels/sharding.go
```diff
@@ -0,0 +1,39 @@
...
+// StableHash is a labels hashing implementation which is guaranteed to not change over time.
+// This function should be used whenever labels hashing backward compatibility must be guaranteed.
+func StableHash(l Labels) uint64 {
+	hash, _ := labelsXXHash(make([]byte, 0, 1024), l)
+	return hash
+}
+
+// labelsXXHash replicates historical hash computation from Prometheus circa 2022.
+// The buffer b can be recycled to avoid allocations.
+func labelsXXHash(b []byte, ls Labels) (uint64, []byte) {
```
bboreham, line 16: I think having a separate function was only useful if it could be called with a reused buffer, and this ability is not used in this PR.
Maybe profiling will tell us we need that, but otherwise you can elide the two functions.

bboreham (commented):
lgtm; couple of thoughts below

tsdb/head.go
pracucci: Yep, done.

model/labels/sharding.go
pracucci: Yep, done.

krajorama:
late to the party, but +1

PR #412, opened 2023-01-31
Add extension point for returning different content types from API endpoints

charleskorn:
Duplicate of https://github.com/prometheus/prometheus/pull/11905 to unblock work on new query result payload format.


charleskorn:
CodeQL check is failing on unrelated changes

pracucci:
Reviewed the Prometheus PR and LGTM. Going to merge this to unblock Charles, with the hope Prometheus will accept the change so we'll not have to maintain the diff here.

PR #427, opened 2023-02-02
Extension point for returning different content types from API endpoints: expose QueryData so that implementations of Codec can use it.

charleskorn:
Continuation of #412 - ran into this issue while [implementing the new protobuf codec](https://github.com/grafana/mimir/commit/110a8b1617b7a3df3eff02e4ab456ca2bcba2645) in the querier.

### DIFF
--- a/web/api/v1/api.go
+++ b/web/api/v1/api.go
@@ -388,7 +388,7 @@ func (api *API) Register(r *route.Router) {
 	r.Put("/admin/tsdb/snapshot", wrapAgent(api.snapshot))
 }
 
-type queryData struct {
+type QueryData struct {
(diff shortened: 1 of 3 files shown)


pracucci:
Would be great to have these changes proposed in upstream Prometheus too. For example, exposing a struct like you did here should be non controversial.

charleskorn:
> Would be great to have these changes proposed in upstream Prometheus too. For example, exposing a struct like you did here should be non controversial.

Yep, I've added it to https://github.com/prometheus/prometheus/pull/11905.

charleskorn:
This is now ready for review.

pracucci (approved):

PR #428, opened 2023-02-24
Implement fully-featured content negotiation for API requests, and allow overriding the default API codec.

charleskorn:
This resolves a concern with merging content negotiation to upstream (see https://github.com/prometheus/prometheus/pull/11905). 

It also allows us to override the default codec, which will allow us to use a custom codec that forbids encoding native histograms in the default JSON format between queriers and query frontends.


krajorama (approved):
LGTM

pracucci (approved):
Good job!

PR #430, opened 2023-02-28
Optimized FastRegexMatcher when the regex contains a case insensitive alternation made with concats too

pracucci:
We have a customer running queries with case insensitive regular expression matchers similar to the following one:
```
label=~"(?i:(foo1|foo2|bar)"
```

When we receive such queries, we see almost 40% of ingesters CPU spent running the regex:

Screenshot 2023-02-28 at 17 52 08

These types of regex are expected to be accelerated by `FastRegexMatcher`, in particular by `stringMatcherFromRegexp()` so why isn't the fast path triggered?

The reason is that a regex like `(?i:(foo1|foo2|bar)` is unrolled by the engine as an alternate **and** a concat because of the common prefix between `foo1` and `foo2`. `stringMatcherFromRegexp()` can handle the scenario the regex is case insensitive and has an alternate, but cannot handle the scenario it's case insensitive and has a concat too. In this PR I've done some changes to support the case insensitive concat too.

To support it, I'm returning whether the matchers are case sensitive or not from `findSetMatches()`. `findSetMatches()` returns nil matches if the regex has a mix of case sensitivity.

### Benchmark

I've introduced a couple of new benchmark cases, to show the different before/after:

```
name                                                                                                                                  old time/op    new time/op    delta
FastRegexMatcher/foo-12                                                                                                                 7.42ns ± 2%    6.84ns ± 1%   -7.83%  (p=0.023 n=3+3)
FastRegexMatcher/^foo-12                                                                                                                8.87ns ±28%    6.90ns ± 2%     ~     (p=0.254 n=3+3)
FastRegexMatcher/(foo|bar)-12                                                                                                           8.93ns ± 3%    8.54ns ± 3%     ~     (p=0.132 n=3+3)
FastRegexMatcher/foo.*-12                                                                                                               29.0ns ± 2%    29.1ns ± 3%     ~     (p=0.950 n=3+3)
FastRegexMatcher/.*foo-12                                                                                                               32.0ns ± 3%    31.8ns ± 1%     ~     (p=0.659 n=3+3)
FastRegexMatcher/^.*foo$-12                                                                                                             31.5ns ± 2%    32.2ns ± 1%     ~     (p=0.151 n=3+3)
FastRegexMatcher/^.+foo$-12                                                                                                             31.6ns ± 2%    32.1ns ± 2%     ~     (p=0.321 n=3+3)
FastRegexMatcher/.*-12                                                                                                                  26.3ns ± 0%    26.2ns ± 1%     ~     (p=0.585 n=3+3)
FastRegexMatcher/.+-12                                                                                                                  27.2ns ± 1%    26.1ns ± 3%     ~     (p=0.087 n=3+3)
FastRegexMatcher/foo.+-12                                                                                                               30.3ns ± 4%    28.1ns ± 2%   -7.43%  (p=0.049 n=3+3)
FastRegexMatcher/.+foo-12                                                                                                               31.7ns ± 1%    32.7ns ± 1%   +3.19%  (p=0.014 n=3+3)
FastRegexMatcher/.*foo.*-12                                                                                                             79.4ns ± 4%    79.8ns ± 1%     ~     (p=0.830 n=3+3)
FastRegexMatcher/(?i:foo)-12                                                                                                            21.4ns ± 1%    21.0ns ± 1%     ~     (p=0.187 n=3+3)
FastRegexMatcher/(?i:(foo|bar))-12                                                                                                      42.1ns ± 2%    41.7ns ± 3%     ~     (p=0.684 n=3+3)
FastRegexMatcher/(?i:(foo1|foo2|bar))-12                                                                                                 168ns ± 2%      66ns ± 1%  -60.73%  (p=0.000 n=3+3)
FastRegexMatcher/(?i:(foo1|foo2|aaa|bbb|ccc|ddd|eee|fff|ggg|hhh|iii|lll|mmm|nnn|ooo|ppp|qqq|rrr|sss|ttt|uuu|vvv|zzz|xxx|yyy|www))-12    1.57µs ± 3%    0.39µs ± 2%  -74.98%  (p=0.000 n=3+3)
FastRegexMatcher/(prometheus|api_prom)_api_v1_.+-12                                                                                     45.8ns ± 1%    45.0ns ± 1%     ~     (p=0.086 n=3+3)
FastRegexMatcher/((fo(bar))|.+foo)-12                                                                                                   58.4ns ± 1%    60.0ns ± 1%   +2.67%  (p=0.019 n=3+3)

name                                                                                                                                  old alloc/op   new alloc/op   delta
FastRegexMatcher/foo-12                                                                                                                  0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/^foo-12                                                                                                                 0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/(foo|bar)-12                                                                                                            0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/foo.*-12                                                                                                                0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/.*foo-12                                                                                                                0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/^.*foo$-12                                                                                                              0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/^.+foo$-12                                                                                                              0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/.*-12                                                                                                                   0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/.+-12                                                                                                                   0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/foo.+-12                                                                                                                0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/.+foo-12                                                                                                                0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/.*foo.*-12                                                                                                              0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/(?i:foo)-12                                                                                                             0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/(?i:(foo|bar))-12                                                                                                       0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/(?i:(foo1|foo2|bar))-12                                                                                                 0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/(?i:(foo1|foo2|aaa|bbb|ccc|ddd|eee|fff|ggg|hhh|iii|lll|mmm|nnn|ooo|ppp|qqq|rrr|sss|ttt|uuu|vvv|zzz|xxx|yyy|www))-12     0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/(prometheus|api_prom)_api_v1_.+-12                                                                                      0.00B          0.00B          ~     (zero variance)
FastRegexMatcher/((fo(bar))|.+foo)-12                                                                                                    0.00B          0.00B          ~     (zero variance)

name                                                                                                                                  old allocs/op  new allocs/op  delta
FastRegexMatcher/foo-12                                                                                                                   0.00           0.00          ~     (zero variance)
FastRegexMatcher/^foo-12                                                                                                                  0.00           0.00          ~     (zero variance)
FastRegexMatcher/(foo|bar)-12                                                                                                             0.00           0.00          ~     (zero variance)
FastRegexMatcher/foo.*-12                                                                                                                 0.00           0.00          ~     (zero variance)
FastRegexMatcher/.*foo-12                                                                                                                 0.00           0.00          ~     (zero variance)
FastRegexMatcher/^.*foo$-12                                                                                                               0.00           0.00          ~     (zero variance)
FastRegexMatcher/^.+foo$-12                                                                                                               0.00           0.00          ~     (zero variance)
FastRegexMatcher/.*-12                                                                                                                    0.00           0.00          ~     (zero variance)
FastRegexMatcher/.+-12                                                                                                                    0.00           0.00          ~     (zero variance)
FastRegexMatcher/foo.+-12                                                                                                                 0.00           0.00          ~     (zero variance)
FastRegexMatcher/.+foo-12                                                                                                                 0.00           0.00          ~     (zero variance)
FastRegexMatcher/.*foo.*-12                                                                                                               0.00           0.00          ~     (zero variance)
FastRegexMatcher/(?i:foo)-12                                                                                                              0.00           0.00          ~     (zero variance)
FastRegexMatcher/(?i:(foo|bar))-12                                                                                                        0.00           0.00          ~     (zero variance)
FastRegexMatcher/(?i:(foo1|foo2|bar))-12                                                                                                  0.00           0.00          ~     (zero variance)
FastRegexMatcher/(?i:(foo1|foo2|aaa|bbb|ccc|ddd|eee|fff|ggg|hhh|iii|lll|mmm|nnn|ooo|ppp|qqq|rrr|sss|ttt|uuu|vvv|zzz|xxx|yyy|www))-12      0.00           0.00          ~     (zero variance)
FastRegexMatcher/(prometheus|api_prom)_api_v1_.+-12                                                                                       0.00           0.00          ~     (zero variance)
FastRegexMatcher/((fo(bar))|.+foo)-12                                                                                                     0.00           0.00          ~     (zero variance)
```

### DIFF
--- a/model/labels/regexp.go
+++ b/model/labels/regexp.go
@@ -49,27 +49,26 @@ func NewFastRegexMatcher(v string) (*FastRegexMatcher, error) {
 	if parsed.Op == syntax.OpConcat {
 		m.prefix, m.suffix, m.contains = optimizeConcatRegex(parsed)
 	}
-	m.setMatches = findSetMatches(parsed, "")
+	if matches, caseSensitive := findSetMatches(parsed, ""); caseSensitive {
+		m.setMatches = matches
+	}
 	m.stringMatcher = stringMatcherFromRegexp(parsed)
 
 	return m, nil
 }
 
 // findSetMatches extract equality matches from a regexp.
-// Returns nil if we can't replace the regexp by only equality matchers.
-func findSetMatches(re *syntax.Regexp, base string) []string {
-	// Matches are case sensitive, if we find a case insensitive regexp.
-	// We have to abort.
-	if isCaseInsensitive(re) {
-		return nil
-	}
+// Returns nil if we can't replace the regexp by only equality matchers or the regexp contains
+// a mix of case sensitive and case insensitive matchers.
+func findSetMatches(re *syntax.Regexp, base string) (matches []string, caseSensitive bool) {
 	clearBeginEndText(re)
+
 	switch re.Op {
 	case syntax.OpLiteral:
-		return []string{base + string(re.Rune)}
+		return []string{base + string(re.Rune)}, isCaseSensitive(re)
 	case syntax.OpEmptyMatch:
 		if base != "" {
-			return []string{base}
+			return []string{base}, isCaseSensitive(re)
 		}
 	case syntax.OpAlternate:
 		return findSetMatchesFromAlternate(re, base)
@@ -80,7 +79,7 @@ func findSetMatches(re *syntax.Regexp, base string) []string {
 		return findSetMatchesFromConcat(re, base)
 	case syntax.OpCharClass:
 		if len(re.Rune)%2 != 0 {
-			return nil
+			return nil, false
 		}
 		var matches []string
 		var totalSet int
@@ -91,60 +90,91 @@ func findSetMatches(re *syntax.Regexp, base string) []string {
 		// In some case like negation [^0-9] a lot of possibilities exists and that
 		// can create thousands of possible matches at which points we're better off using regexp.
 		if totalSet > maxSetMatches {
-			return nil
+			return nil, false
 		}
 		for i := 0; i+1 < len(re.Rune); i = i + 2 {
 			lo, hi := re.Rune[i], re.Rune[i+1]
 			for c := lo; c <= hi; c++ {
 				matches = append(matches, base+string(c))
 			}
-
 		}
-		return matches
+		return matches, isCaseSensitive(re)
 	default:
-		return nil
+		return nil, false
 	}
-	return nil
+	return nil, false
 }
 
-func findSetMatchesFromConcat(re *syntax.Regexp, base string) []string {
+func findSetMatchesFromConcat(re *syntax.Regexp, base string) (matches []string, caseSensitive bool) {
 	if len(re.Sub) == 0 {
-		return nil
+		return nil, false
 	}
 	clearCapture(re.Sub...)
-	matches := []string{base}
+
+	matches = []string{base}
+	var matchesCaseSensitive *bool
 
 	for i := 0; i < len(re.Sub); i++ {
 		var newMatches []string
 		for _, b := range matches {
-			m := findSetMatches(re.Sub[i], b)
+			m, caseSensitive := findSetMatches(re.Sub[i], b)
 			if m == nil {
-				return nil
+				return nil, false
 			}
 			if tooManyMatches(newMatches, m...) {
-				return nil
+				return nil, false
 			}
+
+			// All matches must have the same case sensitivity. If it's the first set of matches
+			// returned, we store its sensitivity as the expected case, and then we'll check all
+			// other ones.
+			if matchesCaseSensitive == nil {
+				matchesCaseSensitive = &caseSensitive
+			}
+			if *matchesCaseSensitive != caseSensitive {
+				return nil, false
+			}
+
 			newMatches = append(newMatches, m...)
 		}
 		matches = newMatches
 	}
 
-	return matches
+	// matchesCaseInsensitive is guaranteed to be never nil.
+	return matches, *matchesCaseSensitive
 }
 
-func findSetMatchesFromAlternate(re *syntax.Regexp, base string) []string {
-	var setMatches []string
-	for _, sub := range re.Sub {
-		found := findSetMatches(sub, base)
-		if found == nil {
-			return nil
-		}
-		if tooManyMatches(setMatches, found...) {
-			return nil
-		}
-		setMatches = append(setMatches, found...)
+func findSetMatchesFromAlternate(re *syntax.Regexp, base string) (matches []string, caseSensitive bool) {
+	if len(re.Sub) == 0 {
+		return nil, false
 	}
-	return setMatches
+
+	var matchesCaseSensitive *bool
+
+	for _, sub := range re.Sub {
+		found, caseSensitive := findSetMatches(sub, base)
+		if found == nil {
+			return nil, false
+		}
+		if tooManyMatches(matches, found...) {
+			return nil, false
+		}
+
+		// All matches must have the same case sensitivity. If it's the first set of matches
+		// returned, we store its sensitivity as the expected case, and then we'll check all
+		// other ones.
+		if matchesCaseSensitive == nil {
+			matchesCaseSensitive = &caseSensitive
+		}
+		if *matchesCaseSensitive != caseSensitive {
+			return nil, false
+		}
+
+		matches = append(matches, found...)
+	}
+
+	// matchesCaseInsensitive is guaranteed to be never nil.
+	return matches, *matchesCaseSensitive
 }
 
 // clearCapture removes capture operation as they are not used for matching.
@@ -182,6 +212,12 @@ func clearBeginEndText(re *syntax.Regexp) {
 // The flag should be check at each level of the syntax tree.
 func isCaseInsensitive(reg *syntax.Regexp) bool {
 	return (reg.Flags & syntax.FoldCase) != 0
+}
+
+// isCaseSensitive tells if a regexp is case sensitive.
+// The flag should be check at each level of the syntax tree.
+func isCaseSensitive(reg *syntax.Regexp) bool {
+	return !isCaseInsensitive(reg)
 }
 
 // tooManyMatches guards against creating too many set matches
@@ -273,6 +309,7 @@ type StringMatcher interface {
 func stringMatcherFromRegexp(re *syntax.Regexp) StringMatcher {
 	clearCapture(re)
 	clearBeginEndText(re)
+
 	switch re.Op {
 	case syntax.OpPlus, syntax.OpStar:
 		if re.Sub[0].Op != syntax.OpAnyChar && re.Sub[0].Op != syntax.OpAnyCharNotNL {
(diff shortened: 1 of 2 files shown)


model/labels/regexp.go
```diff
@@ -91,60 +90,91 @@ func findSetMatches(re *syntax.Regexp, base string) []string {
...
+
+			// All matches must have the same case sensitivity. If it's the first set of matches
+			// returned, we store its sensitivity as the expected case, and then we'll check all
+			// other ones.
+			if matchesCaseSensitive == nil {
+				matchesCaseSensitive = &caseSensitive
 			}
+			if *matchesCaseSensitive != caseSensitive {
+				return nil, false
+			}
```
colega, line 136: Very opinonated, but IMO just checking indexes `j > 0` (with j being the index of `b`) reads easier than pointer setting:

```suggestion
			if j > 0 && matchesCaseSensitive != caseSensitive {
				return nil, false
			}
			matchesCaseSensitive = caseSensitive
```

Feel free to ignore this.
_(Same applied to the `findSetMatchesFromAlternate` below)_

model/labels/regexp.go
```diff
@@ -91,60 +90,91 @@ func findSetMatches(re *syntax.Regexp, base string) []string {
...
+	// matchesCaseInsensitive is guaranteed to be never nil.
+	return matches, *matchesCaseSensitive
 }
 
-func findSetMatchesFromAlternate(re *syntax.Regexp, base string) []string {
-	var setMatches []string
+func findSetMatchesFromAlternate(re *syntax.Regexp, base string) (matches []string, caseSensitive bool) {
+	if len(re.Sub) == 0 {
+		return nil, false
+	}
```
colega, line 150: If you accept my suggestion above, you can remove this condition (because zero-iteration loop will return `nil, false` too)

model/labels/regexp.go
```diff
@@ -324,22 +361,28 @@ func stringMatcherFromRegexp(re *syntax.Regexp) StringMatcher {
...
 			}
 			return orStringMatcher(or)
 		}
-		// others we found literals in the middle.
-		if len(matches) > 0 {
+
+		// We found literals in the middle. We can triggered the fast path only if
+		// the matches are case sensitive because containsStringMatcher doesn't
+		// support case insensitive.
+		if matchesCaseSensitive {
```
colega, line 385: > _because containsStringMatcher doesn't support case insensitive._

I feel this is straightforward to implement, we could actually make an insensitive decorator that just converts the matches to lowercase and passes on the Matches() call, are we going to to this?

model/labels/regexp_test.go
```diff
@@ -225,6 +227,9 @@ func BenchmarkFastRegexMatcher(b *testing.B) {
 		".+foo",
 		".*foo.*",
 		"(?i:foo)",
+		"(?i:(foo|bar))",
+		"(?i:(foo1|foo2|bar))",
+		"(?i:(foo1|foo2|aaa|bbb|ccc|ddd|eee|fff|ggg|hhh|iii|lll|mmm|nnn|ooo|ppp|qqq|rrr|sss|ttt|uuu|vvv|zzz|xxx|yyy|www))",
```
colega, line 232:
```suggestion
		"(?i:(foo1|foo2|aaa|bbb|ccc|ddd|eee|fff|ggg|hhh|iii|lll|mmm|nnn|ooo|ppp|qqq|rrr|sss|ttt|uuu|vvv|www|xxx|yyy|zzz))",
```

:trollface:

model/labels/regexp_test.go
```diff
@@ -147,60 +147,62 @@ func TestOptimizeConcatRegex(t *testing.T) {
...
+		{"fo.o|bar?|^baz", nil, false},
 		// case sensitive wrapper.
-		{"(?i)foo", nil},
+		{"(?i)foo", []string{"FOO"}, false},
 		// case sensitive wrapper on alternate.
-		{"(?i)foo|bar|baz", nil},
-		// case sensitive wrapper on concat.
-		{"(api|rpc)_(v1|prom)_((?i)push|query)", nil},
+		{"(?i)foo|bar|baz", []string{"FOO", "BAR", "BAZ", "BAr", "BAz"}, false},
+		// mixed case sensitive.
```
colega, line 191: Nit:
```suggestion
		// mixed case sensitivity.
```

model/labels/regexp_test.go
```diff
@@ -147,60 +147,62 @@ func TestOptimizeConcatRegex(t *testing.T) {
...
+		{"(api|rpc)_(v1|prom)_((?i)push|query)", nil, false},
```
colega, line 192: I'd like to suggest some more basic tests and also tests for unsetting the case insensitivity.

```suggestion
		{"(api|rpc)_(v1|prom)_((?i)push|query)", nil, false},
		// mixed case sensitivity concatenation only without capture group.
		{"api_v1_(?i)push", nil, false},
		// mixed case sensitivity alternation only without capture group.
		{"api|(?i)rpc)", nil, false},
		// case sensitive after unsetting insensitivity.
		{"rpc|(?i)(?-i)api", []string{"rpc", "api"}, true},
		// case sensitive after unsetting insensitivity in all alternation options.
		{"(?i)((?-i)api|(?-i)rpc)", []string{"api", "rpc"}, true},
		// mixed case sensitivity after unsetting insensitivity.
		{"(?i)rpc|(?-i)api", nil, false},
```

colega (approved):
LGTM.

I think we can also improve the containsStringMatcher, but it doesn't have to be this PR.

I also think this is a good candidate to run some [fuzzing](https://go.dev/security/fuzz/)

model/labels/regexp.go
pracucci: We can't just check `j` but we also have to check `i`. Anyway, I committed a slightly modified version of your idea:
https://github.com/grafana/mimir-prometheus/pull/430/commits/8008a119fbb13edda7074ddf0aecf1f62d050491

WDYT?

model/labels/regexp.go
pracucci: I also think it should be straightforward, but I would prefer to do it in a separate PR.

model/labels/regexp_test.go
pracucci: Good suggestions, thanks!

colega (approved):
Thank you!

pracucci:
Going to merge this PR and then will work on fuzzy testing in a follow up PR. I've already found an issue with a basic fuzzy test, but the issue happens in `main` too so it's unrelated from this PR.

model/labels/regexp.go
pstibrany: > just converts the matches to lowercase and passes on the Matches() call, are we going to to this?

This naive approach works for english characters only. Use `strings.EqualFold` at least.

colega: Sure, whatever the regexp's strategy is.

PR #431, opened 2023-03-01
Fix containsStringMatcher.Matches() when the text contains multiple occurrences of a substring

pracucci:
Oleg suggested in #430 to write a fuzzy test for the regexp optimizations and it opened a Pandora's box. I found multiple issues, which I will try to address 1-by-1. In this PR I'm addressing a bug occurring in `containsStringMatcher.Matches()` when there are multiple occurrences of a substring inside the text to match against. In this case, we have to try to match it against all substrings (stopping as soon as we find a valid match), and not just stopping at the first one.

### Example

- Regexp: `.+foo.+`
- Text: `foofoo0`

With the code in `main`, the `FastRegexMatcher.MatchString()` return false, while it should return true. The reason why it returns false is because the code in `main` only check the 1st occurence of the substring `foo` inside `foofoo0` (for which the regexp doesn't match because of the `.+` on the left side). However, if we also check the 2nd occurrence of `foo` inside `foofoo0`, then it matches, because the left `.+` is honored by `foo` at position 0 and the right `.+` is honored by `0`.

### DIFF
--- a/model/labels/regexp_test.go
+++ b/model/labels/regexp_test.go
@@ -19,7 +19,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/grafana/regexp"
 	"github.com/grafana/regexp/syntax"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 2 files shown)


model/labels/regexp_test.go
```diff
@@ -72,9 +72,7 @@ func TestNewFastRegexMatcher(t *testing.T) {
 				t.Parallel()
 				m, err := NewFastRegexMatcher(r)
 				require.NoError(t, err)
-				re, err := regexp.Compile("^(?:" + r + ")$")
```
pracucci, line 75: Note to reviewers: no need to re-compile it. We already have it in `m.re`.

colega (approved):
Wow! LGTM

PR #433, opened 2023-03-01
Do not optimize regexps with begin/end text anchors inside

pracucci:
I've added another fuzz test and found several issues related to how we optimize regexps when they contain the being/end text anchors inside. In the `main` version, we always remove being/end text anchors and we treat the regexp like if there were no anchors because of the assumption that any regex is anchored in Prometheus (which is true).

However, removing the begin/end text anchors from inside is not always safe. For example, if we remove the anchor from `^|x`, the resulting optimization is different because the empty string in one case is valid, while in the other case it's not.

After some experimentation, I found very difficult to correctly handle all begin/end text anchors, so in this PR I propose to not optimize a regexp if it contains them, unless they're really at tbe begin/end of the whole regexp.

I'm a bit sad with this PR because we may stop optimizing some regexps which were already correctly handled. Alternative, may be keep leaving with the bug in `main`, assuming customers don't run regexps like the absurd ones written by the fuzzy test.

I've run both fuzzy tests for few minutes and, with the changes in this PR, they don't report any more issue. However, it's also true that with the changes in this PR, a bunch of regexps generated by the fuzzy test are not optimized anymore.

### DIFF
--- a/model/labels/regexp.go
+++ b/model/labels/regexp.go
@@ -325,6 +325,14 @@ func stringMatcherFromRegexpInternal(re *syntax.Regexp) StringMatcher {
 	clearCapture(re)
 
 	switch re.Op {
+	case syntax.OpBeginText:
+		// Correctly handling the begin text operator inside a regex is tricky,
+		// so in this case we fallback to the regex engine.
+		return nil
+	case syntax.OpEndText:
+		// Correctly handling the end text operator inside a regex is tricky,
+		// so in this case we fallback to the regex engine.
+		return nil
 	case syntax.OpPlus, syntax.OpStar:
 		if re.Sub[0].Op != syntax.OpAnyChar && re.Sub[0].Op != syntax.OpAnyCharNotNL {
 			return nil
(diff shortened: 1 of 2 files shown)


colega:
I think we should aim for 100% correctness, so this change LGTM. 

> However, it's also true that with the changes in this PR, a bunch of regexps generated by the fuzzy test are not optimized anymore.

This surprises me, as fuzzier shouldn't keep trying things that are not interesting. I checked the fuzzier again, and we have statically defined regexs matching on fuzzied values. 
I think we need to fuzz the regexps too.
I added this fuzzier:
```go
func FuzzFastRegexMatcher_WithFuzziedRegexpsAndValues(f *testing.F) {
	for _, re := range regexes {
		for _, v := range values {
			f.Add(re, v)
		}
	}

	f.Fuzz(func(t *testing.T, re, val string) {
		m, err := NewFastRegexMatcher(re)
		if err != nil {
			return // Not interesting
		}
		require.Equalf(t, m.re.MatchString(val), m.MatchString(val), "regexp: %s val: %s", m.re.String(), val)
	})
}
```

And it failed after 3s with:
```
            	Error:      	Not equal: 
            	            	expected: false
            	            	actual  : true
            	Test:       	FuzzFastRegexMatcher_WithFuzziedRegexpsAndValues
            	Messages:   	regexp: ^(?:(?-s:.)*\A0)$ val: 00000000
```

PR #447, opened 2023-03-07
Rules: Increase tolerance for missed iterations on alerts

ying-jeanne:
This PR is to increase the valid until time in rules, since the current EndsAt time is not enough, It could cause resending notifications of alerts in certain cases (when temporary system failure or on system restart). See [Issue](https://github.com/grafana/mimir/issues/4118)


dimitarvdimitrov (approved):
LGTM, the tests are failing, but I suspect it's a flake and not related to your change.

PR #448, opened 2023-03-08
Optimize regex star operation

pracucci:
We have a customer running many queries with label matchers like `label=~".*"`, causing about 20% of ingesters CPU being spent in `FastRegexMatcher.MatchString()`. I want to look if we can optimise it, I started to look at `strings.ContainsRune()`  and ended up with a path I wasn't expecting, which is what I'm proposing in this PR. The performance benefit is not huge, but given doesn't make the logic more complicated, I think we could give it a try.

In this PR:
- Compile the `MatchString()` function. The idea is that we don't have to go through all the `if` statements if the only optimization we have is the `stringMatcher`.
- Replace `anyStringMatcher` with specialised matchers. Each matcher has less checks to do than before.
- Replace `strings.ContainsRune()` with `strings.IndexByte()`. There's no logic difference. Internally, `strings.ContainsRune()` checks if the rune to look for is an ASCII character and, if so, it calls `strings.IndexByte()`. Since we always look for the newline (a static rune), we can directly call `strings.IndexByte()`.

The performance benefit we get is the sum of the 3 changes listed above. I tried to do it 1-by-1, and at each step I got some benefit for the `.*` and `.+` case.

Other than unit tests, I've also run the 2 fuzzy tests, and I've kept them running for few minutes with no errors.

Benchmark (posting only CPU, no difference for memory which has 0 allocs):

```
name                                                  old time/op    new time/op    delta
FastRegexMatcher/foo-12                                 62.0ns ± 5%    64.2ns ± 0%     ~     (p=0.282 n=3+3)
FastRegexMatcher/^foo-12                                61.3ns ± 1%    63.7ns ± 2%   +3.88%  (p=0.026 n=3+3)
FastRegexMatcher/(foo|bar)-12                           68.3ns ± 2%    76.5ns ± 3%  +12.10%  (p=0.008 n=3+3)
FastRegexMatcher/foo.*-12                                122ns ± 1%     116ns ± 1%   -5.17%  (p=0.005 n=3+3)
FastRegexMatcher/.*foo-12                                136ns ± 1%     131ns ± 1%   -4.18%  (p=0.007 n=3+3)
FastRegexMatcher/^.*foo$-12                              135ns ± 2%     135ns ± 0%     ~     (p=0.786 n=3+3)
FastRegexMatcher/^.+foo$-12                              134ns ± 0%     134ns ± 1%     ~     (p=0.744 n=3+3)
FastRegexMatcher/.*-12                                   173ns ± 2%     125ns ± 1%  -27.78%  (p=0.000 n=3+3)
FastRegexMatcher/.+-12                                   171ns ± 3%     128ns ± 3%  -25.35%  (p=0.000 n=3+3)
FastRegexMatcher/foo.+-12                                124ns ± 0%     118ns ± 1%   -5.16%  (p=0.009 n=3+3)
FastRegexMatcher/.+foo-12                                135ns ± 1%     133ns ± 1%     ~     (p=0.076 n=3+3)
FastRegexMatcher/foo_.+-12                               109ns ± 1%     102ns ± 3%   -6.33%  (p=0.017 n=3+3)
FastRegexMatcher/foo_.*-12                               110ns ± 2%     103ns ± 2%   -6.32%  (p=0.012 n=3+3)
FastRegexMatcher/.*foo.*-12                              276ns ± 1%     266ns ± 1%   -3.54%  (p=0.015 n=3+3)
FastRegexMatcher/.+foo.+-12                              286ns ± 2%     277ns ± 2%     ~     (p=0.136 n=3+3)
FastRegexMatcher/(?s:.*)-12                             89.5ns ± 2%    61.2ns ± 3%  -31.67%  (p=0.000 n=3+3)
FastRegexMatcher/(?s:.+)-12                             91.8ns ± 2%    72.2ns ± 2%  -21.29%  (p=0.000 n=3+3)
FastRegexMatcher/(?s:^.*foo$)-12                         132ns ± 3%     128ns ± 1%     ~     (p=0.254 n=3+3)
FastRegexMatcher/(?i:foo)-12                             142ns ± 7%     108ns ± 2%  -23.93%  (p=0.028 n=3+3)
FastRegexMatcher/(?i:(foo|bar))-12                       289ns ± 2%     236ns ± 4%  -18.47%  (p=0.002 n=3+3)
FastRegexMatcher/(?i:(foo1|foo2|bar))-12                 428ns ± 2%     424ns ± 1%     ~     (p=0.413 n=3+3)
FastRegexMatcher/^(?i:foo|oo)|(bar)$-12                 1.00µs ± 3%    1.02µs ± 1%     ~     (p=0.243 n=3+3)
FastRegexMatcher/(?i:(foo1|foo2|aaa|bbb|ccc|ddd|e-12    1.97µs ± 1%    1.96µs ± 1%     ~     (p=0.491 n=3+3)
FastRegexMatcher/((.*)(bar|b|buzz)(.+)|foo)$-12          714ns ± 2%     714ns ± 1%     ~     (p=0.944 n=3+3)
FastRegexMatcher/^$-12                                  86.6ns ± 1%    63.4ns ± 0%  -26.76%  (p=0.001 n=3+3)
FastRegexMatcher/(prometheus|api_prom)_api_v1_.+-12      214ns ± 1%     200ns ± 2%   -6.61%  (p=0.005 n=3+3)
FastRegexMatcher/10\.0\.(1|2)\.+-12                      109ns ± 3%     102ns ± 2%   -6.93%  (p=0.025 n=3+3)
FastRegexMatcher/10\.0\.(1|2).+-12                       108ns ± 1%     101ns ± 3%   -6.24%  (p=0.039 n=3+3)
FastRegexMatcher/((fo(bar))|.+foo)-12                    357ns ± 2%     341ns ± 3%     ~     (p=0.058 n=3+3)
FastRegexMatcher/zQPbMkNO|NNSPdvMi|iWuuSoAl|qbvKM-12     925ns ± 1%    1036ns ± 2%  +11.98%  (p=0.001 n=3+3)
FastRegexMatcher/(?i:(zQPbMkNO|NNSPdvMi|iWuuSoAl|-12    1.95µs ± 1%    1.94µs ± 1%     ~     (p=0.336 n=3+3)
```

### DIFF
--- a/model/labels/regexp.go
+++ b/model/labels/regexp.go
@@ -37,6 +37,9 @@ type FastRegexMatcher struct {
 	prefix        string
 	suffix        string
 	contains      string
+
+	// matchString is the "compiled" function to run by MatchString().
+	matchString func(string) bool
 }
 
 func NewFastRegexMatcher(v string) (*FastRegexMatcher, error) {
@@ -61,7 +64,40 @@ func NewFastRegexMatcher(v string) (*FastRegexMatcher, error) {
 	}
 	m.stringMatcher = stringMatcherFromRegexp(parsed)
 
+	m.matchString = m.compileMatchStringFunction()
 	return m, nil
+}
+
+// compileMatchStringFunction returns the function to run by MatchString().
+func (m *FastRegexMatcher) compileMatchStringFunction() func(string) bool {
+	// If the only optimization available is the string matcher, then we can just run it.
+	if len(m.setMatches) == 0 && m.prefix == "" && m.suffix == "" && m.contains == "" && m.stringMatcher != nil {
+		return m.stringMatcher.Matches
+	}
+
+	return func(s string) bool {
+		if len(m.setMatches) != 0 {
+			for _, match := range m.setMatches {
+				if match == s {
+					return true
+				}
+			}
+			return false
+		}
+		if m.prefix != "" && !strings.HasPrefix(s, m.prefix) {
+			return false
+		}
+		if m.suffix != "" && !strings.HasSuffix(s, m.suffix) {
+			return false
+		}
+		if m.contains != "" && !strings.Contains(s, m.contains) {
+			return false
+		}
+		if m.stringMatcher != nil {
+			return m.stringMatcher.Matches(s)
+		}
+		return m.re.MatchString(s)
+	}
 }
 
 // isOptimized returns true if any fast-path optimization is applied to the
@@ -250,27 +286,7 @@ func tooManyMatches(matches []string, new ...string) bool {
 }
 
 func (m *FastRegexMatcher) MatchString(s string) bool {
-	if len(m.setMatches) != 0 {
-		for _, match := range m.setMatches {
-			if match == s {
-				return true
-			}
-		}
-		return false
-	}
-	if m.prefix != "" && !strings.HasPrefix(s, m.prefix) {
-		return false
-	}
-	if m.suffix != "" && !strings.HasSuffix(s, m.suffix) {
-		return false
-	}
-	if m.contains != "" && !strings.Contains(s, m.contains) {
-		return false
-	}
-	if m.stringMatcher != nil {
-		return m.stringMatcher.Matches(s)
-	}
-	return m.re.MatchString(s)
+	return m.matchString(s)
 }
 
 func (m *FastRegexMatcher) SetMatches() []string {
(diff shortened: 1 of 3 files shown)


pracucci:
> Because .+ is what Grafana sends as All value for templating variables.

Right, but the customer is using `.*` so it's important we optimise `.*` too for the case I'm investigating.

model/labels/regexp.go
```diff
@@ -61,9 +64,42 @@ func NewFastRegexMatcher(v string) (*FastRegexMatcher, error) {
...
+
+	return func(s string) bool {
+		if len(m.setMatches) != 0 {
+			for _, match := range m.setMatches {
+				if match == s {
+					return true
+				}
+			}
+			return false
+		}
```
colega, line 86: We can do this check only once, not per each value checked.

```suggestion
	if len(m.setMatches) != 0 {
		return func(s string) bool {
			for _, match := range m.setMatches {
				if match == s {
					return true
				}
			}
			return false
		}
	}

	return func(s string) bool {
```

model/labels/regexp.go
```diff
@@ -531,20 +558,37 @@ func (m *equalMultiStringMatcher) Matches(s string) bool {
...
+
+	// We need to make sure it non-empty and doesn't contain a newline.
+	// Since the newline is an ASCII character, we can use strings.IndexByte().
+	return len(s) > 0 && strings.IndexByte(s, '\n') == -1
+}
+
+// trueMatcher is a stringMatcher which matches any string (always returns true).
+type trueMatcher struct{}
+
+func (m trueMatcher) Matches(_ string) bool {
```
colega, line 591: In the future we should detect this (expose a `MatchesAll()` method here and check it in the query path) and remove this matcher from the query, as it still would need to enumerate all labels which is a waste of time.

colega (approved):
LGTM, thanks. Hopefully `IndexByte` uses the magic SIMD instructions to make it quicker, but unfortunately we still have to call it on each string separately.

pracucci:
> Hopefully IndexByte uses the magic SIMD instructions to make it quicker, but unfortunately we still have to call it on each string separately.

`strings.ContainsRune()` was already calling `strings.IndexByte()` behind the hood, so no performance difference is expected from this change (see more details in the PR description).

model/labels/regexp.go
colega: It's probably one cpu cycle difference, but why not having it when the complexity of the code is the same? :)

Thank you.

colega (approved):
LGTM, thank you :crossed_fingers:

duricanikolic:
Hi @pracucci @colega,
Shouldn’t we document that this PR breaks deep equality of `label.Matcher`’s? Maybe in `CHANGELOG.md` or in the PR description? 
I realized that while I was [upgrading mimir and mimir-prometheus](https://github.com/grafana/transmog/pull/1722)  in `transmog` and I had to fix some failing tests.

PR #453, opened 2023-03-13
Sync with upstream Prometheus

codesome:
Need https://github.com/prometheus/prometheus/pull/11962


codesome (commented):
I looked at the changes that affect us and it seems to be fine

jesusvazquez (approved):
Also read through the changes and I think they all look good

PR #461, opened 2023-03-23
Revert "Implement Is() for multierrors"

colega:
Reverts grafana/mimir-prometheus#460

I'll send an upstream PR, and cherry-pick those commits here, to ensure we have the same commits.


pracucci (approved):

colega:
See https://github.com/grafana/mimir-prometheus/pull/462 for follow up

PR #462, opened 2023-03-23
Fix block compaction failed when shutting down cherry picked

colega:
Required to solve https://github.com/grafana/mimir/issues/4564, cherry-picked PR from prometheus https://github.com/prometheus/prometheus/pull/12179

--- 

`*LeveledCompactor.Compact` doesn't handle the `context.Canceled` errors properly for two reasons:
- `err` is never `== context.Canceled`: it's wrapped through `errors.Wrap()` and sometimes wrapped through `tsdb_errors.NewMulti(err)`.
- That `tsdb_errors.NewMulti(err)` is opaque to `errors.Is(...)` call.

This causes multiple issues:
- We mark block compaction as failed, which isn't the case.
- We log some extra wrong errors here and there.
- Downstream projects extending Prometheus [get an error](https://github.com/grafana/mimir/issues/4564) that can't be checked to be `errors.Is(err, context.Canceled)` so they misinterpret compaction as failed rather than a normal shutdown.

### DIFF
--- a/tsdb/compact.go
+++ b/tsdb/compact.go
@@ -539,7 +539,7 @@ func (c *LeveledCompactor) compact(dest string, dirs []string, open []*Block, sh
 	}
 
 	errs := tsdb_errors.NewMulti(err)
-	if err != context.Canceled {
+	if !errors.Is(err, context.Canceled) {
 		for _, b := range bs {
 			if err := b.setCompactionFailed(); err != nil {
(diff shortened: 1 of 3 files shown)


pracucci (approved):
👍  Remember to merge with merge commit.

pstibrany:
> 👍 Remember to merge with merge commit.

Cherry-picking creates independent commit already, so this is not a merge of upstream project, and merge-commit was not necessary here.

PR #464, opened 2023-03-30
Improve TestAnalyzeRealQueries

pracucci:
While working on further optimizing regex matcher based on real world queries, I've found `TestAnalyzeRealQueries` being very useful. In the last few hours I've further improved it to compute more statistics and in this PR I'm proposing to upstream them. For each regexp found in the logs, it prints

- Parsing time
- Whether it's optimized or not
- Num split queries
- Num sharded queries
- Min/avg/max query start time difference, between consecutive queries (to understand if any cache with a low TTL would be effective or not)

### DIFF
--- a/model/labels/regexp_test.go
+++ b/model/labels/regexp_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"sort"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


model/labels/regexp_test.go
```diff
@@ -371,8 +373,23 @@ func FuzzFastRegexMatcher_WithFuzzyRegularExpressions(f *testing.F) {
 func TestAnalyzeRealQueries(t *testing.T) {
 	t.Skip("Decomment this test only to manually analyze real queries")
```
colega, line 374: Not really related to your PR, but just to make everyone's life simplier:
- we could do the `os.Open("logs.txt")` here, and do `t.Skip` if it doesn't exist, and also have `logs.txt` in `.gitignore`
- we could check `os.Env("ANALYZE_LOGS_FILE") == ""`, and skip if empty, or open that one if provided, so the usage would be `ANALYZE_LOGS_FILE=/tmp/logs.txt go test -run TestAnalyzeRealQueries`

So we wouldn't need to bother commenting/uncommenting this.

colega (approved):
LGTM

PR #480, opened 2023-04-14
Sync with Prometheus up to 2023-04-14 (7309ac27)

zenador:
This brings changes from prometheus/prometheus up to 2024-04-14 commit https://github.com/prometheus/prometheus/commit/7309ac272195cb856b879306d6a27af7641d3346

Major merge conflicts with rules/manager and tsdb/compact (both code and test), you might want to rewrite some of that, especially the latter.


tsdb/compact.go
```diff
@@ -938,12 +938,16 @@ func timeFromMillis(ms int64) time.Time {
...
+	PopulateBlock(ctx context.Context, metrics *CompactorMetrics, logger log.Logger, chunkPool chunkenc.Pool, mergeFunc storage.VerticalChunkSeriesMergeFunc, concurrencyOpts LeveledCompactorConcurrencyOptions, blocks []BlockReader, minT, maxT int64, outBlocks []shardedBlock) error
+}
+
+type DefaultBlockPopulator struct{}
+
+// PopulateBlock fills the index and chunk writers with new data gathered as the union
+// of the provided blocks. It returns meta information for the new block.
 // It expects sorted blocks input by mint.
-// If there is more than 1 output block, each output block will only contain series that hash into its shard
-// (based on total number of output blocks).
```
krajorama, line 945: I think we need this comment

krajorama (approved):
LGTM, I've taken a closer look at rules directory and tsdb/compact.go as per author request.

Also checked the diff of diffs between prometheus and mimir-prometheus.

For sure needs another set of eyes.

tsdb/compact.go
```diff
@@ -938,12 +938,16 @@ func timeFromMillis(ms int64) time.Time {
 	return time.Unix(0, ms*int64(time.Millisecond))
 }
 
-// populateBlock fills the index and chunk writers of output blocks with new data gathered as the union
-// of the provided blocks.
+type BlockPopulator interface {
+	PopulateBlock(ctx context.Context, metrics *CompactorMetrics, logger log.Logger, chunkPool chunkenc.Pool, mergeFunc storage.VerticalChunkSeriesMergeFunc, concurrencyOpts LeveledCompactorConcurrencyOptions, blocks []BlockReader, minT, maxT int64, outBlocks []shardedBlock) error
```
krajorama, line 942: Note: this is not the signature from upstream, but it's unclear right now how to reuse upstream signature directly. I'd leave a more sophisticated solution to a new PR and live with the current straightforward conflict resolution.

Original: https://github.com/prometheus/prometheus/blob/7309ac272195cb856b879306d6a27af7641d3346/tsdb/compact.go#L667

zenador: @pstibrany since you wrote https://github.com/grafana/mimir-prometheus/pull/13 you'd probably know how to best fix the conflicts. Would it make sense for us to request a signature change from upstream like in https://github.com/prometheus/prometheus/pull/11711 ?

colega: Once we change the signature of this interface, it's not compatible with any implementation provided by upstream anymore, so I'm not sure if there's any benefit of having this interface at all.

I tried to think about this like week, and I was wondering if we could refactor the caller stack leaving this interface as is, and calling it multiple times per each outBlock we have?

We reach this code through `CompactWithBlockPopulator->write`, so maybe we could leave that path as it is in upstream, and how our own implementation for `CompactWithSplittingAndPopulator` that would `writeWithSplitting` and call the populator multiple times? 

My main reason to further think about this is to keep at least the public API of the Compactor & BlockPopulator compatible with upstream, to reduce issues in case we want to import some other third party code that also depends on Prometheus.

colega:
This LGTM, I think we should spend some time thinking about how to tackle the compactor changes, I feel this is more a refactor task than a merge-upstream task. Anyway I'll leave any decision here to @pstibrany as he has more context about the compactor change.

tsdb/compact.go
pstibrany: > @pstibrany since you wrote #13 you'd probably know how to best fix the conflicts. Would it make sense for us to request a signature change from upstream like in [prometheus/prometheus#11711](https://github.com/prometheus/prometheus/pull/11711) ?

When I discussed possibility to upstream #13 with @codesome, he was very skeptical about that, since Prometheus wouldn't use that.

> call the populator multiple times?

That's certainly an option, but requires multiple traversals of the input blocks :(

Perhaps we can have our own extension as `MultiBlockPopulator`, and let compactor work with both `BlockPopulator` (from upstream) and `MultiBlockPopulator`.

pstibrany:
I am fine merging it in the current form, but agree that we should fix this soon and use upstream version of `BlockPopulator`.

One alternative could be to fork our sharding compactor code (reusing as much as possible), and use it from Mimir, while we only keep upstream Prometheus compactor in `mimir-prometheus`.

tsdb/compact.go
zenador: Yep maybe not upstream the whole thing, but just enough so that we wouldn't need to modify the signature in our fork.

krajorama: Add issue for tracking: https://github.com/grafana/mimir-prometheus/issues/484  also added into Native Histograms EPIC so it's not lost

colega: > That's certainly an option, but requires multiple traversals of the input blocks

I was thinking about implementing the `BlockPopulator` interface with a closure created in `writeWithSpliting`, which would do the block traversing only once.

codesome:
@pstibrany if you have any interface change/update/extension suggestions for upstream, please let us know.

PR #486, opened 2023-04-26
Sync upstream 26 apr 2023

zenador:
This brings changes from prometheus/prometheus up to 2024-04-26 commit https://github.com/prometheus/prometheus/commit/d4bf47766e39aef38ad7e94de146ddab1d70766f

Minor merge conflicts only.


krajorama (approved):
LGTM, I've read through the whole thing and couldn't tell where you would have run into a conflict.

krajorama:
Just ran lint locally:
```
$ golangci-lint run
tsdb/chunks/head_chunks_test.go:555:2: redefines-builtin-id: redefinition of the built-in function len (revive)
	len := rand.Int() % 120
	^
tsdb/postings_for_matchers_cache_test.go:264:45: unused-parameter: parameter 'name' seems to be unused, consider removing or renaming it as _ (revive)
func (idx indexForPostingsMock) LabelValues(name string, matchers ...*labels.Matcher) ([]string, error) {
                                            ^
tsdb/postings_for_matchers_cache_test.go:268:42: unused-parameter: parameter 'name' seems to be unused, consider removing or renaming it as _ (revive)
func (idx indexForPostingsMock) Postings(name string, values ...string) (index.Postings, error) {
                                         ^
tsdb/symbols_batch.go:102:2: empty-block: this block is empty, you can remove it (revive)
	for range f.jobs {
		// drain the channel, don't do more flushing. only used when error occurs.
	}
```

Checking if these are Grafana mods (would be simpler if we didn't use main) or maybe fixed upstream

PR #502, opened 2023-05-01
Disable PRs from dependabot

bboreham:
This repo should be be as far as possible identical to upstream Prometheus, so we don't want these PRs.

This is what I found when I looked for instructions to disable [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
-# Disable version updates; we will get them when we update from upstream Prometheus.
+    # Disable version updates; we will get them when we update from upstream Prometheus.
     open-pull-requests-limit: 0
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,19 +4,25 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+# Disable version updates; we will get them when we update from upstream Prometheus.
+    open-pull-requests-limit: 0
```
aknuds1, line 8: [Nit] Aligned indentation for readability:
```suggestion
    # Disable version updates; we will get them when we update from upstream Prometheus.
    open-pull-requests-limit: 0
```

aknuds1 (approved):
Makes sense, thanks!

PR #503, opened 2023-05-03
labels: add stringlabels version of StableHash

bboreham:
This leaves the slice-based implementation alone, and adds an implementation for `-tags stringlabels` which
inlines the `Range()` implementation to avoid adding overhead.


model/labels/sharding_string.go
colega, line 1: Just like in: https://github.com/prometheus/prometheus/pull/12328
I'd suggest (ask? :pray:) calling this with a different suffix than `_string.go`, like `_stringlabels.go`.

colega (approved):
Thank you, @bboreham, this LGTM, please consider renaming the file.

I hope you don't mind that I pushed here a test and a CI definition to test with `stringlabels`.

model/labels/sharding_string.go
bboreham: No objection; it parallels `labels_string.go` but that could be changed upstream.

colega: Yes, I also sent a request to change that upstream: https://github.com/prometheus/prometheus/pull/12328

colega: Renamed here: https://github.com/grafana/mimir-prometheus/pull/503/commits/1f14d27d4006bf55d3e554247251fc4dbd513c20

PR #504, opened 2023-05-04
Test StableHash and implement it for stringlabels

colega:
We need a test that checks that `StableHash` is really stable. I think checking three hashes should be ok.

Also implemented this function for the `stringlabels` build tag.

Finally, added a CI piepline to ensure that this doesn't break anymore.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,3 +28,27 @@ jobs:
 
       - name: Run Tests
         run: GO=/usr/local/go/bin/go make common-test
+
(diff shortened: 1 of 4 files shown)


colega:
Okay, I guess I messed it up when I ran the tests with the tag (I ran them without).
It was too suspicious to see that the hash was valid (as I recalled that it shouldn't be).

colega:
Closing this in favor of #503, where I'll cherry pick the test and the CI job.

PR #505, opened 2023-05-22
Sync upstream 24 may 2023

zenador:
This brings changes from prometheus/prometheus up to 2023-05-24 commit https://github.com/prometheus/prometheus/commit/37e5249e33e4c2d7bc49c4ac587faf12eb6785b7

Minor merge conflicts only.


charleskorn (approved):
LGTM, but there are a lot of changes here, so might be worth getting a second pair of eyes over this.

zenador:
Thanks for the review! Marking as draft now while waiting for the fix in https://github.com/prometheus/prometheus/pull/12387 to be merged, then will merge from upstream again. Other than that, this should be fine if https://github.com/grafana/mimir/pull/5049 passes all tests (still running).

tsdb/head.go
```diff
@@ -1239,9 +1246,9 @@ func (h *Head) truncateWAL(mint int64) error {
 			return true
 		}
 		h.deletedMtx.Lock()
-		_, ok := h.deleted[id]
+		keepUntil, ok := h.deleted[id]
 		h.deletedMtx.Unlock()
-		return ok
+		return ok && keepUntil > last
```
bboreham, line 1251: I'm hoping to see a reduction in 2-hourly CPU spikes due to this change.

PR #507, opened 2023-06-29
Remediate PostingsForMatchers race condition with appends

dimitarvdimitrov:
PostingsForMatchers has a race condition with appending to the TSDB head block.
PostingsForMatchers returns the list of series IDs for some label matchers.

### Bug

This bug can manifest when a new series is appended to the head block while PostingsForMatchers is computing its result. Then the result from PostingsForMatchers can contains series references that do not match the matchers.

This is less likely to affect Prometheus because there most queries request the `__name__` label and it is generally computed first.

### Example

The query is 
```
{datacenter!="dc1", datacenter=~"dc.*"}
```

PostingsForMatchers computes its result like so
* find the series IDs for `datacenter!=dc1`; call this $D1$
* find the series IDs for `datacenter=~dc.*`; call this $D2$
* return $P = D2 - D1$

If there is an append with a new timeseries between computing $D1$ and $D2$, then $P$ can contain more series than is correct.

* find the series IDs for `datacenter!=dc1`; call this $D1$
* __in another request__ append a new timeseries to the TSDB head `{datacenter="dc1"}`
* find the series IDs for `datacenter=~dc.*`; call this $D2$
* return $P = D2 - D1$

There is no synchronisation between the calculation of the different series sets, so now $D2$ contains the new timeseries, but $D1$ doesn't.

### Tests

I added two test cases that exemplify the bug. Without the fix the first one - `n=""` triggers way more often, for the second, `n!="495",n=~".+"` needs a few hundred runs to fail (`go test ... -count=500`).

### Remediation

Always compute the series IDs for intersecting matchers first and for subtracting matchers second. This means that intersected sets of series will not contain new series, so even if the subtracting sets contain new series, then the result is still valid (only slightly outdated).

In typical Prometheus queries `__name__`, which is an intersecting matcher is computed first, therefore this race condition doesn't manifest.

### DIFF
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
+	"golang.org/x/exp/slices"
 
 	"github.com/prometheus/prometheus/model/histogram"
 	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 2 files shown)


colega:
As mentioned offline: this is probably not a big deal for fetching series in Prometheus because with isolation enabled, even if the newly appended series is returned from `PostingsForMatchers`, it will be discarded later.

However, I also see that headIndexReader doesn't take into account the isolation state, so it will not filter out those wrong results for index-only queries, like `LabelValues`. Also it's probably important to mention in Prometheus that there's no isolation on the index-only operations at all (although not having isolation isn't as bad as returning results that don't even match the query).

Finally, I wonder if this will have any performance impact, especially on code that we (Grafana) might have relying on the lexicographical order of matchers evaluation (is it really that order? need to double check).

dimitarvdimitrov:
When using a querier instead of only PostingsForMatchers, I couldn't reproduce the incorrect results neither with nor without isolation. I'll close this for now.

dimitarvdimitrov:
ok I think I found out why the tests with the regular querier (even _with_ isolation) weren't failing. I believe I hadn't properly undone the fix when I changed the test.

I reverted both of the fixes with https://github.com/grafana/mimir-prometheus/pull/507/commits/c78e3d8884bc4b26915bb5d3137f15d26a94df87 and https://github.com/grafana/mimir-prometheus/pull/507/commits/f377bca119769ea1614608364cf6364ee0d6cac1 and enabled isolation in the test TestQuerierIsolationRace with https://github.com/grafana/mimir-prometheus/pull/507/commits/5967476c309030d26db63c19a003b4a7c49f5de2. In that commit the test is failing - [example](https://github.com/grafana/mimir-prometheus/actions/runs/5523118838/jobs/10073596579?pr=507)

I am able to reproduce quite easily with 

```
go test ./tsdb/... -run=TestQuerierIsolationRace -count=1
```


Can you review the test if it's not testing something incorrectly?

P.S. thanks @colega for making me revisit this

dimitarvdimitrov:
This is superseded by https://github.com/prometheus/prometheus/pull/12558

PR #513, opened 2023-07-05
Add `ChunkCount()` method to `ChunkSeries`

charleskorn:
See https://github.com/grafana/mimir/pull/5369.

### DIFF
--- a/storage/interface.go
+++ b/storage/interface.go
@@ -415,6 +415,12 @@ type ChunkSeriesSet interface {
 type ChunkSeries interface {
 	Labels
 	ChunkIterable
+
+	// EstimatedChunkCount returns an estimate of the number of chunks available from this ChunkSeries.
+	//
+	// This estimate is used by Mimir's ingesters to report the number of chunks expected to be returned by a query,
+	// which is used by queriers to enforce the 'max chunks per query' limit.
+	EstimatedChunkCount() int
 }
 
 // Labels represents an item that has labels e.g. time series.
--- a/storage/interface.go
+++ b/storage/interface.go
@@ -420,7 +420,7 @@ type ChunkSeries interface {
 	//
 	// This estimate is used by Mimir's ingesters to report the number of chunks expected to be returned by a query,
 	// which is used by queriers to enforce the 'max chunks per query' limit.
-	EstimatedChunkCount() int
+	EstimatedChunkCount() (int, error)
 }
 
 // Labels represents an item that has labels e.g. time series.
--- a/storage/interface.go
+++ b/storage/interface.go
@@ -416,11 +416,11 @@ type ChunkSeries interface {
 	Labels
 	ChunkIterable
 
-	// EstimatedChunkCount returns an estimate of the number of chunks available from this ChunkSeries.
+	// ChunkCount returns the number of chunks available from this ChunkSeries.
 	//
-	// This estimate is used by Mimir's ingesters to report the number of chunks expected to be returned by a query,
+	// This value is used by Mimir's ingesters to report the number of chunks expected to be returned by a query,
 	// which is used by queriers to enforce the 'max chunks per query' limit.
-	EstimatedChunkCount() (int, error)
+	ChunkCount() (int, error)
 }
 
 // Labels represents an item that has labels e.g. time series.
--- a/storage/merge.go
+++ b/storage/merge.go
@@ -658,6 +658,7 @@ func NewCompactingChunkSeriesMerger(mergeFunc VerticalSeriesMergeFunc) VerticalC
 		if len(series) == 0 {
 			return nil
 		}
+
 		return &ChunkSeriesEntry{
 			Lset: series[0].Labels(),
 			ChunkIteratorFn: func(chunks.Iterator) chunks.Iterator {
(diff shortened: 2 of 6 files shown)


aknuds1 (approved):
I lack familiarity, but LGTM FWIW.

storage/interface.go
```diff
@@ -415,6 +415,12 @@ type ChunkSeriesSet interface {
 type ChunkSeries interface {
 	Labels
 	ChunkIterable
+
+	// EstimatedChunkCount returns an estimate of the number of chunks available from this ChunkSeries.
```
krajorama, line 419: I think this should specify the requirement - if applicable - whether the estimate should be greater or equal or less or equal of the actual count. Is there such implicit requirement or can the estimate return anything?

storage/merge.go
```diff
@@ -670,8 +671,84 @@ func NewCompactingChunkSeriesMerger(mergeFunc VerticalSeriesMergeFunc) VerticalC
...
+//     duplicated
+//   - it does not check if overlapping chunks have samples that swap back and forth between
+//     different encodings over time, and so may underestimate the resulting number of chunks
+func estimateCompactedChunkCount(series []ChunkSeries) int {
+	h := make(chunkIteratorHeap, 0, len(series))
+
+	for _, s := range series {
+		iter := s.Iterator(nil)
+		if iter.Next() {
+			h.Push(iter)
```
krajorama, line 696: Is this supposed to behave as a heap? Cause if I understand correctly you'd need to use the standard heap functions: https://pkg.go.dev/container/heap#Init 
Right now I think it doesn't order the elements in it - I'd just change to a slice if heap is not actually needed.

storage/merge.go
```diff
@@ -670,8 +671,84 @@ func NewCompactingChunkSeriesMerger(mergeFunc VerticalSeriesMergeFunc) VerticalC
...
+				heap.Push(&h, iter)
+			}
+
+			if next.MaxTime > maxTime {
+				maxTime = next.MaxTime
+			}
+
+			if prev.Chunk.Encoding() != next.Chunk.Encoding() {
+				// If we have more than seriesToChunkEncoderSplit samples, account for the additional chunks we'll create.
+				chunkCountForThisTimePeriod += sampleCount / seriesToChunkEncoderSplit
```
krajorama, line 727: Note: there was talk about introducing size limit instead of sample count limit for histogram chunks - not implemented yet. That would mean underestimating the chunk count, which is fine.

krajorama (changes requested):
Do I see correctly that we're going to iterate all samples? seriesToChunkEncoder.EstimatedChunkCount suggests so. 
That would mean that we touch the chunk, so we might as well just fetch the chunk count instead of estimating it. Also even better we could count up the size of chunks which can be all over the place due to OOO and native histograms.

I think this PR could be used for normal float series situations, but I think we should revisit it to look at actual chunk sizes and not estimate based on sample count.

storage/merge.go
codesome: If chunks are sorted by their mint, then we don't need the heap

storage/series.go
```diff
@@ -399,6 +402,41 @@ func (s *seriesToChunkEncoder) Iterator(it chunks.Iterator) chunks.Iterator {
...
 }
 
+// EstimatedChunkCount returns an estimate of the number of chunks produced by Iterator.
+//
+// It is perfectly accurate except when histograms are present in series. When histograms are
+// present, EstimatedChunkCount will underestimate the number of chunks produced, as the
+// estimation does not consider individual samples and so triggers for new chunks such as
+// counter resets, changes to the bucket schema and changes to the zero threshold are not
+// taken into account.
+func (s *seriesToChunkEncoder) EstimatedChunkCount() int {
```
codesome, line 412: Assuming this is called everytime for limits, this will effectively make us iterate on all the samples twice. I don't have a good way to make that more efficient at the moment (other than maybe `Series` interface getting maybe a `NumSamples()`), so noting this for the future in case we see any regression.

storage/series.go
```diff
@@ -399,6 +402,41 @@ func (s *seriesToChunkEncoder) Iterator(it chunks.Iterator) chunks.Iterator {
...
+		if chunkCount == 0 {
+			// We'll always have at least one chunk if there's at least one sample.
+			chunkCount++
+		} else if lastType != typ {
+			// If the sample type changes, then we'll cut a new chunk.
+			chunkCount++
+			samplesInChunk = 0
+		}
+
+		if samplesInChunk == seriesToChunkEncoderSplit {
```
codesome, line 428: There is an off-by-1 case if we end up having `seriesToChunkEncoderSplit` number of samples in the last chunk.

storage/interface.go
charleskorn: It's really hard to set either requirement, so I've deliberately left it open. Happy to add that as a comment if you think it's necessary.

storage/merge.go
charleskorn: >Is this supposed to behave as a heap? Cause if I understand correctly you'd need to use the standard heap functions: https://pkg.go.dev/container/heap#Init

Good catch - this should be using `heap.Push`, not `h.Push`. (`Push`ing each element is the same as creating the slice and then calling `Init`, and `compactChunkIterator.Next()` does the same thing, so I'd like to keep them consistent, even though it's more expensive.)

> If chunks are sorted by their mint, then we don't need the heap

`compactChunkIterator.Next()` (which this estimation function tries to emulate) doesn't assume the series are sorted, so we can't either.

storage/merge.go
```diff
@@ -670,8 +671,84 @@ func NewCompactingChunkSeriesMerger(mergeFunc VerticalSeriesMergeFunc) VerticalC
...
+}
+
+// estimateCompactedChunkCount computes an estimate of the resulting number of chunks
+// after compacting series.
+//
+// The estimate is imperfect in a few ways, due to the fact it does not examine individual samples:
+//   - it does not check for duplicate samples in overlapping chunks, and so may overestimate
+//     the resulting number of chunks if duplicate samples are present, or if an entire chunk is
+//     duplicated
+//   - it does not check if overlapping chunks have samples that swap back and forth between
```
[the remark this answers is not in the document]
charleskorn: Good point, I've added a comment in https://github.com/grafana/mimir-prometheus/commit/f71a97b460a691074fbf6766798b3377b8fa21f6

storage/merge.go
charleskorn: That's fine, we can also improve the estimation once that's in place.

storage/series.go
charleskorn: An alternative to both of these more involved estimates (`seriesToChunkEncoder` and `NewCompactingChunkSeriesMerger`) is to just materialise the chunks and count them. 

Given neither of these more involved estimates seem to actually be used on the ingester query path in Mimir (both based on setting breakpoints on a local instance and also examining profiles from real prod cells), I'm tempted to try that as a simpler first approach and then we can try to more cheaply estimate the number of chunks if this turns out to be problematic.

charleskorn:
Based on tracing through the code a bit more, as well as verifying with a debugger, I don't believe `EstimatedChunkCount()` will ever be called on anything apart from `chunkSeriesEntry`. 

So to simplify this PR, I've replaced the complex estimations with simple but expensive iteration of the final chunks themselves for `seriesToChunkEncoder` and `NewCompactingChunkSeriesMerger`. These should never actually be used, but if my understanding turns out to be incorrect, then these will at least still work and we can optimise if need be.

krajorama (approved):
LGTM. Note: in case of native histograms, we'd need https://github.com/prometheus/prometheus/pull/12054 implemented to get best memory overload protection out of it, because currently histogram chunks can be 1K-64K.

charleskorn:
Given the approval from @krajorama, I'm going to merge this, but if anyone has any late feedback, let me know.

PR #518, opened 2023-07-25
Optimize `labelValuesWithMatchers` to fetch label values from selected series

pstibrany:
This PR optimizes labelValues operation with matchers by iterating over all series matched by provided matchers, and collecting unique label values.

On its own this is pretty expensive operation, so this is only activated when number of matched series is tiny compared to number of all values for given label name.

Specifically, we require that queried label has at least 1000 unique label values, and matchers matched less than `(number of label values / 100)` series. (I did not do any analysis to compute these numbers, and I'm open to suggestions how to come with better ones.)

Benchmark for labelValuesForMatchers (updated for https://github.com/grafana/mimir-prometheus/pull/518/commits/e4d815d2383ba307c24a2cc72d55897d1e58179a):

```
goos: darwin
goarch: arm64
                                                                         │   before.txt   │              after.txt              │
                                                                         │     sec/op     │    sec/op     vs base               │
Querier/Head/labelValuesWithMatchers/i_with_i="1"-10                          1.159m ± 3%   1.142m ±  3%        ~ (p=0.310 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1"-10                          97.22m ± 4%   95.98m ±  4%        ~ (p=0.485 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="^.+$"-10                       119.8m ± 4%   119.4m ± 17%        ~ (p=0.937 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j!="foo"-10                 113.4m ± 4%   112.2m ±  3%        ~ (p=0.180 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-10               18295.8µ ± 6%   908.4µ ±  2%  -95.04% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-10           17927.7µ ± 6%   902.2µ ±  2%  -94.97% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="X",j!="foo"-10                 90.19m ± 5%   83.65m ±  3%   -7.25% (p=0.004 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-10       126.8m ± 7%   120.4m ±  8%        ~ (p=0.180 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=533701-10            375686.4µ ± 7%   893.0µ ±  2%  -99.76% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=20-10                233119.2µ ± 7%   898.4µ ±  4%  -99.61% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-10             387.8m ± 5%   403.0m ±  3%        ~ (p=0.093 n=6)
Querier/Head/labelValuesWithMatchers/n_with_j!="foo"-10                       248.6m ± 4%   247.8m ±  2%        ~ (p=0.310 n=6)
Querier/Head/labelValuesWithMatchers/n_with_i="1"-10                          5.301µ ± 1%   5.310µ ±  2%        ~ (p=0.937 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i="1"-10                         1.325m ± 1%   1.323m ±  2%        ~ (p=0.818 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1"-10                         86.92m ± 1%   89.00m ±  1%   +2.39% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="^.+$"-10                      101.2m ± 2%   101.4m ±  1%        ~ (p=0.818 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j!="foo"-10                152.4m ± 2%   159.0m ±  1%   +4.31% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-10               45.283m ± 1%   1.193m ±  2%  -97.37% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-10           45.355m ± 1%   1.205m ±  3%  -97.34% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="X",j!="foo"-10                227.7m ± 1%   231.8m ±  1%   +1.80% (p=0.009 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-10      155.0m ± 1%   162.3m ±  1%   +4.69% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=533701-10            123.369m ± 1%   1.098m ±  1%  -99.11% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=20-10                103.955m ± 1%   1.101m ±  2%  -98.94% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-10            114.7m ± 0%   119.7m ±  1%   +4.30% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_j!="foo"-10                      293.9m ± 2%   292.7m ±  1%   -0.41% (p=0.015 n=6)
Querier/Block/labelValuesWithMatchers/n_with_i="1"-10                         2.358m ± 1%   2.358m ±  1%        ~ (p=0.485 n=6)
geomean                                                                       47.22m        12.73m        -73.05%

                                                                         │  before.txt   │               after.txt               │
                                                                         │     B/op      │     B/op      vs base                 │
Querier/Head/labelValuesWithMatchers/i_with_i="1"-10                        1.531Mi ± 0%   1.531Mi ± 0%        ~ (p=1.000 n=6) ¹
Querier/Head/labelValuesWithMatchers/i_with_n="1"-10                        16.14Mi ± 0%   16.15Mi ± 0%   +0.05% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="^.+$"-10                     32.31Mi ± 0%   32.32Mi ± 0%   +0.02% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j!="foo"-10               16.14Mi ± 0%   16.15Mi ± 0%   +0.05% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-10              10.695Mi ± 0%   1.539Mi ± 0%  -85.61% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-10          10.695Mi ± 0%   1.539Mi ± 0%  -85.61% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="X",j!="foo"-10               16.14Mi ± 0%   16.15Mi ± 0%   +0.05% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-10     17.67Mi ± 0%   17.68Mi ± 0%   +0.04% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=533701-10            10.695Mi ± 0%   1.539Mi ± 0%  -85.61% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=20-10                10.695Mi ± 0%   1.540Mi ± 0%  -85.60% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-10           19.89Mi ± 0%   19.90Mi ± 0%   +0.04% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/n_with_j!="foo"-10                     5.532Ki ± 0%   5.532Ki ± 0%        ~ (p=1.000 n=6)
Querier/Head/labelValuesWithMatchers/n_with_i="1"-10                        4.445Ki ± 0%   4.445Ki ± 0%        ~ (p=1.000 n=6) ¹
Querier/Block/labelValuesWithMatchers/i_with_i="1"-10                       1.531Mi ± 0%   1.531Mi ± 0%        ~ (p=1.000 n=6) ¹
Querier/Block/labelValuesWithMatchers/i_with_n="1"-10                       17.66Mi ± 0%   17.67Mi ± 0%   +0.04% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="^.+$"-10                    33.84Mi ± 0%   33.85Mi ± 0%   +0.02% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j!="foo"-10              17.66Mi ± 0%   17.67Mi ± 0%   +0.04% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-10             12.222Mi ± 0%   1.540Mi ± 0%  -87.40% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-10         12.221Mi ± 0%   1.539Mi ± 0%  -87.40% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="X",j!="foo"-10              17.66Mi ± 0%   17.67Mi ± 0%   +0.04% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-10    19.19Mi ± 0%   19.20Mi ± 0%   +0.04% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=533701-10           12.221Mi ± 0%   1.540Mi ± 0%  -87.40% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=20-10               12.221Mi ± 0%   1.540Mi ± 0%  -87.40% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-10          21.41Mi ± 0%   21.42Mi ± 0%   +0.04% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_j!="foo"-10                    7.078Ki ± 0%   7.078Ki ± 0%        ~ (p=1.000 n=6) ¹
Querier/Block/labelValuesWithMatchers/n_with_i="1"-10                       5.977Ki ± 0%   5.977Ki ± 0%        ~ (p=1.000 n=6) ¹
geomean                                                                     3.902Mi        2.106Mi       -46.03%
¹ all samples are equal

                                                                         │   before.txt    │               after.txt               │
                                                                         │    allocs/op    │  allocs/op   vs base                  │
Querier/Head/labelValuesWithMatchers/i_with_i="1"-10                            5.000 ± 0%    5.000 ± 0%         ~ (p=1.000 n=6) ¹
Querier/Head/labelValuesWithMatchers/i_with_n="1"-10                           200.0k ± 0%   200.0k ± 0%    +0.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="^.+$"-10                        300.1k ± 0%   300.1k ± 0%    +0.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j!="foo"-10                  200.0k ± 0%   200.0k ± 0%    +0.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-10              200010.000 ± 0%    9.000 ± 0%  -100.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-10           200012.00 ± 0%    11.00 ± 0%   -99.99% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="X",j!="foo"-10                  200.0k ± 0%   200.0k ± 0%    +0.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-10        200.0k ± 0%   200.0k ± 0%    +0.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=533701-10            200009.000 ± 0%    8.000 ± 0%  -100.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=20-10                 200011.00 ± 0%    11.00 ± 0%   -99.99% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-10              208.9k ± 0%   208.9k ± 0%    +0.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/n_with_j!="foo"-10                         98.00 ± 0%    98.00 ± 0%         ~ (p=1.000 n=6) ¹
Querier/Head/labelValuesWithMatchers/n_with_i="1"-10                            87.00 ± 0%    87.00 ± 0%         ~ (p=1.000 n=6) ¹
Querier/Block/labelValuesWithMatchers/i_with_i="1"-10                           7.000 ± 0%    7.000 ± 0%         ~ (p=1.000 n=6) ¹
Querier/Block/labelValuesWithMatchers/i_with_n="1"-10                          300.0k ± 0%   300.0k ± 0%    +0.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="^.+$"-10                       400.1k ± 0%   400.1k ± 0%    +0.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j!="foo"-10                 300.0k ± 0%   300.0k ± 0%    +0.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-10              300013.00 ± 0%    12.00 ± 0%  -100.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-10          300012.00 ± 0%    11.00 ± 0%  -100.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="X",j!="foo"-10                 300.0k ± 0%   300.0k ± 0%    +0.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-10       300.0k ± 0%   300.0k ± 0%    +0.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=533701-10            300011.00 ± 0%    16.00 ± 0%   -99.99% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=20-10                300013.00 ± 0%    31.00 ± 0%   -99.99% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-10             308.9k ± 0%   308.9k ± 0%    +0.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_j!="foo"-10                        141.0 ± 0%    141.0 ± 0%         ~ (p=1.000 n=6) ¹
Querier/Block/labelValuesWithMatchers/n_with_i="1"-10                           129.0 ± 0%    129.0 ± 0%         ~ (p=1.000 n=6) ¹
geomean                                                                        34.09k        1.627k        -95.23%
¹ all samples are equal
```

### DIFF
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -346,10 +346,7 @@ func inversePostingsForMatcher(ix IndexPostingsReader, m *labels.Matcher) (index
 	return ix.Postings(m.Name, res...)
 }
 
-const (
-	minSeriesForPostingsLabelValues = 1000 // Minimum number of label values when we try to use labels from series instead.
-	maxExpandedFactor               = 100  // Division factor for maximum number of matched series.
-)
+const maxExpandedPostingsFactor = 100 // Division factor for maximum number of matched series.
 
 func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Matcher) ([]string, error) {
 	p, err := PostingsForMatchers(r, matchers...)
@@ -360,36 +357,6 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 	allValues, err := r.LabelValues(name)
 	if err != nil {
 		return nil, errors.Wrapf(err, "fetching values of label %s", name)
-	}
-
-	// If there are many values, let's see if expanded postings for matchers have smaller cardinality.
-	if len(allValues) >= minSeriesForPostingsLabelValues {
-		// We will only expand this number of postings.
-		maxPostings := len(allValues) / maxExpandedFactor
-		// Add space for one extra posting when checking if we expanded all.
-		expanded := make([]storage.SeriesRef, 0, maxPostings+1)
-
-		for len(expanded) < maxPostings && p.Next() {
-			expanded = append(expanded, p.At())
-		}
-
-		// Did we reach end of postings?
-		reachedEnd := true
-		if len(expanded) == maxPostings && p.Next() {
-			reachedEnd = false
-			expanded = append(expanded, p.At())
-		}
-
-		if reachedEnd {
-			// We have expanded all the postings. Given that number of postings is
-			// much smaller than number of all label values, let's compute label
-			// values from series instead.
-			// (If we did this for all postings, it would be very expensive.)
-			return labelValuesFromSeries(r, name, expanded)
-		}
-
-		// If we haven't reached end of postings, we prepend our expanded postings to "p", and continue.
-		p = newPrependPostings(expanded, p)
 	}
 
 	// If we have a matcher for the label name, we can filter out values that don't match
@@ -409,6 +376,34 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 			}
 		}
 		allValues = filteredValues
+	}
+
+	// Let's see if expanded postings for matchers have smaller cardinality than label values.
+	// Since computing label values from series is expensive, we apply a limit on number of expanded
+	// postings (and series).
+	maxExpandedPostings := len(allValues) / maxExpandedPostingsFactor
+	if maxExpandedPostings > 0 {
+		// Add space for one extra posting when checking if we expanded all postings.
+		expanded := make([]storage.SeriesRef, 0, maxExpandedPostings+1)
+
+		for len(expanded) < maxExpandedPostings && p.Next() {
+			expanded = append(expanded, p.At())
+		}
+
+		// Did we reach end of postings?
+		reachedEnd := true
+		if len(expanded) == maxExpandedPostings && p.Next() {
+			reachedEnd = false
+			expanded = append(expanded, p.At())
+		}
+
+		if reachedEnd {
+			// We have expanded all the postings -- all returned label values will be from these series only.
+			return labelValuesFromSeries(r, name, expanded)
+		}
+
+		// If we haven't reached end of postings, we prepend our expanded postings to "p", and continue.
+		p = newPrependPostings(expanded, p)
 	}
 
 	valuesPostings := make([]index.Postings, len(allValues))
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -386,18 +386,12 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 		// Add space for one extra posting when checking if we expanded all postings.
 		expanded := make([]storage.SeriesRef, 0, maxExpandedPostings+1)
 
-		for len(expanded) < maxExpandedPostings && p.Next() {
+		// Call p.Next() even if len(expanded) == maxExpandedPostings. This tells us if there are more postings or not.
+		for len(expanded) <= maxExpandedPostings && p.Next() {
 			expanded = append(expanded, p.At())
 		}
 
-		// Did we reach the end of postings?
-		reachedEnd := true
-		if len(expanded) == maxExpandedPostings && p.Next() {
-			reachedEnd = false
-			expanded = append(expanded, p.At())
-		}
-
-		if reachedEnd {
+		if len(expanded) <= maxExpandedPostings {
 			// When we're here, p.Next() must have returned false, so we need to check for errors.
 			if err := p.Err(); err != nil {
 				return nil, errors.Wrap(err, "expanding postings for matchers")
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -390,7 +390,7 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 			expanded = append(expanded, p.At())
 		}
 
-		// Did we reach end of postings?
+		// Did we reach the end of postings?
 		reachedEnd := true
 		if len(expanded) == maxExpandedPostings && p.Next() {
 			reachedEnd = false
@@ -398,6 +398,11 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 		}
 
 		if reachedEnd {
+			// When we're here, p.Next() must have returned false, so we need to check for errors.
+			if err := p.Err(); err != nil {
+				return nil, errors.Wrap(err, "expanding postings for matchers")
+			}
+
 			// We have expanded all the postings -- all returned label values will be from these series only.
 			return labelValuesFromSeries(r, name, expanded)
 		}
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -398,7 +398,7 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 			}
 
 			// We have expanded all the postings -- all returned label values will be from these series only.
-			// (We supply allValues[:0] as a buffer for storing results. It should be big enough already, since it holds all possible label values.)
+			// (We supply allValues as a buffer for storing results. It should be big enough already, since it holds all possible label values.)
 			return labelValuesFromSeries(r, name, expanded, allValues)
 		}
 
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -346,6 +346,11 @@ func inversePostingsForMatcher(ix IndexPostingsReader, m *labels.Matcher) (index
 	return ix.Postings(m.Name, res...)
 }
 
+const (
+	minSeriesForPostingsLabelValues = 1000 // Minimum number of label values when we try to use labels from series instead.
+	maxExpandedFactor               = 100  // Division factor for maximum number of matched series.
+)
+
 func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Matcher) ([]string, error) {
 	p, err := PostingsForMatchers(r, matchers...)
 	if err != nil {
@@ -355,6 +360,36 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 	allValues, err := r.LabelValues(name)
 	if err != nil {
 		return nil, errors.Wrapf(err, "fetching values of label %s", name)
+	}
+
+	// If there are many values, let's see if expanded postings for matchers have smaller cardinality.
+	if len(allValues) >= minSeriesForPostingsLabelValues {
+		// We will only expand this number of postings.
+		maxPostings := len(allValues) / maxExpandedFactor
+		// Add space for one extra posting when checking if we expanded all.
+		expanded := make([]storage.SeriesRef, 0, maxPostings+1)
+
+		for len(expanded) < maxPostings && p.Next() {
+			expanded = append(expanded, p.At())
+		}
+
+		// Did we reach end of postings?
+		reachedEnd := true
+		if len(expanded) == maxPostings && p.Next() {
+			reachedEnd = false
+			expanded = append(expanded, p.At())
+		}
+
+		if reachedEnd {
+			// We have expanded all the postings. Given that number of postings is
+			// much smaller than number of all label values, let's compute label
+			// values from series instead.
+			// (If we did this for all postings, it would be very expensive.)
+			return labelValuesFromSeries(r, name, expanded)
+		}
+
+		// If we haven't reached end of postings, we prepend our expanded postings to "p", and continue.
+		p = newPrependPostings(expanded, p)
 	}
 
 	// If we have a matcher for the label name, we can filter out values that don't match
@@ -394,6 +429,78 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 	}
 
 	return values, nil
+}
+
+// labelValuesFromSeries returns all unique label values from for given label name from supplied series. Values are not sorted.
+func labelValuesFromSeries(r IndexReader, labelName string, refs []storage.SeriesRef) ([]string, error) {
+	values := map[string]struct{}{}
+
+	var builder labels.ScratchBuilder
+	for _, ref := range refs {
+		err := r.Series(ref, &builder, nil)
+		if err != nil {
+			return nil, errors.Wrapf(err, "label values for label %s", labelName)
+		}
+
+		v := builder.Labels().Get(labelName)
+		if v != "" {
+			values[v] = struct{}{}
+		}
+	}
+
+	result := make([]string, 0, len(values))
+	for v := range values {
+		result = append(result, v)
+	}
+	return result, nil
+}
+
+func newPrependPostings(a []storage.SeriesRef, b index.Postings) index.Postings {
+	return &prependPostings{
+		ix:     -1,
+		prefix: a,
+		rest:   b,
+	}
+}
+
+// prependPostings returns series references from "prefix" before using "rest" postings.
+type prependPostings struct {
+	ix     int
+	prefix []storage.SeriesRef
+	rest   index.Postings
+}
+
+func (p *prependPostings) Next() bool {
+	p.ix++
+	if p.ix < len(p.prefix) {
+		return true
+	}
+	return p.rest.Next()
+}
+
+func (p *prependPostings) Seek(v storage.SeriesRef) bool {
+	for p.ix < len(p.prefix) {
+		if p.ix >= 0 && p.prefix[p.ix] >= v {
+			return true
+		}
+		p.ix++
+	}
+
+	return p.rest.Seek(v)
+}
+
+func (p *prependPostings) At() storage.SeriesRef {
+	if p.ix >= 0 && p.ix < len(p.prefix) {
+		return p.prefix[p.ix]
+	}
+	return p.rest.At()
+}
+
+func (p *prependPostings) Err() error {
+	if p.ix >= 0 && p.ix < len(p.prefix) {
+		return nil
+	}
+	return p.rest.Err()
 }
 
 func labelNamesWithMatchers(r IndexReader, matchers ...*labels.Matcher) ([]string, error) {
--- a/tsdb/querier.go
+++ b/tsdb/querier.go
@@ -404,7 +404,8 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 			}
 
 			// We have expanded all the postings -- all returned label values will be from these series only.
-			return labelValuesFromSeries(r, name, expanded)
+			// (We supply allValues[:0] as a buffer for storing results. It should be big enough already, since it holds all possible label values.)
+			return labelValuesFromSeries(r, name, expanded, allValues)
 		}
 
 		// If we haven't reached end of postings, we prepend our expanded postings to "p", and continue.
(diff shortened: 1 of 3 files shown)


tsdb/querier.go
```diff
@@ -357,6 +362,36 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 		return nil, errors.Wrapf(err, "fetching values of label %s", name)
 	}
 
+	// If there are many values, let's see if expanded postings for matchers have smaller cardinality.
+	if len(allValues) >= minSeriesForPostingsLabelValues {
```
pracucci, line 366: I think this block should be moved after the next `for` loop (the one used to run the matcher on `__name__`).

pracucci, line 366: Do we need this condition at all? Couldn't just be based on `maxExpandedFactor`?

tsdb/querier.go
```diff
@@ -357,6 +362,36 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
...
+		// We will only expand this number of postings.
+		maxPostings := len(allValues) / maxExpandedFactor
```
pracucci, line 368: [food for thought] We have a similar logic in the store-gateway (I mean to query label values from expanded postings) but the trigger logic is a bit different there (and was based on several testing we did):

https://github.com/grafana/mimir/blob/20327509d198b8a71ad8906f0617702cc9aa3f2c/pkg/storegateway/bucket_index_postings.go#L523-L525

The main problem replicating a similar logic here (not the exact same) is that we miss the estimated number of series matching each label value. It's trivial to know for the TSDB head, but it's non trivial to guess when querying the TSDB block index (I mean, we've already done it in the store-gateway, but we would have to replicate the same logic here).

As a follow up work, would be interesting investigating if we can match the store-gateway trigger logic here.

tsdb/querier.go
```diff
@@ -346,6 +346,11 @@ func inversePostingsForMatcher(ix IndexPostingsReader, m *labels.Matcher) (index
 	return ix.Postings(m.Name, res...)
 }
 
+const (
+	minSeriesForPostingsLabelValues = 1000 // Minimum number of label values when we try to use labels from series instead.
+	maxExpandedFactor               = 100  // Division factor for maximum number of matched series.
```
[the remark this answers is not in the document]
pstibrany: You're right, it's an arbitrary number, trying to express that "computing label values from series is much more expensive than doing it via postings intersection for label values". I'm not sure what would be the right value here (how much more) 

I guess we could try to compare both approaches for some random blocks and labels and compute cost of both, and figure out some value that way.

pstibrany:
Thanks for review @pracucci . I've updated the PR (and benchmarks) after addressing your comments.

tsdb/querier.go
pstibrany: Done in https://github.com/grafana/mimir-prometheus/pull/518/commits/d6daeb7cbc3695d1009b43da17e791d02bf0677b.

pstibrany: I've simplified the condition in https://github.com/grafana/mimir-prometheus/pull/518/commits/d6daeb7cbc3695d1009b43da17e791d02bf0677b.

tsdb/querier.go
```diff
@@ -376,6 +378,34 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
...
+		reachedEnd := true
+		if len(expanded) == maxExpandedPostings && p.Next() {
+			reachedEnd = false
+			expanded = append(expanded, p.At())
+		}
+
+		if reachedEnd {
+			// We have expanded all the postings -- all returned label values will be from these series only.
+			return labelValuesFromSeries(r, name, expanded)
+		}
```
colega, line 403: We should check `p.Err()` here.

colega, line 403: Also, can we just do:
```suggestion
		reachedEnd := true
		if len(expanded) < maxExpandedPostings { // Implies that `p.Next()` was false last time we called it.
			if p.Err() != nil {
				return nil, p.Err()
			}
			return labelValuesFromSeries(r, name, expanded)
		}
```

tsdb/querier.go
```diff
@@ -396,6 +426,78 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
...
+			return nil, errors.Wrapf(err, "label values for label %s", labelName)
+		}
+
+		v := builder.Labels().Get(labelName)
+		if v != "" {
+			values[v] = struct{}{}
+		}
+	}
+
+	result := make([]string, 0, len(values))
```
colega, line 446: Ignorable microoptimization: if we provide `allValues` to this method as `buf`, we can use that instead of allocating a new slice (because we don't need `allValues` anymore, and it's big enough).

tsdb/querier.go
pstibrany: We could, although that would allow for one less expanded postings than our computed `max`. I'll keep the current version, but add missing `p.Err()` (thanks for catching that!).

pstibrany: Fixed in https://github.com/grafana/mimir-prometheus/pull/518/commits/5679e446f14f46ecac123e609b5638289f7ac1b9.

colega: True, but we could fix that by changing both `for` and `if` to match less or equal instead of strictly less. I just find it strange and unnecessary to have a separate "add one more" case for the last posting (with an extra bool to indicate we didn't find it), instead of using the condition of the for loop that we've done already.

tsdb/querier.go
pstibrany: Ha, nice one. Why not.

colega (approved):
LGTM, thanks for addressing my comments.

pstibrany:
> LGTM, thanks for addressing my comments.

Thanks for your review and suggestions!

tsdb/querier.go
```diff
@@ -376,6 +378,34 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
...
+		// Call p.Next() even if len(expanded) == maxExpandedPostings. This tells us if there are more postings or not.
+		for len(expanded) <= maxExpandedPostings && p.Next() {
+			expanded = append(expanded, p.At())
+		}
+
+		if len(expanded) <= maxExpandedPostings {
+			// When we're here, p.Next() must have returned false, so we need to check for errors.
+			if err := p.Err(); err != nil {
+				return nil, errors.Wrap(err, "expanding postings for matchers")
+			}
```
pracucci, line 398: Why not moving it above the `if` block, so that it's always checked? If there was an error I think we should return in any case, otherwise we'll hit it later anyway.

tsdb/querier.go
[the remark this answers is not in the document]
pstibrany: I've seen your other comment. The link points to this decision function:

```
func (w labelValuesPostingsStrategy) preferSeriesToPostings(postings []storage.SeriesRef) bool {
	return int64(len(postings)*tsdb.EstimatedSeriesP99Size) < postingsListsTotalSize(w.allLabelValues)
}
```

I'm not sure how we could directly apply that (but you also hint as much in your comment), and I'm not quite sure what you mean by "replicating store-gateway logic". In any case, that can be a later follow up, once we know what to do there.

>  that was the result of quite a lot of testing

This was my main concern why I didn't do any such analysis here. I meant this PR as a quick fix for some edge cases, and even if it's not ideal, I have specific use-case in mind where it will help tremendously.

tsdb/querier.go
```diff
@@ -396,6 +426,83 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
 	return values, nil
 }
 
+// labelValuesFromSeries returns all unique label values from for given label name from supplied series. Values are not sorted.
+// buf is space for holding result (if it isn't big enough, it will be ignored), may be nil.
+func labelValuesFromSeries(r IndexReader, labelName string, refs []storage.SeriesRef, buf []string) ([]string, error) {
```
pracucci, line 431: labelValuesFromSeries() doesn’t return sorted labels. labelValuesWithMatchers() doesn’t as well, because index.FindIntersectingPostings() doesn’t guarantee sorted indexes in the output. Problem is that I can’t find you short these label values because Mimir expects them sorted.

_I'm still looking at it._

tsdb/querier.go
```diff
@@ -376,6 +378,34 @@ func labelValuesWithMatchers(r IndexReader, name string, matchers ...*labels.Mat
...
+		}
+
+		if len(expanded) <= maxExpandedPostings {
+			// When we're here, p.Next() must have returned false, so we need to check for errors.
+			if err := p.Err(); err != nil {
+				return nil, errors.Wrap(err, "expanding postings for matchers")
+			}
+
+			// We have expanded all the postings -- all returned label values will be from these series only.
+			// (We supply allValues[:0] as a buffer for storing results. It should be big enough already, since it holds all possible label values.)
```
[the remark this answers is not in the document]
pstibrany: Fixed :)

tsdb/querier.go
pstibrany: My impression was that we're supposed to only check for errors after `p.Next()` returns `false`. But I see that `Postings` interface actually doesn't say that, and it's only other `Iterator` interfaces saying that. I would still adhere to that pattern.

tsdb/querier.go
pstibrany: IndexReader.LabelValues explicitly says:

```
	// LabelValues returns possible label values which may not be sorted.
	LabelValues(name string, matchers ...*labels.Matcher) ([]string, error)
```

pstibrany: There is also `SortedLabelValues` method in that interface that client can call if it requires ordering.

pracucci: It's a bit hidden. When we call `LabelValues()` in the querier, under the hood it actually calls `SortedLabelValues()`:
https://github.com/grafana/mimir-prometheus/blob/2a075ced62391361192bd9d28a7c71d56c21eccc/tsdb/querier.go#L74-L77

Looks the `blockBaseQuerier` is both used when querying the TSDB head and compacted blocks.

tsdb/querier.go
pracucci: Ok I see your point and agree.

pracucci (approved):

samjewell:
Just wanted to say thank you @pstibrany (and also thanks to @pracucci and @colega). This looks perfect for my use-case ❤️ 
cc @gtk-grafana

PR #520, opened 2023-08-03
Update Prometheus to 2023-08-03

colega:
This brings latest Prometheus up to 2023-08-03's 61daa30bb which includes https://github.com/prometheus/prometheus/pull/12626 which is the commit  I need.


colega:
There are quite a lot of changes here, including many @krajorama's changes for native histograms.

I think the most sensitive change is the one about switching to a linked list for chunks  https://github.com/prometheus/prometheus/pull/11818

jesusvazquez (approved):
Approving. I did review this in two attempts (there is a lot of changes) and did not find something that troubles me other than all the changes related to the new linked list implementation for the heacd chunk of each memseries. Lets try this out in dev

pracucci (commented):
The change to mmap-ed chunks makes me quite nervous. @jesusvazquez mentioned it's just in `main` and not battle tested yet. I'm honestly not sure it's a good idea to move forward and include it in Mimir until it will be battle tested by Prometheus.

A part from this, I noticed a couple of things I suggest you to pay attention:
1. `jsoniter.RegisterTypeEncoderFunc("labels.Labels", unsafeMarshalLabelsJSON, labelsIsEmpty)` --> anything we can do in Mimir to use it? I just don't remember.
2. In `tsdb/db.go` (`run()`) there's new call to `db.head.mmapHeadChunks()`. We run with auto-compaction disabled and I'm wondering if that could impact us anyway.

jesusvazquez (commented):
I'll run another prometheus benchmark to the PR that has the troubling change to the head chunk for a couple of days to give us more assurances for the future.

Next prometheus release should happen mid august.

pracucci:
@colega I changed this PR to draft to avoid accidentally merging it.

PR #522, opened 2023-08-04
Pass ref to SeriesLifecycleCallback.PostDeletion (#12626)

colega:
When a particular SeriesLifecycleCallback tries to optimize and run closer to the Head, keeping track of the HeadSeriesRef instead of the labelsets, it's impossible to handle the PostDeletion callback properly as there's no way to know which series refs were deleted from the head.

This changes the callback to provide the series refs alongside the labelsets, so the implementation can choose what to do.


jesusvazquez (approved):
LGTM

PR #533, opened 2023-09-27
Add PostingsForMatchers cache size by bytes support

pracucci:
The `PostingsForMatchers` cache is per head / block, and the max size is expressed as "number of items". This makes me uncomfortable to enable because we have no control over the cache memory utilization and we can't configure a cap. In this PR I propose to do a first step, which introducing the support to limit the cache by bytes in addition to by number of items.

I kept the "max items" limit to introduce a breaking change at this point, but if the "max bytes" work well, we may consider to remove the "max items".

The idea is that this would be a first step. Next step would be make the cache global and not per head / block (which is more complicated), so that we can have better control over it. However, I believe that in some caches (a Mimir cluster with 1 tenant) just having the limit by bytes may be good enough, because it would be predictable (in 1 ingester with 1 tenant there's 1 TSDB Head and up to 6 blocks when using 13h retention).

### DIFF
--- a/tsdb/index/postings.go
+++ b/tsdb/index/postings.go
@@ -853,16 +853,6 @@ type PostingsCloner struct {
 // and it shouldn't be used once provided to the PostingsCloner.
 func NewPostingsCloner(p Postings) *PostingsCloner {
 	ids, err := ExpandPostings(p)
-
-	// The ExpandedPostings() doesn't know the total number of postings beforehand,
-	// so the returned slice capacity may be well above the actual number of items.
-	// In such case, we shrink it.
-	if float64(len(ids)) < float64(cap(ids))*0.70 {
-		shrunk := make([]storage.SeriesRef, len(ids))
-		copy(shrunk, ids)
-		ids = shrunk
-	}
-
 	return &PostingsCloner{ids: ids, err: err}
 }
 
--- a/tsdb/index/postings.go
+++ b/tsdb/index/postings.go
@@ -853,6 +853,16 @@ type PostingsCloner struct {
 // and it shouldn't be used once provided to the PostingsCloner.
 func NewPostingsCloner(p Postings) *PostingsCloner {
 	ids, err := ExpandPostings(p)
+
+	// The ExpandedPostings() doesn't know the total number of postings beforehand,
+	// so the returned slice capacity may be well above the actual number of items.
+	// In such case, we shrink it.
+	if float64(len(ids)) < float64(cap(ids))*0.70 {
+		shrinked := make([]storage.SeriesRef, len(ids))
+		copy(shrinked, ids)
+		ids = shrinked
+	}
+
 	return &PostingsCloner{ids: ids, err: err}
 }
 
--- a/tsdb/index/postings.go
+++ b/tsdb/index/postings.go
@@ -858,9 +858,9 @@ func NewPostingsCloner(p Postings) *PostingsCloner {
 	// so the returned slice capacity may be well above the actual number of items.
 	// In such case, we shrink it.
 	if float64(len(ids)) < float64(cap(ids))*0.70 {
-		shrinked := make([]storage.SeriesRef, len(ids))
-		copy(shrinked, ids)
-		ids = shrinked
+		shrunk := make([]storage.SeriesRef, len(ids))
+		copy(shrunk, ids)
+		ids = shrunk
 	}
 
 	return &PostingsCloner{ids: ids, err: err}
--- a/tsdb/index/postings_test.go
+++ b/tsdb/index/postings_test.go
@@ -1108,13 +1108,13 @@ func TestNewPostingsCloner_ShrinkExpandedPostingsSlice(t *testing.T) {
 	t.Run("should not shrink expanded postings if length is >= 70% capacity", func(t *testing.T) {
 		cloner := NewPostingsCloner(NewListPostings(make([]storage.SeriesRef, 60)))
 		assert.Equal(t, 60, len(cloner.ids))
-		assert.Equal(t, 64, cap(cloner.ids)) // Not shrinked.
+		assert.Equal(t, 64, cap(cloner.ids)) // Not shrunk.
 	})
 
 	t.Run("should shrink expanded postings if length is < 70% capacity", func(t *testing.T) {
 		cloner := NewPostingsCloner(NewListPostings(make([]storage.SeriesRef, 33)))
 		assert.Equal(t, 33, len(cloner.ids))
-		assert.Equal(t, 33, cap(cloner.ids)) // Shrinked.
+		assert.Equal(t, 33, cap(cloner.ids)) // Shrunk.
 	})
 }
 
--- a/tsdb/index/postings_test.go
+++ b/tsdb/index/postings_test.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/prometheus/prometheus/model/labels"
@@ -1100,6 +1101,20 @@ func TestPostingsCloner(t *testing.T) {
 		p := pc.Clone()
 		require.False(t, p.Next())
 		require.Equal(t, expectedErr, p.Err())
+	})
+}
+
+func TestNewPostingsCloner_ShrinkExpandedPostingsSlice(t *testing.T) {
+	t.Run("should not shrink expanded postings if length is >= 70% capacity", func(t *testing.T) {
+		cloner := NewPostingsCloner(NewListPostings(make([]storage.SeriesRef, 60)))
+		assert.Equal(t, 60, len(cloner.ids))
+		assert.Equal(t, 64, cap(cloner.ids)) // Not shrinked.
+	})
+
+	t.Run("should shrink expanded postings if length is < 70% capacity", func(t *testing.T) {
+		cloner := NewPostingsCloner(NewListPostings(make([]storage.SeriesRef, 33)))
+		assert.Equal(t, 33, len(cloner.ids))
+		assert.Equal(t, 33, cap(cloner.ids)) // Shrinked.
 	})
 }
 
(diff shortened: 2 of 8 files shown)


tsdb/index/postings_test.go
```diff
@@ -1103,6 +1104,20 @@ func TestPostingsCloner(t *testing.T) {
...
 }
 
+func TestNewPostingsCloner_ShrinkExpandedPostingsSlice(t *testing.T) {
+	t.Run("should not shrink expanded postings if length is >= 70% capacity", func(t *testing.T) {
+		cloner := NewPostingsCloner(NewListPostings(make([]storage.SeriesRef, 60)))
+		assert.Equal(t, 60, len(cloner.ids))
+		assert.Equal(t, 64, cap(cloner.ids)) // Not shrunk.
+	})
+
+	t.Run("should shrink expanded postings if length is < 70% capacity", func(t *testing.T) {
```
krajorama, line 1114: nit: doesn't test that the actual values are retained (i.e. copy was called)

tsdb/postings_for_matchers_cache.go
```diff
@@ -31,14 +37,15 @@ type IndexPostingsReader interface {
 // NewPostingsForMatchersCache creates a new PostingsForMatchersCache.
 // If `ttl` is 0, then it only deduplicates in-flight requests.
 // If `force` is true, then all requests go through cache, regardless of the `concurrent` param provided.
```
krajorama, line 39: nit: had to look up what's `concurrent`, maybe
```suggestion
// If `force` is true, then all requests go through cache, regardless of the `concurrent` param provided to the PostingsForMatchers method.
```

tsdb/postings_for_matchers_cache.go
```diff
@@ -101,13 +110,20 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Contex
 		cloner = index.NewPostingsCloner(postings)
 	}
 
-	c.created(key, c.timeNow())
+	// Estimate the size of the cache entry, in bytes. We use max() because
+	// size.Of() returns -1 if the value is nil.
+	estimatedSizeBytes := int64(len(key)) + max(0, int64(size.Of(outerErr))) + max(0, int64(size.Of(cloner)))
```
krajorama, line 115: Nit: a little bit weird to call this estimatedSizeBytes as we're using it as 
```
     // Size of he cached entry, in bytes.
     sizeBytes int64
```
later. Also at this point we have the data, so maybe call it `cachedSizeBytes` or just straight up `sizeBytes` cause the estimate part through me for a loop.

tsdb/index/postings.go
```diff
@@ -853,6 +853,16 @@ type PostingsCloner struct {
 // and it shouldn't be used once provided to the PostingsCloner.
 func NewPostingsCloner(p Postings) *PostingsCloner {
 	ids, err := ExpandPostings(p)
+
+	// The ExpandedPostings() doesn't know the total number of postings beforehand,
+	// so the returned slice capacity may be well above the actual number of items.
```
colega, line 858: How much above? 2x?

tsdb/postings_for_matchers_cache.go
```diff
@@ -101,13 +110,20 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Contex
...
+
+	c.created(key, c.timeNow(), estimatedSizeBytes)
 	return promise
 }
 
 type postingsForMatchersCachedCall struct {
 	key string
 	ts  time.Time
+
+	// Size of he cached entry, in bytes.
```
colega, line 125:
```suggestion
	// Size of the cached entry, in bytes.
```

tsdb/index/postings.go
krajorama: That's what I'm assuming based on how append realloc works...

krajorama (approved):
LGTM with some nitpicks. I was wondering if we're changing the WoW too much, but I guess we'll measure and for most regular users the 100*100K size will be plenty.

tsdb/index/postings.go
pracucci: Yes, 2x. Do you think we can ignore it and leave with the "up to 2x sized entries" in the cache?

tsdb/postings_for_matchers_cache.go
colega, line 115: I had to read the rest of the code to understand what is `outerErr` and why we're accounting for it.

I'd propose to extract all promise-related data to a struct, and change the anonymous function here to a receiver on it, like this: https://github.com/grafana/mimir-prometheus/commit/b7c4f994b2ff13cf556207ece169f5fe7434e6ad

Note that I also had to fix the test, because your implementation didn't account for the fact that we're also holding `sync.WaitGroup` in memory (and probably something else? because `estimatedSizeBytes` changed from 309 to 345.)

tsdb/index/postings.go
colega: I'd just ignore it here, and try to optimize `ExpandPostings` itself to improve sizing of the slice (it could check whether a given postings implements `LengthEstimate() int` method and then implement that method on the block index readers, etc.

tsdb/postings_for_matchers_cache.go
pracucci: I didn't count the wait group because it shouldn't be passed as a copy. All in all the difference is relatively small. I will consider your refactoring for a follow up PR, given doesn't look a blocking thing.

colega: The wait group escapes to the heap:
```
tsdb/postings_for_matchers_cache.go:86:3: moved to heap: wg
```

This isn't a big deal, I can send my commit as a separate PR after merging your if you want.

tsdb/index/postings.go
pracucci: Ok, let me revert this change then we can discuss how to improve it.

tsdb/index/postings_test.go
pracucci: I removed this change for now.

tsdb/postings_for_matchers_cache.go
pracucci: Agree. Renamed to `sizeBytes`. I was dubious too.

colega (approved):

pracucci:
Benchmark comparing before / after:

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
                                               │    sec/op    │    sec/op     vs base               │
PostingsForMatchersCache/no_evictions-12         453.9n ± ∞ ¹   461.1n ± ∞ ¹        ~ (p=0.690 n=5)
PostingsForMatchersCache/high_eviction_rate-12   2.727µ ± ∞ ¹   3.590µ ± ∞ ¹  +31.65% (p=0.008 n=5)
geomean                                          1.113µ         1.287µ        +15.64%
¹ need >= 6 samples for confidence interval at level 0.95

                                               │  before.txt   │               after.txt               │
                                               │     B/op      │     B/op       vs base                │
PostingsForMatchersCache/no_evictions-12           136.0 ± ∞ ¹     136.0 ± ∞ ¹       ~ (p=1.000 n=5) ²
PostingsForMatchersCache/high_eviction_rate-12   2.414Ki ± ∞ ¹   2.602Ki ± ∞ ¹  +7.77% (p=0.008 n=5)
geomean                                            579.8           601.9        +3.81%
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal

                                               │ before.txt  │              after.txt               │
                                               │  allocs/op  │  allocs/op   vs base                 │
PostingsForMatchersCache/no_evictions-12         7.000 ± ∞ ¹   7.000 ± ∞ ¹        ~ (p=1.000 n=5) ²
PostingsForMatchersCache/high_eviction_rate-12   22.00 ± ∞ ¹   25.00 ± ∞ ¹  +13.64% (p=0.008 n=5)
geomean                                          12.41         13.23         +6.60%
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal
```

pracucci:
The benchmark was not an apple-to-apple comparison with the `main` version because it was actually evicting at a slightly smaller rate in `main` (`maxBytes` was always triggering in the PR, while `maxItems` in `main` was triggering less times). After fixing the benchmark we can see no difference in memory allocations (as expected). According to profiling, the increase CPU is a mix of GC + `size.Of()`. However, in absolute terms, the CPU increase looks negligible to me:

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
                                               │    sec/op    │    sec/op     vs base               │
PostingsForMatchersCache/no_evictions-12         397.8n ± ∞ ¹   388.4n ± ∞ ¹   -2.36% (p=0.032 n=5)
PostingsForMatchersCache/high_eviction_rate-12   2.351µ ± ∞ ¹   2.955µ ± ∞ ¹  +25.69% (p=0.008 n=5)
geomean                                          967.1n         1.071µ        +10.78%
¹ need >= 6 samples for confidence interval at level 0.95

                                               │  before.txt   │               after.txt               │
                                               │     B/op      │     B/op       vs base                │
PostingsForMatchersCache/no_evictions-12           136.0 ± ∞ ¹     136.0 ± ∞ ¹       ~ (p=1.000 n=5) ²
PostingsForMatchersCache/high_eviction_rate-12   2.602Ki ± ∞ ¹   2.602Ki ± ∞ ¹       ~ (p=1.000 n=5) ²
geomean                                            601.9           601.9        +0.00%
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal

                                               │ before.txt  │              after.txt              │
                                               │  allocs/op  │  allocs/op   vs base                │
PostingsForMatchersCache/no_evictions-12         7.000 ± ∞ ¹   7.000 ± ∞ ¹       ~ (p=1.000 n=5) ²
PostingsForMatchersCache/high_eviction_rate-12   25.00 ± ∞ ¹   25.00 ± ∞ ¹       ~ (p=1.000 n=5) ²
geomean                                          13.23         13.23        +0.00%
¹ need >= 6 samples for confidence interval at level 0.95
² all samples are equal
```

tsdb/postings_for_matchers_cache.go
colega: Following up with the refactor: https://github.com/grafana/mimir-prometheus/pull/535

PR #534, opened 2023-09-27
`LabelValues()` with matchers should use cache

colega:
When cache was introduced, `LabelValues(matchers...)` could never be called with `concurrent=true` flag so it didn't make sense to use the cached call through the Head.

However, since the introduction of forced cache, we should use it, as even with `concurrent=false` the cache may be used.


pracucci (approved):
Good job, I agree!

PR #535, opened 2023-09-28
Refactor PostingsForMatcherCache promise

colega:
Extract promise payload as a struct, to make size calculation easier.


pracucci (approved):
Thanks!

P.S. I would have done it, but you've been faster than me ;)

PR #539, opened 2023-10-06
Sync with latest Prometheus upstream

aknuds1:
Sync with latest Prometheus upstream ([a5a4eab679ccf2d432ab34b5143ef9e1c6482839](https://github.com/prometheus/prometheus/tree/a5a4eab679ccf2d432ab34b5143ef9e1c6482839)).
 Significant changes include:

* golang.org/x/exp gets updated to revision 613f0c0eb8a1, which requires some rewriting of code (no behavioral changes that I'm aware of)

Note that I've had to [modify](https://github.com/grafana/mimir-prometheus/pull/539/files#diff-4f61bbb68838f886c4c6584c06bcc0760eb3176e748ae35988a92ee28d8a0c41L993) some `PostingsCloner` tests, since they assumed that it would be safe to call `At` on `ListPostings` before calling `Next/Seek` and after `Next/Seek` returns false. This is no longer the case in the latest revision of Prometheus. In my experience, standard iterator protocol is anyway that one should only call `At` after `Next` returns true.


krajorama:
fyi https://github.com/prometheus/prometheus/pull/12874 was already on main and thus it's not new - therefore not present in this diff

aknuds1:
@krajorama ah thanks, I will drop it from the PR description then.

tsdb/compact.go
```diff
@@ -232,8 +232,8 @@ func (c *LeveledCompactor) Plan(dir string) ([]string, error) {
 }
 
 func (c *LeveledCompactor) plan(dms []dirMeta) ([]string, error) {
-	slices.SortFunc(dms, func(a, b dirMeta) bool {
-		return a.meta.MinTime < b.meta.MinTime
+	slices.SortFunc(dms, func(a, b dirMeta) int {
+		return int(a.meta.MinTime - b.meta.MinTime)
```
krajorama, line 236: Hmm, on a 32bit architecture this will be theoretically wrong as it can lead to overflow. I'm not sure we need to worry about it (@bboreham ?) , but the following is correct regardless of bits
```suggestion
		switch {
		case a.meta.MinTime < b.meta.MinTime:
			return -1
		case a.meta.MinTime > b.meta.MinTime:
			return 1
		default:
			return 0
		}
```

aknuds1: @krajorama will you open a PR to change this upstream also @krajorama?

aknuds1: The fix makes sense to me FWIW.

krajorama: I'd like to get some feedback from @bboreham , because I might be too paranoid here :(

aknuds1: OK. I already committed the suggestion since I thought it makes perfect sense, and prevents you from having to worry about potential overflow.

tsdb/index/postings.go
```diff
@@ -755,39 +759,34 @@ func newListPostings(list ...storage.SeriesRef) *ListPostings {
 }
 
 func (it *ListPostings) At() storage.SeriesRef {
-	return it.cur
+	return it.list[it.pos-1]
```
krajorama, line 762: note: I've looked at the code using At() and most of it seems trivially correct , having Next/Seek above the function. There's the heap Less function though which is probably fine as far as I can tell. I guess we have to rely on that in Prometheus more eyes looked at this and that calling At without Next/Seek didn't make much sense before this either.

aknuds1: @krajorama yeah I have to say I'm not sure why we encoded in tests that it should be OK to call `At` before calling `Next`, or after `Next` returns false. It diverges from typical iterator protocol, and from the iterator implementations I've seen in Prometheus, it was also non-standard there.

krajorama: Sure, I think it was probably something that seemed like a good idea at the time. cc @simonswine

tsdb/compact.go
krajorama: My fix is probably slightly slower, although it shouldn't matter in compactor Plan phase, so it's fine.

tsdb/index/postings.go
colega: I did the `ListPostings` change in prometheus, and then I sent another PR to clarify the contract: https://github.com/prometheus/prometheus/pull/12921

It's incorrect to call `At` before `Next`, as it would always return the `0` seriesref.

tsdb/compact.go
bboreham: Change was from @gouthamve I think.
My calculations say we get overflow after 600 hours, which does seem a little close for comfort.

tsdb/index/postings.go
bboreham: Since you're mostly talking about `Next()`, note that the tests were specifically checking that `At()` returns zero value after a failed call to `Seek()`.

I don't think they should do this.

aknuds1: Most of the tests were checking `At` after a failed `Seek`, that's true, but it's wrong for the same reasons as for `Next`. A couple of tests were also checking `At` before and after (unsuccessfully) calling `Next` though, so it's a combination of the two.

tsdb/compact.go
aknuds1: Thanks for the input @bboreham - I guess @krajorama's fix is the way to go then?

krajorama (approved):
LGTM, we'll fix the 32bit issue upstream, doesn't affect us

colega:
@aknuds1 would it be possible to bring 4 more commits here? I merged https://github.com/prometheus/prometheus/pull/12955 which reverts https://github.com/prometheus/prometheus/pull/12906 which has a performance regression and we're bringing it here, so the changes we'd need to bring are these, which shouldn't affect this PR a lot:
```
commit fe90dcccffec8596ff3901b943aef5e1982172d6
Author: Oleg Zaytsev <mail@olegzaytsev.com>

    Revert ListPostings change (#12955)

commit 5bd8c8c561ecc8bf3dd37b80eb111eba64baa659
Author: Oleg Zaytsev <mail@olegzaytsev.com>

    Clarify Postings.At() contract (#12921)

commit b787e5f190428ba23de9b6dcaa18eb963f239a76
Merge: d3554d842 0cbf0c1c6
Author: Björn Rabenstein <beorn@grafana.com>

    Merge pull request #12931 from zenador/warning-non-monotonic-classic-histogram

commit d3554d84215e6279c4bf3d0bb2229206f012b689
Author: Danny Kopping <danny.kopping@grafana.com>

    Show group interval in Rules display (#12943)
```

aknuds1:
@colega sure.

aknuds1:
@krajorama I fixed the other 32-bit potential overflow comparison issues I could find. Also merged latest Prometheus main on request from @colega.

tsdb/index/postingsstats.go
```diff
@@ -63,8 +63,15 @@ func (m *maxHeap) push(item Stat) {
 }
 
 func (m *maxHeap) get() []Stat {
-	slices.SortFunc(m.Items, func(a, b Stat) bool {
-		return a.Count > b.Count
+	slices.SortFunc(m.Items, func(a, b Stat) int {
+		switch {
+		case b.Count < a.Count:
+			return -1
+		case b.Count > a.Count:
```
krajorama, line 70:
```suggestion
		case a.Count < b.Count:
			return -1
		case a.Count > b.Count:
```

cmd/promtool/tsdb.go
```diff
@@ -459,7 +459,16 @@ func analyzeBlock(ctx context.Context, path, blockID string, limit int, runExten
 	postingInfos := []postingInfo{}
 
 	printInfo := func(postingInfos []postingInfo) {
-		slices.SortFunc(postingInfos, func(a, b postingInfo) bool { return a.metric > b.metric })
+		slices.SortFunc(postingInfos, func(a, b postingInfo) int {
+			switch {
+			case b.metric < a.metric:
+				return -1
+			case b.metric > a.metric:
```
krajorama, line 466:
```suggestion
			case a.metric < b.metric:
				return -1
			case a.metric > b.metric:
```

tsdb/index/postingsstats.go
aknuds1: Doesn't this invert the original comparison @krajorama?

cmd/promtool/tsdb.go
aknuds1: Doesn't this invert the original comparison @krajorama?

krajorama: you're right

krajorama (approved):
LGTM

krajorama:
Opened https://github.com/prometheus/prometheus/pull/12978 upstream for the 32bit overflow issue.

PR #540, opened 2023-10-11
Fix issue where `chainSampleIterator` can obscure errors

charleskorn:
If any iterator in a `chainSampleIterator` has a sample after the desired timestamp when `Seek()` is called, `Seek()` will not return `ValNone`. 

If some iterators have failed (ie. calling `innerIterator.Err()` returns a non-`nil` value) and others have not, this behaviour may hide errors from those iterators: callers of `Seek()` will not receive `ValNone` and therefore won't know to check `chainSampleIterator.Err()`.

`Next()` is susceptible to a similar issue if `Next()` is not called until the stream is completely exhausted. 

I'll submit this fix upstream once we've tested Mimir with it.

### DIFF
--- a/storage/merge.go
+++ b/storage/merge.go
@@ -594,7 +594,17 @@ func (c *chainSampleIterator) Next() chunkenc.ValueType {
 
 	for {
 		currValueType = c.curr.Next()
-		if currValueType != chunkenc.ValNone {
+
+		if currValueType == chunkenc.ValNone {
+			if c.curr.Err() != nil {
+				// Abort if we've hit an error.
+				return chunkenc.ValNone
+			} else if len(c.h) == 0 {
+				// No iterator left to iterate.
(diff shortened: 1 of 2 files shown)


jhesketh:
> If some iterators have failed (ie. calling `innerIterator.Err()` returns a non-`nil` value) and others have not, this behaviour may hide errors from those iterators, as callers of `Seek()` will not receive `ValNone` and therefore check `chainSampleIterator.Err()`.

Do you mean "... and therefore not check `chainSampleIterator.Err()`?"

charleskorn:
English is ambiguous :) I've tried to make it clearer.

krajorama (approved):
LGTM

jhalterman (approved):

pracucci:
> I'll submit this fix upstream once we've tested Mimir with it.

👍

PR #541, opened 2023-10-12
Include trace ID in log lines emitted during rule evaluation

charleskorn:
This makes it easier to debug rule evaluation failures, and mirrors the pattern of behaviour for logs emitted by other parts of Mimir.

I'll submit this change upstream once we've tested Mimir with it.


jhesketh (approved):
lgtm, but seems like a harmless enough change that could go straight upstream?

PR #542, opened 2023-10-17
Add support for ingestion of out-of-order native histogram samples

carrieedwards:
This PR adds initial support for out-of-order ingestion of native histograms. It also includes unit tests for ingestion, compaction, and querying of OOO native histogram samples and chunks. 

Relevant issue: [#11220](https://github.com/prometheus/prometheus/issues/11220)


tsdb/head_read.go
```diff
@@ -772,3 +766,45 @@ func makeStopIterator(c chunkenc.Chunk, it chunkenc.Iterator, stopAfter int) chu
 		stopAfter: stopAfter,
 	}
 }
+
+func GetBytes(encoding chunkenc.Encoding, it chunkenc.Iterator) []byte {
```
fionaliao, line 770: Not used anymore, is removed in https://github.com/grafana/mimir-prometheus/pull/588

tsdb/head.go
```diff
@@ -457,10 +462,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			Name: "prometheus_tsdb_head_samples_appended_total",
 			Help: "Total number of appended samples.",
 		}, []string{"type"}),
-		outOfOrderSamplesAppended: prometheus.NewCounter(prometheus.CounterOpts{
+		outOfOrderSamplesAppended: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Name: "prometheus_tsdb_head_out_of_order_samples_appended_total",
 			Help: "Total number of appended out of order samples.",
-		}),
+		}, []string{"type"}),
 		outOfBoundSamples: prometheus.NewCounterVec(prometheus.CounterOpts{
```
jesusvazquez, line 469: If we make `outOfOrderSamplesAppended` a countervec i think `outOfBoundSamples` should be a countervec too.

tsdb/head_append.go
```diff
@@ -581,10 +629,16 @@ func (a *headAppender) AppendHistogram(ref storage.SeriesRef, lset labels.Labels
 		return 0, storage.ErrNativeHistogramsDisabled
 	}
 
-	if t < a.minValidTime {
+	// For OOO inserts, this restriction is irrelevant and will be checked later once we confirm the histogram sample is an in-order append.
+	// If OOO inserts are disabled, we may as well as check this as early as we can and avoid more work.
```
jesusvazquez, line 633: I like this comment but I'd rephrase it to say something along the lines of "fail fast if ooo is disabled"

tsdb/head_append.go
```diff
@@ -581,10 +629,16 @@ func (a *headAppender) AppendHistogram(ref storage.SeriesRef, lset labels.Labels
...
+	if a.oooTimeWindow == 0 && t < a.minValidTime {
 		a.head.metrics.outOfBoundSamples.WithLabelValues(sampleMetricTypeHistogram).Inc()
 		return 0, storage.ErrOutOfBounds
 	}
+	// If OOO is enabled, but OOO native histogram ingestion is disabled
```
jesusvazquez, line 638: Like above "also fail fast if ..."

tsdb/head_append.go
```diff
@@ -633,15 +687,25 @@ func (a *headAppender) AppendHistogram(ref storage.SeriesRef, lset labels.Labels
 	switch {
 	case h != nil:
 		s.Lock()
-		if err := s.appendableHistogram(t, h); err != nil {
-			s.Unlock()
-			if errors.Is(err, storage.ErrOutOfOrderSample) {
+		// TODO(codesome): If we definitely know at this point that the sample is ooo, then optimise
```
jesusvazquez, line 690: Does this todo still apply here?

tsdb/head_append.go
```diff
@@ -1463,33 +1713,41 @@ func (s *memSeries) cutNewHeadChunk(mint int64, e chunkenc.Encoding, chunkRange
...
-	})
+		// OOO is not enabled or there is no head chunk, so nothing to m-map here.
+		return nil
+	}
+	chks, err := s.ooo.oooHeadChunk.chunk.ToEncodedChunks(math.MinInt64, math.MaxInt64)
+	if err != nil {
+		handleChunkWriteError(err)
+		return nil
+	}
+	chunkRefs := make([]chunks.ChunkDiskMapperRef, 0, 1)
```
jesusvazquez, line 1738: Could the list of chunk refs also be len(chks)

jesusvazquez (commented):
I've done a first pass to the ingest part. Tomorrow I'll do first pass to reads. 

Overall looks great nice work @fionaliao and @carrieedwards 

There is some code duplication, specially in the Commit() bits but I think its an issue with how the code was before, its hard to abstract it so we have less duplication at this point. It would be a big effort so we have to live with it.

tsdb/head.go
fionaliao: `outOfBoundSamples` is already a countervec. Previously it only tracked oob floats though, this PR starts tracking oob histograms as well

tsdb/head_append.go
fionaliao: Yes - it's not been optimised yet. Currently, OOO samples are always appended to the WAL.

tsdb/head_append.go
fionaliao: Updated comment in https://github.com/grafana/mimir-prometheus/pull/542/commits/ccbf860c6a9a78e4389e3ee4aea7ba8f8ff6f093

tsdb/head_append.go
fionaliao: Yep, updated in https://github.com/grafana/mimir-prometheus/pull/542/commits/ccbf860c6a9a78e4389e3ee4aea7ba8f8ff6f093

util/teststorage/storage.go
```diff
@@ -50,6 +50,25 @@ func New(t testutil.T) *TestStorage {
 	return &TestStorage{DB: db, exemplarStorage: es, dir: dir}
 }
 
+func NewTestStorageWithOpts(t testutil.T, opts *tsdb.Options) *TestStorage {
```
krajorama, line 53: Seems like the function `New` above could just use this if we didn't set OutOfOrderCapMax here?

PR #544, opened 2023-10-18
Add counter reset test for mmapped OOO chunks

fionaliao:
Fixed a couple of bugs that were discovered while writing the test:
- mint for ooo chunk needed to be set when a new chunk was created
- first chunk was using the wrong previous appender


carrieedwards (approved):
Looks good to me

PR #546, opened 2023-10-20
Fix issue where `PostingsForMatchersCache` can cache a context cancellation error

charleskorn:
If the initial request for a set of matchers to `PostingsForMatchersCache.PostingsForMatchers()` cancels its context before the call returns, this cancellation can be cached, causing all subsequent calls to the cache for the same matchers to also return a context cancellation error while the cache entry is still live.

While this PR is sufficient to fix the problem above, there are some further improvements we could consider in the future:

* Should other kinds of errors also not be cached?
* If all waiting calls to `PostingsForMatchersCache.PostingsForMatchers()` have cancelled their context, should we cancel the context passed to the underlying `PostingsForMatchers` implementation?

### DIFF
--- a/tsdb/postings_for_matchers_cache.go
+++ b/tsdb/postings_for_matchers_cache.go
@@ -93,6 +93,11 @@ func (p *postingsForMatcherPromise) result(ctx context.Context) (index.Postings,
 	case <-ctx.Done():
 		return nil, ctx.Err()
 	case <-p.done:
+		// Checking context error is necessary for deterministic tests,
+		// as channel selection order is random
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
 		if p.err != nil {
 			return nil, p.err
 		}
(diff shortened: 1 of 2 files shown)


aknuds1 (approved):
LGTM; I had to fix `postingsForMatcherPromise.result`, so it's deterministic and not randomly fail tests.

tsdb/postings_for_matchers_cache.go
```diff
@@ -107,12 +112,18 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ix IndexPostingsRe
 	key := matchersKey(ms)
 	oldPromise, loaded := c.calls.LoadOrStore(key, promise)
 	if loaded {
-		promise = oldPromise.(*postingsForMatcherPromise)
-		return promise.result
+		// promise was not stored, we return a previously stored promise, that's possibly being fulfilled in another goroutine
+		close(promise.done)
```
charleskorn, line 116: @aknuds1 why is closing the channel here necessary?

aknuds1: @charleskorn I guess it was old habit, thinking that the unused channel should be closed, but thinking it through again, it's actually not necessary. It should make no difference for garbage collection, AFAIK.

PR #550, opened 2023-10-24
Hide position info for warnings

zenador:
This is needed in Mimir because sharded queries (hidden from user) are different from raw queries (not what is actually used), so we can't use either as reference for line positions.

Needed for https://github.com/grafana/mimir/pull/6391

### DIFF
--- a/util/annotations/annotations.go
+++ b/util/annotations/annotations.go
@@ -116,6 +116,9 @@ type annoErr struct {
 }
 
 func (e annoErr) Error() string {
+	if e.Query == "" {
+		return fmt.Sprintf("%s", e.Err)
+	}
 	return fmt.Sprintf("%s (%s)", e.Err, e.PositionRange.StartPosInput(e.Query, 0))
 }
 


util/annotations/annotations.go
```diff
@@ -116,6 +116,9 @@ type annoErr struct {
 }
 
 func (e annoErr) Error() string {
+	if e.Query == "" {
+		return fmt.Sprintf("%s", e.Err)
```
charleskorn, line 120: Can we just do this?

```suggestion
		return e.Err.Error()
```

charleskorn (commented):
Is this a change you think would be accepted by upstream Prometheus? 

If not, I'm wondering if we need to take a different approach to minimise the number of differences in this fork.

util/annotations/annotations.go
zenador: Good point, thanks! Moving this to https://github.com/prometheus/prometheus/pull/13031

Issue #548, opened 2023-10-20
Follow up on known `PostingsForMatchersCache` issues


aknuds1:
### Proposal

As mentioned by @charleskorn in #546, there are a couple of outstanding issues in `PostingsForMatchersCache` that should be followed up on:

* Should other kinds of errors also not be cached?
* If all waiting calls to PostingsForMatchersCache.PostingsForMatchers() have cancelled their context, should we cancel the context passed to the underlying PostingsForMatchers implementation?

charleskorn:
One other thing that needs to be done once those questions are answered: we should upstream the fix in #546 (and any other related changes) to Prometheus.

charleskorn:
@aknuds1 I see you've assigned yourself to this issue, but let me know if you need any help.

aknuds1:
> we should upstream the fix in https://github.com/grafana/mimir-prometheus/pull/546 (and any other related changes) to Prometheus

@charleskorn which changes from #546, considering `PostingsForMatchersCache` isn't upstream? Sorry if I'm missing the obvious.

aknuds1:
> I see you've assigned yourself to this issue, but let me know if you need any help.

@charleskorn I unassigned myself, since I figured it's better to leave it free in the backlog, so whoever has the capacity to pick it up can.

charleskorn:
> > we should upstream the fix in https://github.com/grafana/mimir-prometheus/pull/546 (and any other related changes) to Prometheus
> 
> @charleskorn which changes from #546, considering `PostingsForMatchersCache` isn't upstream? Sorry if I'm missing the obvious.

Oh, I misunderstood, I thought this was an upstream feature. In that case, no upstreaming required 🙂

aknuds1:
Haven spoken offline with @charleskorn, it might be preferable if I pick this issue up if I have the capacity, given that I possess the most context wrt. making Prometheus support query cancellation.

PR #551, opened 2023-10-24
Make `PostingsForMatchersCache.PostingsForMatchers` respect context cancellation

aknuds1:
Make `PostingsForMatchersCache.PostingsForMatchers` respect context cancellation when executing queries (to populate cache), by executing them in a background goroutine taking a dedicated context, which gets canceled once no `PostingsForMatchers` caller needs it any more.

Fixes https://github.com/grafana/mimir-prometheus/issues/548.

TODO:

- [ ] Add test exercising concurrent access, where most accesses should use the cache

### DIFF
--- a/tsdb/postings_for_matchers_cache.go
+++ b/tsdb/postings_for_matchers_cache.go
@@ -3,6 +3,7 @@ package tsdb
 import (
 	"container/list"
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 	"time"
@@ -114,7 +115,7 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
 		return c.waitOnPromise(ctx, promise, key)
 	}
 
-	queryCtx, cancel := context.WithCancel(context.Background())
+	queryCtx, cancel := context.WithCancelCause(context.Background())
 	promise = &postingsForMatchersPromise{
 		done:    make(chan struct{}),
 		waiting: atomic.NewInt32(1),
@@ -155,7 +156,7 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
 type postingsForMatchersPromise struct {
 	done    chan struct{}
 	waiting *atomic.Int32
-	cancel  context.CancelFunc
+	cancel  context.CancelCauseFunc
 
 	cloner *index.PostingsCloner
 	err    error
@@ -173,11 +174,11 @@ func (c *PostingsForMatchersCache) waitOnPromise(ctx context.Context, promise *p
 		if waiting > 0 {
 			// Promise is in use by other goroutines
 			c.cachedMtx.Unlock()
-			return nil, ctx.Err()
+			return nil, errors.Wrap(ctx.Err(), "PostingsForMatchers context error")
 		}
 
 		// There are no more waiting goroutines, cancel the promise and remove it from the cache
-		promise.cancel()
+		promise.cancel(fmt.Errorf("no remaining callers interested in query"))
 		delete(c.calls, key)
 		for elem := c.cached.Front(); elem != nil; elem = elem.Next() {
 			cc := elem.Value.(*postingsForMatchersCachedCall)
@@ -191,15 +192,12 @@ func (c *PostingsForMatchersCache) waitOnPromise(ctx context.Context, promise *p
 
 		// Wait for query execution goroutine to finish
 		<-promise.done
-		return nil, ctx.Err()
+		return nil, errors.Wrap(ctx.Err(), "PostingsForMatchers context error")
 	case <-promise.done:
 		// Checking context error is necessary for deterministic tests,
 		// as channel selection order is random
 		if ctx.Err() != nil {
-			span.AddEvent("completed postingsForMatchers promise, but context has error", trace.WithAttributes(
-				attribute.String("err", ctx.Err().Error()),
-			))
-			return nil, ctx.Err()
+			return nil, errors.Wrap(ctx.Err(), "PostingsForMatchers context error")
 		}
 
 		if promise.err != nil {
--- a/tsdb/postings_for_matchers_cache.go
+++ b/tsdb/postings_for_matchers_cache.go
@@ -117,7 +117,7 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
 
 	queryCtx, cancel := context.WithCancelCause(context.Background())
 	promise = &postingsForMatchersPromise{
-		done:    make(chan struct{}),
+		done:    make(chan int64),
 		waiting: atomic.NewInt32(1),
 		cancel:  cancel,
 	}
@@ -147,14 +147,15 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
 		}
 
 		sizeBytes := int64(len(key) + size.Of(promise))
-		c.created(key, c.timeNow(), sizeBytes)
+		promise.done <- sizeBytes
 	}()
 
 	return c.waitOnPromise(ctx, promise, key)
 }
 
 type postingsForMatchersPromise struct {
-	done    chan struct{}
+	// done signals when the promise is fulfilled, sending the cache entry size for storing by a consumer
+	done    chan int64
 	waiting *atomic.Int32
 	cancel  context.CancelCauseFunc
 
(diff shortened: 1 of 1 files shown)


tsdb/postings_for_matchers_cache.go
```diff
@@ -78,62 +80,121 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
...
+			if queryCtx.Err() != nil {
+				// Canceled queries will be pruned from the cache
+				return
+			}
+
+			promise.cloner = index.NewPostingsCloner(postings)
+		}
+
+		sizeBytes := int64(len(key) + size.Of(promise))
+		c.created(key, c.timeNow(), sizeBytes)
```
dimitarvdimitrov, line 139: isn't there a race between calling `created` here can removing the item from the cache in `waitOnPromise`? I think holding `cachedMtx` while checking `queryCtx.Err()` is one way to avoid it

tsdb/postings_for_matchers_cache.go
```diff
@@ -78,62 +80,121 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
...
+			if cc.key == key {
+				c.cached.Remove(elem)
+				c.cachedBytes -= cc.sizeBytes
+				break
+			}
+		}
+		c.cachedMtx.Unlock()
+
+		// Wait for query execution goroutine to finish
+		<-promise.done
```
dimitarvdimitrov, line 183: what's the reason for waiting here?

tsdb/postings_for_matchers_cache.go
```diff
@@ -78,62 +80,121 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
...
 		return nil, ctx.Err()
-	case <-p.done:
+	case <-promise.done:
 		// Checking context error is necessary for deterministic tests,
 		// as channel selection order is random
 		if ctx.Err() != nil {
 			return nil, ctx.Err()
```
dimitarvdimitrov, line 189: nitpick but let's warp the error with something. Chasing plain `context canceled` isn't fun.

dimitarvdimitrov: same comment applies above

tsdb/postings_for_matchers_cache.go
```diff
@@ -78,62 +80,121 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
...
+		promise.cancel()
+		delete(c.calls, key)
+		for elem := c.cached.Front(); elem != nil; elem = elem.Next() {
+			cc := elem.Value.(*postingsForMatchersCachedCall)
+			if cc.key == key {
+				c.cached.Remove(elem)
+				c.cachedBytes -= cc.sizeBytes
+				break
+			}
+		}
```
dimitarvdimitrov, line 179: I'm afraid this can be slow. Is there a way to make sure we don't insert it into the linked list if all the contexts are cancelled?

perhaps have a channel to which the producer writes an empty struct when it's done with producing the result. Any consumer will take this empty struct and commit to consuming the result. If there is no consumer waiting on the channel, then the producer just discards the result.

dimitarvdimitrov (commented):
thanks for the work. I'm not super sure about the linked list traversal. I think if we change the linked list traversal, then we'll also take care of the race with the producer.

tsdb/postings_for_matchers_cache.go
aknuds1: Maybe it's not necessary any longer, have to review it.

tsdb/postings_for_matchers_cache.go
```diff
@@ -78,62 +80,121 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
...
+		c.cachedMtx.RUnlock()
+		return c.waitOnPromise(ctx, promise, key)
+	}
+
+	c.cachedMtx.RUnlock()
+
+	c.cachedMtx.Lock()
+	if c.calls[key] != nil {
+		// A promise has been injected into the cache in the meantime
+		promise = c.calls[key]
```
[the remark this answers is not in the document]
aknuds1: Yep, you're right :facepalm:

tsdb/postings_for_matchers_cache.go
aknuds1: Yeah, it shouldn't be necessary any longer (it was in a previous revision).

tsdb/postings_for_matchers_cache.go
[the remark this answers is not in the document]
aknuds1: Done.

tsdb/postings_for_matchers_cache.go
aknuds1: Actually, after removing this line I get a data race in tests. Not sure why yet.

tsdb/postings_for_matchers_cache.go
aknuds1: Done, PTAL. I also added a cause to the cancellation, to ease diagnosis.

charleskorn (commented):
This is quite a complex change - how necessary is it that we cancel postings for matchers calls if the caller cancels? (How expensive are these calls?) And how often does the caller cancel? 

I'm wondering if we can just accept that occasionally we'll waste some cycles completing a postings for matchers call for a caller that has already cancelled.

tsdb/postings_for_matchers_cache.go
aknuds1: Actually more or less the same idea came to when reviewing the code this morning.

aknuds1: I tried implementing this idea @dimitarvdimitrov, by sending the cache entry size over `promise.done` for a consumer to pick up and commit, could you have a look?

aknuds1:
> I'm wondering if we can just accept that occasionally we'll waste some cycles completing a postings for matchers call for a caller that has already cancelled.

@charleskorn as discussed offline, I see your concern, and I think we'll need to analyze the necessity of being able to cancel executing queries in the light of past issues at Grafana Labs.

tsdb/postings_for_matchers_cache.go
dimitarvdimitrov: I see. Then we're holding the whole request until the PostingsForMatchers finishes. Can we have a `select` with `default` or with `case <-ctx.Done()` when sending to `promise.done`? Then we can immediately return the cancelled request and let the background goroutine clean up when it's ready

tsdb/postings_for_matchers_cache.go
aknuds1: Stale.

tsdb/postings_for_matchers_cache.go
```diff
@@ -78,62 +81,120 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
...
+	queryCtx, cancel := context.WithCancelCause(context.Background())
+	promise = &postingsForMatchersPromise{
+		done:    make(chan int64),
+		waiting: atomic.NewInt32(1),
+		cancel:  cancel,
+	}
+	c.calls[key] = promise
+	c.cachedMtx.Unlock()
+
+	go func() {
```
charleskorn, line 118: Should we move this to its own method on `postingsForMatchersPromise` perhaps? This method is starting to get quite big.

tsdb/postings_for_matchers_cache.go
```diff
@@ -78,62 +81,120 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
...
+		waiting := promise.waiting.Dec()
+		if waiting > 0 {
+			// Promise is in use by other goroutines
+			c.cachedMtx.Unlock()
+			return nil, errors.Wrap(ctx.Err(), "PostingsForMatchers context error")
+		}
+
+		// There are no more waiting goroutines, cancel the promise and remove it from the cache
+		promise.cancel(fmt.Errorf("no remaining callers interested in query"))
+		delete(c.calls, key)
```
charleskorn, line 174: I don't think this is quite right - calling `delete` here will remove the promise from the cache if a subsequent call's context is cancelled, even if the promise succeeded. Imagine this sequence of events:

1. Initial call for a set of matchers arrives, promise succeeds and value is cached and returned.
2. Subsequent call arrives with a context that is already cancelled. Cached promise is retrieved from cache, `waiting` is incremented to 1 and then `waitOnPromise` is called.
3. `waitOnPromise` observes cancelled context, decrements `waiting` to 0 and removes cached promise from cache.

Instead, we only want to cancel and remove the call from `c.calls` if there's nothing waiting and the call is still in progress.

tsdb/postings_for_matchers_cache.go
charleskorn: I think we can solve the blocking on `promise.done` by moving the call to `c.created` to the goroutine, and `promise.done` is used only to indicate completion (see comment below) - then `promise.done` is only ever closed, and never sent to.

charleskorn (commented):
I know this is a WIP, but I'd suggest adding some unit tests for the different cases as a next step, to help flush out some of the subtle bugs that might be introduced before we refactor and address other issues we've noticed.

pracucci:
Superseded by #644

PR #553, opened 2023-10-25
PostingsForMatchersCache: Add tracing

aknuds1:
I propose to add (OTel, as in the rest of Prometheus) tracing spans to `tsdb.PostingsForMatchersCache`, to make it easier to diagnose eventual problems when using its `PostingsForMatchers` method. Since `PostingsForMatchers` might cache calls, insight into its operations can be really useful.

TODO:
- [ ] Manual testing to see spans in practice


tsdb/postings_for_matchers_cache.go
```diff
@@ -74,11 +77,27 @@ type PostingsForMatchersCache struct {
 }
 
 func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix IndexPostingsReader, concurrent bool, ms ...*labels.Matcher) (index.Postings, error) {
+	var matcherStrs []string
+	for _, m := range ms {
+		matcherStrs = append(matcherStrs, m.String())
+	}
```
charleskorn, line 83: [nit] Is there an existing method that will format a list of matchers? If not, might be nice to move this functionality into a helper method to keep this method focused

tsdb/postings_for_matchers_cache.go
```diff
@@ -74,11 +77,27 @@ type PostingsForMatchersCache struct {
...
+	ctx, span := otel.Tracer("").Start(ctx, "PostingsForMatchersCache.PostingsForMatchers", trace.WithAttributes(
+		attribute.StringSlice("matchers", matcherStrs),
+		attribute.Bool("concurrent", concurrent),
+		attribute.Bool("force", c.force),
+	))
+	defer span.End()
+
 	if !concurrent && !c.force {
-		return c.postingsForMatchers(ctx, ix, ms...)
+		span.AddEvent("cache not in use")
```
charleskorn, line 92: [nit]
```suggestion
		span.AddEvent("cache not used")
```

tsdb/postings_for_matchers_cache.go
```diff
@@ -74,11 +77,27 @@ type PostingsForMatchersCache struct {
...
+		p, err := c.postingsForMatchers(ctx, ix, ms...)
+		span.RecordError(err)
```
charleskorn, line 94: Do we also need to call `SetStatus(codes.Error, ...)` here if `err` is not `nil`? 

(same comment applies below - might be handy to create / use a helper method for this)

tsdb/postings_for_matchers_cache.go
```diff
@@ -89,34 +108,57 @@ type postingsForMatcherPromise struct {
...
 		// Checking context error is necessary for deterministic tests,
 		// as channel selection order is random
 		if ctx.Err() != nil {
+			span.AddEvent("successful postingsForMatchers promise, but context has error", trace.WithAttributes(
+				attribute.String("err", ctx.Err().Error()),
+			))
 			return nil, ctx.Err()
 		}
 		if p.err != nil {
+			span.AddEvent("failed postingsForMatchers promise ", trace.WithAttributes(
```
charleskorn, line 130:
```suggestion
			span.AddEvent("failed postingsForMatchers promise", trace.WithAttributes(
```

tsdb/postings_for_matchers_cache.go
```diff
@@ -89,34 +108,57 @@ type postingsForMatcherPromise struct {
...
+
 	promise := new(postingsForMatcherPromise)
 	promise.done = make(chan struct{})
 
 	key := matchersKey(ms)
 	oldPromise, loaded := c.calls.LoadOrStore(key, promise)
 	if loaded {
 		// promise was not stored, we return a previously stored promise, that's possibly being fulfilled in another goroutine
+		span.AddEvent("using cached postingsForMatchers promise", trace.WithAttributes(
+			attribute.String("key", key),
```
charleskorn, line 152: Is the key relevant here? It should be the same for all calls with the same set of matchers, which we logged earlier (in `PostingsForMatchers()`)

(same comment applies below)

tsdb/postings_for_matchers_cache.go
```diff
@@ -89,34 +108,57 @@ type postingsForMatcherPromise struct {
...
+		))
 		close(promise.done)
 		return oldPromise.(*postingsForMatcherPromise).result
 	}
 
+	span.AddEvent("cached new postingsForMatchers promise, executing query", trace.WithAttributes(
```
charleskorn, line 158: I'd suggest rephrasing this to make it a little clearer:

```suggestion
	span.AddEvent("no postingsForMatchers promise in cache, executing query", trace.WithAttributes(
```

tsdb/postings_for_matchers_cache.go
```diff
@@ -144,13 +190,20 @@ type postingsForMatchersCachedCall struct {
 	sizeBytes int64
 }
 
-func (c *PostingsForMatchersCache) expire() {
+func (c *PostingsForMatchersCache) expire(ctx context.Context) {
```
charleskorn, line 193: I'm not sure the tracing in this method is necessary - in the context of a query request, knowing that cache entries for other requests have expired is not so interesting.

tsdb/postings_for_matchers_cache.go
```diff
@@ -191,8 +245,15 @@ func (c *PostingsForMatchersCache) evictHead() {
 
 // created has to be called when returning from the PostingsForMatchers call that creates the promise.
 // the ts provided should be the call time.
-func (c *PostingsForMatchersCache) created(key string, ts time.Time, sizeBytes int64) {
+func (c *PostingsForMatchersCache) created(ctx context.Context, key string, ts time.Time, sizeBytes int64) {
+	_, span := otel.Tracer("").Start(ctx, "PostingsForMatchersCache.created")
```
charleskorn, line 249: Rather than creating a separate span for `created`, what if we instead just attach these events to the caller's span (ie. `postingsForMatchersPromise()`'s span)?

tsdb/postings_for_matchers_cache.go
```diff
@@ -206,6 +267,13 @@ func (c *PostingsForMatchersCache) created(key string, ts time.Time, sizeBytes i
 		sizeBytes: sizeBytes,
 	})
 	c.cachedBytes += sizeBytes
+	span.AddEvent("recorded cached promise size", trace.WithAttributes(
```
charleskorn, line 270: [nit] This seems like a more accurate description (to me at least)

```suggestion
	span.AddEvent("added cached value to expiry queue", trace.WithAttributes(
```

tsdb/postings_for_matchers_cache.go
aknuds1: I made this into a span attribute `cache_key` in `postingsForMatchersPromise` instead, as it already computes the key from the matchers.

tsdb/postings_for_matchers_cache.go
aknuds1: Removed.

tsdb/postings_for_matchers_cache.go
aknuds1: Removed.

tsdb/postings_for_matchers_cache.go
aknuds1: Done.

tsdb/postings_for_matchers_cache.go
```diff
@@ -74,11 +77,24 @@ type PostingsForMatchersCache struct {
...
+		p, err := c.postingsForMatchers(ctx, ix, ms...)
+		span.RecordError(err)
+		return p, err
 	}
+
+	span.AddEvent("using cache")
 	c.expire()
-	return c.postingsForMatchersPromise(ix, ms)(ctx)
+	p, err := c.postingsForMatchersPromise(ctx, ix, ms)(ctx)
+	span.RecordError(err)
```
ying-jeanne, line 96: RecordError is adding error event into span, but it would not set span status to error. we need to call span.SetStatus() together doc: https://opentelemetry.io/docs/instrumentation/go/manual/

charleskorn (approved):

tsdb/postings_for_matchers_cache.go
aknuds1: Done, thanks.

tsdb/postings_for_matchers_cache.go
aknuds1: Done.

tsdb/postings_for_matchers_cache.go
```diff
@@ -89,34 +122,55 @@ type postingsForMatcherPromise struct {
...
 		// Checking context error is necessary for deterministic tests,
 		// as channel selection order is random
 		if ctx.Err() != nil {
+			span.AddEvent("completed postingsForMatchers promise, but context has error", trace.WithAttributes(
+				attribute.String("err", ctx.Err().Error()),
+			))
 			return nil, ctx.Err()
 		}
 		if p.err != nil {
+			span.AddEvent("failed postingsForMatchers promise", trace.WithAttributes(
```
charleskorn, line 143: [nit]

```suggestion
			span.AddEvent("postingsForMatchers promise completed with error", trace.WithAttributes(
```

tsdb/postings_for_matchers_cache.go
```diff
@@ -89,34 +122,55 @@ type postingsForMatcherPromise struct {
...
+				attribute.String("err", p.err.Error()),
+			))
 			return nil, p.err
 		}
+		span.AddEvent("successful postingsForMatchers promise")
```
charleskorn, line 148: [nit]

```suggestion
		span.AddEvent("postingsForMatchers promise completed successfully")
```

tsdb/postings_for_matchers_cache.go
```diff
@@ -89,34 +122,55 @@ type postingsForMatcherPromise struct {
...
 		return p.cloner.Clone(), nil
 	}
 }
 
-func (c *PostingsForMatchersCache) postingsForMatchersPromise(ix IndexPostingsReader, ms []*labels.Matcher) func(context.Context) (index.Postings, error) {
-	promise := new(postingsForMatcherPromise)
-	promise.done = make(chan struct{})
+func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Context, ix IndexPostingsReader, ms []*labels.Matcher) func(context.Context) (index.Postings, error) {
+	span := trace.SpanFromContext(ctx)
+	defer span.End()
```
charleskorn, line 155: Should this method be closing a span it didn't start? If so, might be good to add a comment explaining why.

aknuds1: @charleskorn oops, that's a bug. Thanks!

aknuds1: Fixed.

dimitarvdimitrov:
can we have these log events in the span instead of full spans? The result now is that we get one span per regular expression per ingester making the whole trace even less navigable than it used to be. 
Before
Screenshot 2023-12-12 at 12 54 42
After

Screenshot 2023-12-12 at 12 54 32

aknuds1:
@dimitarvdimitrov do you mean to not start a span in `PostingsForMatchersCache.PostingsForMatchers`, instead using the context's span if present?

dimitarvdimitrov:
> @dimitarvdimitrov do you mean to not start a span in `PostingsForMatchersCache.PostingsForMatchers`, instead using the context's span if present?

yes, exactly

aknuds1:
Given the pattern seen here, I think that'd make sense. Do you want to open a PR?

dimitarvdimitrov:
> Given the pattern seen here, I think that'd make sense. Do you want to open a PR?

sure, i opened https://github.com/grafana/mimir-prometheus/pull/575

PR #555, opened 2023-10-26
Add engine tests for OOO native histograms with counter resets

carrieedwards:
This PR adds tests in engine_test.go to test counter resets when using the rate() function on native histograms.

### DIFF
--- a/promql/engine_test.go
+++ b/promql/engine_test.go
@@ -3258,6 +3258,7 @@ func TestNativeFloatHistogramRate(t *testing.T) {
 	require.Equal(t, expectedHistogram, actualHistogram)
 }
 
+// Tests the rate() function for native histograms with both in-order and OOO sample ingestion and counter resets
 func TestNativeHistogramRateWithCounterResets(t *testing.T) {
 	scenarios := map[string]struct {
 		appendFunc func(appender storage.Appender, lbls labels.Labels, ts, value int64) (storage.SeriesRef, error)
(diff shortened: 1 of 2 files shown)


promql/engine_test.go
```diff
@@ -3256,6 +3259,914 @@ func TestNativeFloatHistogramRate(t *testing.T) {
...
+
+	type sampleBatch struct {
+		from   int64
+		until  int64
+		filter filterFunc
+	}
+
+	tests := []struct {
+		name              string
+		appendFunc        func(appender storage.Appender, lbls labels.Labels, ts, value int64) error
```
fionaliao, line 3309: Nit: this appendFunc isn't used

promql/engine_test.go
```diff
@@ -3256,6 +3259,914 @@ func TestNativeFloatHistogramRate(t *testing.T) {
 	require.Equal(t, expectedHistogram, actualHistogram)
 }
 
+// Tests the rate() function for native histograms with both in-order and OOO sample ingestion and counter resets
```
fionaliao, line 3262: Opinion: I don't think this comment is needed, I think the test name + a brief glance of the test cases is explanatory enough by themselves

promql/engine_test.go
```diff
@@ -3256,6 +3259,914 @@ func TestNativeFloatHistogramRate(t *testing.T) {
...
+	seriesName := "sparse_histogram_series"
+	lbls := labels.FromStrings("__name__", seriesName)
+
+	type filterFunc func(v int64) bool
+	defaultFilterFunc := func(v int64) bool { return false }
+
+	type sampleBatch struct {
+		from   int64
+		until  int64
+		filter filterFunc
```
fionaliao, line 3304: I think we should rename `filter` to `shouldReset` - I think `shouldReset` better describes what the function does

promql/engine_test.go
```diff
@@ -3256,6 +3259,914 @@ func TestNativeFloatHistogramRate(t *testing.T) {
...
+					until:  5,
+					filter: defaultFilterFunc,
+				},
+				// OOO samples with counter reset at beginning of interval
+				{
+					from:  16,
+					until: 100,
+					filter: func(v int64) bool {
+						return v == 16
+					},
```
fionaliao, line 3457: Won't this batch be in-order? Because it's the last batch and its `from` is greater than or equal to the `until`s of the previous batches.

promql/engine_test.go
```diff
@@ -3256,6 +3259,914 @@ func TestNativeFloatHistogramRate(t *testing.T) {
...
+			batches: []sampleBatch{
+				{
+					from:  0,
+					until: 100,
+					filter: func(v int64) bool {
+						return v == 16 // first sample captured within interval
+					},
+				},
+			},
+			// Since the counter reset is at the first sample in the 5-minute interval
```
fionaliao, line 3324:
```suggestion
			// Since the counter reset is at the first sample in the query interval
```

We are only querying samples between 4-5 minutes, so not sure if saying 5-minute interval makes sense

promql/engine_test.go
```diff
@@ -3256,6 +3259,914 @@ func TestNativeFloatHistogramRate(t *testing.T) {
...
+					filter: defaultFilterFunc,
+				},
+				// OOO samples with counter resets
+				{
+					from:  17,
+					until: 200,
+					filter: func(v int64) bool {
+						return v == 20
+					},
+				},
```
fionaliao, line 3525: Similar comment to above - I think these would be in-order samples

fionaliao (commented):
I like the way the tests are structured to make it easy to extend with more in-order/ooo/counter reset test cases. I made a few comments on the first tests that also apply to the other tests but no major issues :+1:

promql/engine_test.go
fionaliao: One of the cases of this was fixed, but I found one that wasn't and fixed it myself in d59e73107251a9f7742883cb86a1062b5688b8b6

fionaliao:
I rebased and fixed some small issues, now merging

PR #556, opened 2023-10-30
Don't drop rule name and index information from rule evaluation logs when adding trace ID

charleskorn:
This PR fixes an issue introduced in https://github.com/grafana/mimir-prometheus/pull/541 that was spotted as part of the upstream review: https://github.com/prometheus/prometheus/pull/13034#pullrequestreview-1703822504


56quarters (approved):
LGTM

PR #558, opened 2023-10-31
Cedwards/ooo native histogram flag

carrieedwards:
This PR adds a new flag, EnableOOONativeHistograms. This flag has a default state of false, and can be set to true in order to test the functionality of out-of-order native histograms. The status of the flag is only taken into consideration if both the EnableNativeHistograms flag is set to true, and the OutOfOrderTimeWindow flag has a value > 0.


cmd/prometheus/main.go
```diff
@@ -205,6 +205,9 @@ func (c *flagConfig) setFeatureListOptions(logger log.Logger) error {
 				// Change global variable. Hacky, but it's hard to pass new option or default to unmarshaller.
 				config.DefaultConfig.GlobalConfig.ScrapeProtocols = config.DefaultNativeHistogramScrapeProtocols
 				level.Info(logger).Log("msg", "Experimental native histogram support enabled. Changed default scrape_protocols to prefer PrometheusProto format.", "global.scrape_protocols", fmt.Sprintf("%v", config.DefaultConfig.GlobalConfig.ScrapeProtocols))
+			case "ooo-native-histograms":
+				c.tsdb.EnableOOONativeHistograms = false
```
jesusvazquez, line 209: Should this be true?

tsdb/db.go
```diff
@@ -197,6 +197,11 @@ type Options struct {
 	// EnableNativeHistograms enables the ingestion of native histograms.
 	EnableNativeHistograms bool
 
+	// EnableOOONativeHistograms enables the ingestion of OOO native histograms.
+	// It will only take effect if EnableNativeHistograms is set to true and the
+	// OutOfOrderTimeWindow is > 0
+	EnableOOONativeHistograms bool
```
jesusvazquez, line 203: Would we want to drop this flag once OOO for native histograms is working? If so maybe we should mention it here.

I think OOO ingestion might be optional always because it enables a set of things WBL OOO etc that but OOO Native histograms seems like a temporary flag from the start. What do you think?

jesusvazquez (commented):
I think its looking good 👍 Left a couple questions.

tsdb/db.go
carrieedwards: Yes, the plan is to remove it after we have testing OOO native histogram ingestion

cmd/prometheus/main.go
carrieedwards: I set it to false so that this feature is off by default and can be turned on for testing

tsdb/db_test.go
```diff
@@ -7179,6 +7200,112 @@ func TestNativeHistogramFlag(t *testing.T) {
...
+		require.NoError(t, app.Commit())
+
+		q, err := db.Querier(math.MinInt, math.MaxInt64)
+		require.NoError(t, err)
+		act := query(t, q, labels.MustNewMatcher(labels.MatchEqual, "foo", "bar"))
+		require.Equal(t, map[string][]chunks.Sample{
+			l.String(): {sample{t: 100, h: h}},
+		}, act)
+	})
+	t.Run("Test OOO native histograms if Native Histograms are disabled", func(t *testing.T) {
```
fionaliao, line 7247: Nit: inconsistent native histograms capitalisation

tsdb/head_append.go
```diff
@@ -467,6 +467,9 @@ func (s *memSeries) appendableHistogram(t int64, h *histogram.Histogram, headMax
 
 	// The sample cannot go in the in-order chunk. Check if it can go in the out-of-order chunk.
 	if oooTimeWindow > 0 && t >= headMaxt-oooTimeWindow {
+		if !oooHistogramsEnabled {
+			return true, headMaxt - t, storage.ErrNativeHistogramsDisabled
```
fionaliao, line 471: Should the error be `ErrOOONativeHistogramsDisabled` instead?

fionaliao (approved):
LGTM

PR #563, opened 2023-11-06
Expand OOO native histogram tests to include gauge histograms

carrieedwards:
This PR expands OOO native histogram tests to include testing for gauge-type histograms. This is to check that gauge histograms, which do not have counter resets, are handled properly, with the correct CounterResetHint of histogram.Gauge type.

### DIFF
--- a/tsdb/head_test.go
+++ b/tsdb/head_test.go
@@ -4812,7 +4812,7 @@ func testWBLReplay(t *testing.T, scenario sampleTypeScenario) {
 	require.NoError(t, err)
 	require.NoError(t, h.Init(0))
 
-	var expOOOSamples []sample
+	var expOOOSamples []chunks.Sample
 	l := labels.FromStrings("foo", "bar")
 	appendSample := func(mins int64, isOOO bool) {
 		app := h.Appender(context.Background())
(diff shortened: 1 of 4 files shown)


tsdb/head_test.go
```diff
@@ -4709,10 +4709,10 @@ func testWBLReplay(t *testing.T, scenario sampleTypeScenario) {
 	// TODO(fionaliao): understand counter reset behaviour, might want to modify this later
 	for i := range expOOOSamples {
 		if i != 0 {
-			if expOOOSamples[i].h != nil {
+			if expOOOSamples[i].h != nil && expOOOSamples[i].h.CounterResetHint != histogram.GaugeType {
```
fionaliao, line 4712: We could extract the part in `requireEqualSamples` that actually compares an expected sample slice with the actual sample slice into a separate function, and then in these tests that just have a sample slice and no series to samples map, call that extracted function instead of `require.Equal(t, expOOOSamples, actOOOSamples)`. That way we can avoid repeating the gauge type check in individual tests.

fionaliao (commented):
Adding gauge scenarios to `sampleTypeScenarios` looks good to me, I do have a comment about refactoring requireEqualSamples which I think is worth doing in this PR. Also I think it's worth rebasing the `ooo-native-histograms` branch as the requireEqualSamples function signature has changed (a ignoreCounterResets parameter was added in https://github.com/grafana/mimir-prometheus/pull/565)

fionaliao (approved):

PR #564, opened 2023-11-07
Fix histogram append error

fionaliao:
We should check counterReset condition rather than okToAppend because if there's a counter reset, okToAppend is always set to false.


carrieedwards (approved):

fionaliao:
@bboreham No upstream PR yet. I'm switching to a different approach where we'll be able to remove these two errors completely, so I'm going to close this PR.

fionaliao:
We didn't go for a different approach (i.e. getting rid of the appendOnly boolO) in the end so weren't actually able to remove the two errors - see https://github.com/prometheus/prometheus/pull/13123#issuecomment-1825950407 for more details.

Therefore I opened a prometheus PR to fix the error messages: https://github.com/prometheus/prometheus/pull/13201

PR #566, opened 2023-11-08
Fix issue where a "cannot populate chunk XXX from block XXX: not found" error is returned when head compaction occurs between retrieving the list of chunks for a query and reading the chunk itself

charleskorn:
If head compaction occurs after obtaining the chunk reference, but before reading the chunk, it's expected that `ChunkReader.Chunk()` will return `storage.NotFound`. In this case, the samples in the affected chunk will be included in other chunks, so we should ignore the error.

This PR is based on a similar change in https://github.com/grafana/mimir-prometheus/pull/528


charleskorn:
This is not needed, and would hide legitimate issues like https://github.com/prometheus/prometheus/pull/13115.

PR #567, opened 2023-11-13
TSDB: make the global hash lookup table smaller

bboreham:
This is the same as https://github.com/prometheus/prometheus/pull/13040 upstream; I think it's worth going ahead of upstream to save RAM.

Instead of a map of slices of `*memSeries`, ready for any of them to hold series where hash values collide, split into a map of `*memSeries` and a map of slices which is usually empty, since hash collisions are a one-in-a-billion thing.

The `del` method gets more complicated, to maintain the invariant that a series is only in one of the two maps.

Preceded by some refactoring to make the change in `gc()` simpler - see individual commits.

Previously the path to access was: `head->stripeSeries->hashes->slice->*memSeries`.
Now it is: `head->stripeSeries->hashes->*memSeries`.

I see an 80-byte reduction in allocations per series in the benchmark:
```
goos: darwin
goarch: arm64
               │    before    │             after             │
               │    sec/op    │    sec/op     vs base         │
CreateSeries-8   1.262µ ± 12%   1.236µ ± 22%  ~ (p=0.937 n=6)

               │    before    │               after                │
               │     B/op     │     B/op      vs base              │
CreateSeries-8   1.116Ki ± 0%   1.037Ki ± 6%  -7.09% (p=0.002 n=6)

               │   before   │               after               │
               │ allocs/op  │ allocs/op   vs base               │
CreateSeries-8   5.000 ± 0%   4.000 ± 0%  -20.00% (p=0.002 n=6)
```

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1808,7 +1808,7 @@ func (s *stripeSeries) gc(mint int64, minOOOMmapRef chunks.ChunkDiskMapperRef) (
 	minMmapFile = math.MaxInt32
 
 	// For one series, truncate old chunks and check if any chunks left. If not, mark as deleted and collect the ID.
-	check := func(i int, hash uint64, series *memSeries, deletedForCallback map[chunks.HeadSeriesRef]labels.Labels) {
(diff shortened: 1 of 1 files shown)


bboreham:
Closing; this is merged upstream now.

PR #568, opened 2023-11-17
Support for computing and keeping secondary hash per series

pstibrany:
This PR adds support for computing and storing "secondary hash" for each series in Head.

We plan to use this in Mimir to compute ingesters's "owned series", for a better implementation of per-series limits.

(Code created by @pr00se, I'm just sending PRs while he's away)


tsdb/head.go
```diff
@@ -2258,3 +2274,19 @@ func (h *Head) updateWALReplayStatusRead(current int) {
 
 	h.stats.WALReplayStatus.Current = current
 }
+
+// ForEachSecondaryHash iterates over all series in the Head, and passes cached secondary hash of the series
+// to the function. No locks are held when function is called. Series may get deleted while the
```
pracucci, line 2279: > No locks are held when function is called

The stripe lock is held.

pracucci (approved):

tsdb/head.go
pracucci: > Series may get deleted while the function is running

Is it true, given we keep the stripe lock?

tsdb/head.go
```diff
@@ -2258,3 +2274,19 @@ func (h *Head) updateWALReplayStatusRead(current int) {
...
+// function is running.
+func (h *Head) ForEachSecondaryHash(fn func(secondaryHash uint32)) {
+	for i := 0; i < h.series.size; i++ {
+		h.series.locks[i].RLock()
```
pracucci, line 2283: If we're fine with no atomicity, then I think a way to avoid keeping the lock while the function is running is:

- With stripe lock: store the secondary hashes in a slice
- Without stripe lock: iterate the secondary hashes and call the function (eventually pass all hashes in a single function call)

pracucci: > eventually pass all hashes in a single function call

Don't know the logic on Mimir side yet, but I've the feeling that if we do it there may be some optimizations we could do to lookup which series are still owned.

tsdb/head.go
pstibrany: You're right, I missed the stripe lock 🤦 . It should be fixed now.

tsdb/head.go
pstibrany: > If we're fine with no atomicity, then I think a way to avoid keeping the lock while the function is running is:
> 
> * With stripe lock: store the secondary hashes in a slice
> * Without stripe lock: iterate the secondary hashes and call the function (eventually pass all hashes in a single function call)

It now works this way.

pracucci (approved):
Thanks for addressing my feedback. LGTM!

PR #571, opened 2023-12-07
Ensure manipulating the returned FastRegexMatcher.SetMatches() doesn't pollute the matcher's internal state

pracucci:
We're investigating some invalid query results which apparently happen when we have alternation regexp matchers (and so `FastRegexMatcher.SetMatches()` returns something). I have a weak theory that there may be a case when we manipulate the returned slice, thus polluting the cached `FastRegexMatcher`.

To avoid such headache, in this PR I do a simple change to always return a copy from `FastRegexMatcher.SetMatches()`.

### DIFF
--- a/model/labels/regexp.go
+++ b/model/labels/regexp.go
@@ -21,6 +21,7 @@ import (
 	"github.com/dgraph-io/ristretto"
 	"github.com/grafana/regexp"
 	"github.com/grafana/regexp/syntax"
+	"golang.org/x/exp/slices"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


model/labels/regexp.go
```diff
@@ -339,7 +339,13 @@ func (m *FastRegexMatcher) MatchString(s string) bool {
 }
 
 func (m *FastRegexMatcher) SetMatches() []string {
-	return m.setMatches
+	// IMPORTANT: always return a copy, otherwise if the caller manipulate this slice it will
+	// also get manipulated in the cached FastRegexMatcher instance.
+	ret := make([]string, 0, len(m.setMatches))
+	for _, value := range m.setMatches {
+		ret = append(ret, value)
```
bboreham, line 346: does this reduce to `return append(ret, m.setMatches...)` ?
There's also `slices.Clone`

krajorama (approved):

model/labels/regexp.go
pracucci: I didn't know about `slices.Clone()` but looks perfect. Used it in https://github.com/grafana/mimir-prometheus/pull/571/commits/0a92839843f36d2d246fa7e73203c236408432de

PR #574, opened 2023-12-12
Add test for counter resets when compacting ooo native histograms

fionaliao:


carrieedwards (approved):
This test seems very thorough. Great job!

PR #575, opened 2023-12-12
PostingsForMatchersCache: don't create new spans, reduce logging verbosity

dimitarvdimitrov:
Follow-up of https://github.com/grafana/mimir-prometheus/pull/553

Creating a new span for each promQL matcher for each query can end up being too verbose to navigate in grafana. This PR replaces the span with a trace log.

Overall starting a span was also expensive - according to profiles it's about 9% of the CPU of PostingsForMatchers and it's about 25% of the memory allocations. Adding events is about 2% of the CPU and ~5% of the allocations.

#### CPU

Screenshot 2023-12-12 at 18 21 52

<details>
<summary>CPU Details</summary>

Screenshot 2023-12-12 at 18 23 21
Screenshot 2023-12-12 at 18 23 33
Screenshot 2023-12-12 at 18 23 37

</details>

#### Allocations
Screenshot 2023-12-12 at 18 35 08

In addition to that it removes some log lines in `postingsForMatchersPromise` because I think they just log the same thing multiple times. I also removed some calls to AddEvent to amortize the cost of tracing further.

### DIFF
--- a/tsdb/postings_for_matchers_cache.go
+++ b/tsdb/postings_for_matchers_cache.go
@@ -91,7 +91,7 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
 	span := trace.SpanFromContext(ctx)
 	defer func(startTime time.Time) {
 		span.AddEvent(
-			"returned postings",
+			"PostingsForMatchers returned",
 			trace.WithAttributes(attribute.Bool("concurrent", concurrent), c.ttlAttrib, c.forceAttrib, attribute.Stringer("duration", time.Since(startTime))),
 		)
 	}(time.Now())
--- a/tsdb/postings_for_matchers_cache.go
+++ b/tsdb/postings_for_matchers_cache.go
@@ -3,6 +3,7 @@ package tsdb
 import (
 	"container/list"
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 	"time"
@@ -87,12 +88,13 @@ type PostingsForMatchersCache struct {
 }
 
 func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix IndexPostingsReader, concurrent bool, ms ...*labels.Matcher) (index.Postings, error) {
-	ctx, span := c.tracer.Start(ctx, "PostingsForMatchersCache.PostingsForMatchers", trace.WithAttributes(
-		attribute.Bool("concurrent", concurrent),
-		c.ttlAttrib,
-		c.forceAttrib,
-	))
-	defer span.End()
+	span := trace.SpanFromContext(ctx)
+	defer func(startTime time.Time) {
+		span.AddEvent(
+			"returned postings",
+			trace.WithAttributes(attribute.Bool("concurrent", concurrent), c.ttlAttrib, c.forceAttrib, attribute.Stringer("duration", time.Since(startTime))),
+		)
+	}(time.Now())
 
 	if !concurrent && !c.force {
 		span.AddEvent("cache not used")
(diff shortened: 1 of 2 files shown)


tsdb/postings_for_matchers_cache.go
```diff
@@ -178,15 +168,8 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Contex
...
 	if postings, err := c.postingsForMatchers(context.Background(), ix, ms...); err != nil {
-		span.AddEvent("postingsForMatchers failed", trace.WithAttributes(
-			attribute.String("cache_key", key),
-			attribute.String("err", err.Error()),
-		))
 		promise.err = err
 	} else {
-		span.AddEvent("postingsForMatchers succeeded", trace.WithAttributes(
-			attribute.String("cache_key", key),
-		))
```
dimitarvdimitrov, line 189: removed this event because it should be accounted for in the "returned postings" log line

tsdb/postings_for_matchers_cache.go
```diff
@@ -178,15 +168,8 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Contex
 	// FIXME: do we need to cancel the call to postingsForMatchers if all the callers waiting for the result have
 	// cancelled their context?
 	if postings, err := c.postingsForMatchers(context.Background(), ix, ms...); err != nil {
-		span.AddEvent("postingsForMatchers failed", trace.WithAttributes(
-			attribute.String("cache_key", key),
-			attribute.String("err", err.Error()),
-		))
```
dimitarvdimitrov, line 184: removed because this should be accounted for in the `RecordError` calls in `PostingsForMatchers`

tsdb/postings_for_matchers_cache.go
```diff
@@ -122,30 +124,18 @@ type postingsForMatcherPromise struct {
...
+			return nil, fmt.Errorf("completed postingsForMatchers promise, but context has error: %w", ctx.Err())
 		}
 		if p.err != nil {
-			span.AddEvent("postingsForMatchers promise completed with error", trace.WithAttributes(
-				attribute.String("err", p.err.Error()),
-			))
-			return nil, p.err
+			return nil, fmt.Errorf("postingsForMatchers promise completed with error: %w", p.err)
 		}
-		span.AddEvent("postingsForMatchers promise completed successfully")
```
dimitarvdimitrov, line 148: adding events means that the error will be duplicated with the `RecordError` call in `PostingsForMatchers`. It's also more expensive.


"postingsForMatchers promise completed successfully" should be accounted for in "returned postings"

tsdb/postings_for_matchers_cache.go
```diff
@@ -87,12 +88,13 @@ type PostingsForMatchersCache struct {
...
-		c.ttlAttrib,
-		c.forceAttrib,
-	))
-	defer span.End()
+	span := trace.SpanFromContext(ctx)
+	defer func(startTime time.Time) {
+		span.AddEvent(
+			"returned postings",
+			trace.WithAttributes(attribute.Bool("concurrent", concurrent), c.ttlAttrib, c.forceAttrib, attribute.Stringer("duration", time.Since(startTime))),
+		)
```
aknuds1, line 96: The "returned postings" event will be added even if the call actually failed AFAICT, is it potentially misleading?

aknuds1 (commented):
You'll need to fix the tests. I also left a question about the deferred event adding.

tsdb/postings_for_matchers_cache.go
dimitarvdimitrov: good point. changed in https://github.com/grafana/mimir-prometheus/pull/575/commits/a14f40efb932796d0fce13154842c2104c8dd1a1

aknuds1 (approved):
LGTM, thanks

bboreham:
I didn't see these spans for ingester queries in Mimir, so created https://github.com/grafana/mimir/pull/8242

PR #576, opened 2023-12-13
`tsdb.IndexReader`: Add `LabelValuesFor`/`LabelValuesExcluding` methods

aknuds1:
Add `LabelValuesFor` and `LabelValuesExcluding` methods to `tsdb.IndexReader`. `LabelValuesFor` returns an iterator over label values which postings intersect with a certain postings set, `LabelValuesExcluding` is the inverse version. The reason for adding an inverse version of `LabelValuesFor` is to optimize the case when you only have postings you want to exclude.

This is a dependency of implementing label values streaming in #517 (which is not fully optimized, I've extracted those into another branch).

#### TODO

- [ ] Mark code as experimental, and flag that it should be reviewed in depth before considering non-experimental
- [x] Write some more tests

#### Benchmarks
Benchmarking stats, label value streaming vs non-streaming fetching of label values, below:

<details>
<summary>Block benchmark stats</summary>

```console
goos: darwin
goarch: arm64
                                                                         │   block-6.txt   │    labelvaluesstream-block-6.txt    │
                                                                         │     sec/op      │   sec/op     vs base                │
Querier/Block/labelValuesWithMatchers/i_with_i="1"-12                          1.244m ± 1%   1.090m ± 0%   -12.36% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1"-12                          85.99m ± 0%   15.31m ± 1%   -82.20% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="^.+$"-12                      79.753m ± 0%   1.915m ± 1%   -97.60% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j!="foo"-12                155.86m ± 1%   36.12m ± 0%   -76.83% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-12               1115.44µ ± 1%   93.11µ ± 0%   -91.65% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-12           1115.90µ ± 2%   92.98µ ± 1%   -91.67% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="X",j!="foo"-12                 220.7m ± 0%   128.0m ± 0%   -42.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-12      157.68m ± 0%   37.69m ± 1%   -76.10% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=533701-12               1.023m ± 1%   5.502m ± 0%  +437.62% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=20-12                   1.029m ± 1%   6.342m ± 0%  +516.22% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-12            117.02m ± 1%   18.22m ± 0%   -84.43% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_j!="foo"-12                       280.1m ± 0%   105.8m ± 2%   -62.24% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_i="1"-12                      2314538.0n ± 0%   717.7n ± 1%   -99.97% (p=0.002 n=6)
geomean                                                                        16.03m        3.089m        -80.73%

                                                                         │   block-6.txt   │    labelvaluesstream-block-6.txt     │
                                                                         │      B/op       │     B/op      vs base                │
Querier/Block/labelValuesWithMatchers/i_with_i="1"-12                       1606053.0 ± 0%     144.0 ± 0%   -99.99% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1"-12                         17.67Mi ± 0%   12.56Mi ± 0%   -28.93% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="^.+$"-12                   40290280.0 ± 0%     144.0 ± 0%  -100.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j!="foo"-12               17.672Mi ± 0%   8.490Mi ± 0%   -51.96% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-12              1615325.0 ± 0%     136.0 ± 0%   -99.99% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-12          1614366.0 ± 0%     200.0 ± 0%   -99.99% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="X",j!="foo"-12               17.672Mi ± 0%   4.578Mi ± 0%   -74.09% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-12      19.20Mi ± 0%   10.02Mi ± 0%   -47.82% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=533701-12            1614720.0 ± 0%     456.0 ± 0%   -99.97% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=20-12                1615407.0 ± 0%     648.0 ± 0%   -99.96% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-12           21.821Mi ± 0%   3.209Mi ± 0%   -85.30% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_j!="foo"-12                      7.398Ki ± 0%   1.750Ki ± 0%   -76.35% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_i="1"-12                         6437.00 ± 0%     72.00 ± 0%   -98.88% (p=0.002 n=6)
geomean                                                                       2.231Mi        13.04Ki        -99.43%

                                                                         │   block-6.txt   │    labelvaluesstream-block-6.txt    │
                                                                         │    allocs/op    │  allocs/op   vs base                │
Querier/Block/labelValuesWithMatchers/i_with_i="1"-12                          12.000 ± 0%    1.000 ± 0%   -91.67% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1"-12                          300.0k ± 0%   100.0k ± 0%   -66.66% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="^.+$"-12                   400076.000 ± 0%    1.000 ± 0%  -100.00% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j!="foo"-12                 300.1k ± 0%   100.0k ± 0%   -66.66% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-12                 16.000 ± 0%    6.000 ± 0%   -62.50% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-12             17.000 ± 0%    8.000 ± 0%   -52.94% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="X",j!="foo"-12                 300.1k ± 0%   100.0k ± 0%   -66.67% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-12       300.1k ± 0%   100.0k ± 0%   -66.66% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=533701-12               21.000 ± 0%    9.000 ± 0%   -57.14% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=20-12                    36.00 ± 0%    14.00 ± 0%   -61.11% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-12            308.89k ± 0%   30.31k ± 0%   -90.19% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_j!="foo"-12                       146.00 ± 0%    36.00 ± 0%   -75.34% (p=0.002 n=6)
Querier/Block/labelValuesWithMatchers/n_with_i="1"-12                         134.000 ± 0%    4.000 ± 0%   -97.01% (p=0.002 n=6)
geomean                                                                        2.293k         218.9        -90.45%
```
</details>

<details>
<summary>Head benchmark stats</summary>

```console
goos: darwin
goarch: arm64
                                                                        │   head-6.txt    │     labelvaluesstream-head-6.txt      │
                                                                        │     sec/op      │    sec/op      vs base                │
Querier/Head/labelValuesWithMatchers/i_with_i="1"-12                          1.221m ± 3%    1.420m ±  5%   +16.38% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1"-12                          88.09m ± 0%   103.74m ±  8%   +17.77% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="^.+$"-12                       81.71m ± 1%    23.52m ±  1%   -71.21% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j!="foo"-12                 104.1m ± 1%    120.0m ±  2%   +15.28% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-12             1029292.0n ± 1%    714.2n ±  0%   -99.93% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-12         1028327.0n ± 1%    773.9n ±  0%   -99.92% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="X",j!="foo"-12                 78.45m ± 0%    65.70m ±  1%   -16.25% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-12       108.9m ± 2%    122.7m ±  1%   +12.64% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=533701-12               1.029m ± 0%    2.085m ± 90%  +102.50% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=20-12                   1.028m ± 0%    2.081m ±  4%  +102.43% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-12             377.1m ± 1%    352.0m ±  1%    -6.64% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/n_with_j!="foo"-12                      227.36m ± 0%    72.36m ±  0%   -68.17% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/n_with_i="1"-12                         4996.0n ± 2%    574.3n ±  0%   -88.50% (p=0.002 n=6)
geomean                                                                       9.273m         2.455m         -73.53%

                                                                        │   head-6.txt   │     labelvaluesstream-head-6.txt     │
                                                                        │      B/op      │     B/op      vs base                │
Querier/Head/labelValuesWithMatchers/i_with_i="1"-12                        1.532Mi ± 0%   1.531Mi ± 0%    -0.02% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1"-12                        16.15Mi ± 0%   22.60Mi ± 0%   +39.95% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="^.+$"-12                    36.898Mi ± 0%   1.531Mi ± 0%   -95.85% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j!="foo"-12               16.15Mi ± 0%   18.53Mi ± 0%   +14.74% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-12             1614293.0 ± 0%     152.0 ± 0%   -99.99% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-12         1614357.0 ± 0%     216.0 ± 0%   -99.99% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="X",j!="foo"-12               16.15Mi ± 0%   13.85Mi ± 0%   -14.22% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-12     17.68Mi ± 0%   20.06Mi ± 0%   +13.46% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=533701-12             1.540Mi ± 0%   6.321Mi ± 0%  +310.52% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=20-12                 1.540Mi ± 0%   6.321Mi ± 0%  +310.45% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-12           20.30Mi ± 0%   20.25Mi ± 0%    -0.24% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/n_with_j!="foo"-12                     5.844Ki ± 0%   4.891Ki ± 0%   -16.31% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/n_with_i="1"-12                        4869.00 ± 0%     56.00 ± 0%   -98.85% (p=0.002 n=6)
geomean                                                                     2.069Mi        368.2Ki        -82.62%

                                                                        │   head-6.txt    │     labelvaluesstream-head-6.txt      │
                                                                        │    allocs/op    │  allocs/op   vs base                  │
Querier/Head/labelValuesWithMatchers/i_with_i="1"-12                          10.000 ± 0%    2.000 ± 0%   -80.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1"-12                          200.0k ± 0%   100.1k ± 0%   -49.98% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="^.+$"-12                   300075.000 ± 0%    2.000 ± 0%  -100.00% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j!="foo"-12                 200.0k ± 0%   100.1k ± 0%   -49.98% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"X.+"-12                 14.000 ± 0%    6.000 ± 0%   -57.14% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",j=~"XXX|YYY"-12             17.000 ± 0%    9.000 ± 0%   -47.06% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="X",j!="foo"-12              200050.00 ± 0%    43.00 ± 0%   -99.98% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_n="1",i=~"^.*$",j!="foo"-12       200.1k ± 0%   100.1k ± 0%   -49.98% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=533701-12                13.00 ± 0%    13.00 ± 0%         ~ (p=1.000 n=6) ¹
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=20-12                    16.00 ± 0%    18.00 ± 0%   +12.50% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/i_with_i_times_n=~"12.*""-12             208.9k ± 0%   117.7k ± 0%   -43.65% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/n_with_j!="foo"-12                       103.00 ± 0%    15.00 ± 0%   -85.44% (p=0.002 n=6)
Querier/Head/labelValuesWithMatchers/n_with_i="1"-12                          92.000 ± 0%    3.000 ± 0%   -96.74% (p=0.002 n=6)
geomean                                                                       1.606k         144.1        -91.03%
¹ all samples are equal
```
</details>


tsdb/block.go
```diff
@@ -93,6 +93,10 @@ type IndexReader interface {
 	// the series labels mod shardCount, using a hash function which is consistent over time.
 	ShardedPostings(p index.Postings, shardIndex, shardCount uint64) index.Postings
 
+	// LabelValuesIntersectingPostings returns a LabelValues iterator, for values of the label with the specified
+	// name, where the postings intersect with p.
+	LabelValuesIntersectingPostings(name string, p index.Postings) storage.LabelValues
```
colega, line 98: I don't think we're _intersecting_ here. This is actually a _multiple result_ version of the `LabelValueFor` below. How about we move it there below and call it `LabelValuesFor(p index.Postings, label string)` (for consistency with `LabelValueFor(ctx context.Context, id storage.SeriesRef, label string)`)?

storage/interface.go
```diff
@@ -439,3 +439,19 @@ type ChunkIterable interface {
...
+	// At returns the current label value.
+	At() string
+	// Err is the error that iteration eventually failed with.
+	// When an error occurs, the iterator cannot continue.
+	Err() error
+	// Warnings is a collection of warnings that have occurred during iteration.
+	// Warnings could be non-empty even if iteration has not failed with an error.
+	Warnings() annotations.Annotations
+	// Close the iterator.
+	Close() error
```
colega, line 456: Does it have to be closed by contract? If it does, we should mention it in the docs.

aknuds1: @colega yes, it's added because some `LabelValues` implementations need closing. Do you mean to add to the doc string that it needs calling? If so, what about all other `Close` methods, for example `storage.Storage.Close` doesn't mention it has to be called? I would say that when a type has a `Close` method, it's implied it has to be called.

tsdb/block.go
aknuds1: > I don't think we're intersecting here.

The semantics of this method is to find values for the label where the values' postings intersect with the `p` argument. How is it not intersecting?

storage/interface.go
colega: I think the current comment is useless, and it would be nice for the reader to understand that an iterator has to be closed and know whether it can be closed multiple times.

I wouldn't take anything as implied, implied means that the reader has to dig through all the implementations and double check.

tsdb/block.go
colega: That argument applies to every method that takes postings as input, but they don't say we're intersecting. This method has different naming and as a reader of the interface with 0 context I start wondering if it does something different from `LabelNamesFor` for example (which could be `LabelNamesIntersectingSeriesRef`).

aknuds1: Yep, having looked at `LabelValueFor`, it should be the same concept actually (the postings intersection is more of an implementation detail I think). Thanks!

tsdb/index/postings.go
```diff
@@ -418,6 +419,9 @@ type Postings interface {
 
 	// Err returns the last error of the iterator.
 	Err() error
+
+	// Reset the iterator to its initial state.
+	Reset()
```
colega, line 424: The idea of having an iterator of postings is to enable streaming and avoid having to load everything in memory. I can't see how a `Reset()` method fits into that pattern: we should have already discarded the iterated postings, so we can't reset to them.

tsdb/index/postings.go
```diff
@@ -660,6 +674,21 @@ func (it mergedPostings) Err() error {
...
 
+func (it *mergedPostings) Reset() {
+	// Reset the heap and its iterators
+	it.h = it.hCopy
+	for _, p := range it.h {
+		p.Reset()
+		// All iterators are expected to be initialized
+		if !p.Next() {
+			panic("couldn't issue initial Postings.Next()")
+		}
```
colega, line 685: What if one of the Postings is `emptyPostings`?

tsdb/index/postings.go
aknuds1: First of all, I should say that the `Reset` method is to enable optimizing, so you don't have to keep re-allocating iterators. As I recall it makes a real difference perf wise.

Your argument makes sense on the theoretical level, but in practice all the `Postings` implementations are memory based and I didn't have to jump through any hoops to implement the `Reset` method for them.

tsdb/block.go
aknuds1: Done.

storage/interface.go
aknuds1: What about the other `Close` methods in the same file though? They also don't state explicitly that they have to be called?

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func (it *intersectLabelValues) Close() error {
+	return nil
+}
+
+// LabelValuesFor returns LabelValues for the given label name in the series referred to by postings.
+func (p *MemPostings) LabelValuesFor(postings Postings, name string) storage.LabelValues {
+	p.mtx.RLock()
+	defer p.mtx.RUnlock()
```
colega, line 182: This is a potentially slow operation with it's current implementation and it's locking the entire head's MemPostings. I'd consider gathering the data you need and unlocking ASAP.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,235 @@
...
+	vals := make([]string, 0, len(e))
+	for val, srs := range e {
+		p.curPostings.list = srs
+		p.curPostings.Reset()
+		postings.Reset()
+
+		if checkIntersection(&p.curPostings, postings) {
+			vals = append(vals, val)
+		}
+	}
```
colega, line 199: Wouldn't `FindIntersectingPostings` be faster than this? It tackles all the postings in a single shot instead of resetting them again and again. With that implementation we also don't need the `Postings.Reset()` method at all.

storage/interface.go
aknuds1: Reviewing whether the `Close` method is actually needed within Mimir.

aknuds1: Ah yep, the `Close` method is required for the iterator returned by `Distributor.LabelValuesStream`, since it wraps a `gRPC` stream.

tsdb/index/postings.go
aknuds1: `newMergedPostings` only puts non-empty iterators on the heap, so it wouldn't happen unless `p.Reset()` has a bug.

tsdb/index/labelvalues.go
aknuds1: Would you make a copy of `e` and then unlock? I don't see any better way of doing it.

tsdb/index/labelvalues.go
aknuds1: Thanks for the tip, I can't recall if I've considered it already. Will have to check.

storage/interface.go
aknuds1: Pushed a revision fleshing out the doc comment more.

tsdb/index/labelvalues.go
aknuds1: I've reimplemented `MemPostings.LabelValuesFor` using `FindIntersectingPostings`, PTAL. I don't think I can do the same for `Reader.LabelValuesFor` though?

tsdb/index/labelvalues.go
aknuds1: Resolved in the new implementation.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,244 @@
...
+
+func (it *intersectLabelValues) Close() error {
+	return nil
+}
+
+// LabelValuesFor returns LabelValues for the given label name in the series referred to by postings.
+func (p *MemPostings) LabelValuesFor(postings Postings, name string) storage.LabelValues {
+	p.mtx.RLock()
+
+	e := p.m[name]
```
colega, line 182: Nit: you use `e` everywhere for `p.m[name]`, but I'm struggling to understand what `e` stands for 🤔

aknuds1: I've followed the existing convention in the `MemPostings` code, `e` should stand for entry.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,244 @@
...
+		p.mtx.RUnlock()
+		return storage.EmptyLabelValues()
+	}
+
+	// With thread safety in mind and due to random key ordering in map, we have to construct the array in memory
+	vals := make([]string, 0, len(e))
+	candidates := make([]Postings, 0, len(e))
+	for val, srs := range e {
+		vals = append(vals, val)
+		candidates = append(candidates, NewListPostings(srs))
```
colega, line 193: We're retaining the reference to `srs` here beyond the mutex, and `srs` is mutable.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,244 @@
...
+		return storage.ErrLabelValues(err)
+	}
+
+	// Filter the values, keeping only those with intersecting postings
+	slices.Sort(indexes)
+	for i := range vals {
+		if _, ok := slices.BinarySearch(indexes, i); !ok {
+			vals = slices.Delete(vals, i, i+1)
+		}
+	}
```
colega, line 208: There's a more efficient way of doing this: just go through the indexes and append on `vals` the `vals[i]` where `i` is each one of kept indexes.

tsdb/index/labelvalues.go
aknuds1: Ugh, thanks.

tsdb/index/labelvalues.go
colega: I realized my comment is hard to understand, this is what I mean: 
```suggestion
	// Filter the values, keeping only those with intersecting postings
	if len(vals) != len(indexes) {
		slices.Sort(indexes)
		for i, index := range indexes {
		    vals[i] = vals[index]
		}
		vals = vals[:len(indexes)]
	}
```

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,244 @@
...
+		d2 := encoding.NewDecbufAt(it.b, int(postingsOff), castagnoliTable)
+		if _, err := it.dec.PostingsInPlace(d2.Get(), &it.curPostings); err != nil {
+			it.err = fmt.Errorf("decode postings: %w", err)
+			return false
+		}
+
+		it.exhausted = v == it.lastVal
+
+		it.postings.Reset()
+		if checkIntersection(&it.curPostings, it.postings) {
```
colega, line 150: What if instead of having to reset the postings, we just expand them once? It has a cost, but also resetting the postings has a cost: if they're mergedPostings, we have to merge them all the time, and if they're already a ListPostings, then we just replace one slice with another. 

I would propose to remove the `Reset()` method from `Postings` and leave it only on `ListPostings` implementation. 

Then, when creating `intersectLabelValues` we check whether `postings` is a `*ListPostings`: if it is, we store it as is, and if it's not, we just expand it `NewListPostings(ExpandPostings(postings)...)`. Then we can reset knowing for sure that there's no performance penalty for resetting, and we don't need to modify all the implementations to support resetting. 

WDYT?

colega: Even nicer than this, in order to make it extensible, we could check here whether these postings are a `ResetterPostings` (postings with `Reset()`), in which case we would use and reset them, and if they're not, we'd expand and set ListPostings (which is a ResetterPostings).

tsdb/index/labelvalues.go
colega: I think it's a terrible name 😂 but convention wins readability. Thank you.

tsdb/head_read_test.go
```diff
@@ -552,3 +553,18 @@ func TestMemSeries_chunk(t *testing.T) {
...
+			head: &Head{
+				postings: &index.MemPostings{},
+			},
+		}
+
+		it := r.LabelValuesFor(index.EmptyPostings(), "test")
+		require.False(t, it.Next())
+		require.NoError(t, it.Err())
+		require.Empty(t, it.Warnings())
+	})
```
jesusvazquez, line 569: Do you want to add more unit tests here?

tsdb/index/index.go
```diff
@@ -1849,16 +1850,24 @@ type Decoder struct {
 
 // Postings returns a postings list for b and its number of elements.
 func (dec *Decoder) Postings(b []byte) (int, Postings, error) {
+	var p bigEndianPostings
+	n, err := dec.PostingsInPlace(b, &p)
+	return n, &p, err
+}
+
+func (dec *Decoder) PostingsInPlace(b []byte, p *bigEndianPostings) (int, error) {
```
jesusvazquez, line 1858: What do you mean by `InPlace`?

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+		d:        &d,
+		b:        r.b,
+		dec:      r.dec,
+		lastVal:  lastVal,
+		postings: postings,
+	}
+}
+
+type intersectLabelValuesV1 struct {
+	e           map[string]uint64
```
jesusvazquez, line 54: What do you mean by `e`? Elements in the postings list?

jesusvazquez (commented):
Left a few small comments and questions.

Also, is this something we are planning to upstream to Prometheus? Im worried about some of the changes we've done to the postings like the change in the decoder if we dont upstream that the codebase is drifting further and this is not straightforward to everyone

tsdb/index/labelvalues.go
jesusvazquez: Arve can you explain to me why would we need to Reset() to optimize reading? I think I lack a bit of context

dimitarvdimitrov (commented):
leaving some sparse comments, haven't reviewed the tests

aknuds1:
> Also, is this something we are planning to upstream to Prometheus? Im worried about some of the changes we've done to the postings like the change in the decoder if we dont upstream that the codebase is drifting further and this is not straightforward to everyone

@jesusvazquez yes, the plan is to upstream after battle testing w/ Mimir. I'm actually going to separate the resettable postings iterator changes from this PR, since they're only for performance and we don't need perfect performance during initial testing.

tsdb/index/labelvalues.go
aknuds1: Fixed.

tsdb/index/postings.go
aknuds1: I removed the `Postings.Reset` method from the PR @colega (keeping it in another branch), PTAL.

tsdb/index/labelvalues.go
aknuds1: @jesusvazquez resetting iterators is to avoid having to re-allocate new ones in the hot path, which I found was a serious performance hit during benchmarking. I have removed that part of the PR though in the meantime, keeping it in another branch.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+		vals = vals[:len(indexes)]
+	}
+
+	slices.Sort(vals)
+	return storage.NewListLabelValues(vals)
+}
+
+// checkIntersection returns whether p1 and p2 have at least one series in common.
+func checkIntersection(p1, p2 Postings) bool {
```
colega, line 216: Can I suggest renaming this method to `intersect`?
Then the usage would be `if intersect(curPostings, postings) {`.

OTOH, what if we get rid of it and do:
```
if Intersect(curPostings, postings).Next()
```

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+		dec:     r.dec,
+		lastVal: lastVal,
+		cloner:  NewPostingsCloner(postings),
+	}
+}
+
+type intersectLabelValuesV1 struct {
+	e      map[string]uint64
+	values []string
+	cloner *PostingsCloner
```
colega, line 56: Nit, but I'd rename this field to `postings` instead of `cloner`. 

There are two places where we'll refer to this field: 
- Here in the struct, where the type is annotated on the same line.
- Later in the usage of `it.cloner.Clone()` which needs an extra var to clarify that we're getting postings: `postings := it.cloner.Clone()`. However, if we call this `postings`, then we don't even need the var, and we can do `if checkIntersection(it.postings.Clone(), curPostings)` which IMO is more readable and also shorter.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+		_, curPostings, err := it.dec.Postings(d2.Get())
+		if err != nil {
+			it.err = fmt.Errorf("decode postings: %w", err)
+			return false
+		}
+
+		it.exhausted = v == it.lastVal
+
+		postings := it.cloner.Clone()
+		if checkIntersection(curPostings, postings) {
```
colega, line 150: As the corollary to the other two comments, if we rename `s/cloner/postings/` and we use `Intersect` then this could just be:
```suggestion
		if Intersect(it.postings.Clone(), postings).Next() {
```

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+			it.d.Uvarint()
+			// Label name
+			it.d.UvarintBytes()
+			it.skip -= it.d.Len()
+		} else {
+			it.d.Skip(it.skip)
+		}
+
+		// Label value
+		v := yoloString(it.d.UvarintBytes())
```
colega, line 136: I'm not sure about yolo-ing here. This would retain a reference to the byte slice of the file. Maybe we can start with a string allocation (once we've checked that the intersection isn't empty) and then optimize over time carefully?

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+		val := it.values[0]
+		it.values = it.values[1:]
+
+		postingsOff := it.e[val]
+		// Read from the postings table.
+		d := encoding.NewDecbufAt(it.b, int(postingsOff), castagnoliTable)
+		_, curPostings, err := it.dec.Postings(d.Get())
+		if err != nil {
+			it.err = fmt.Errorf("decode postings: %w", err)
+			return false
```
colega, line 75: We should also do `it.values = nil` here, otherwise next call to `Next()` will just proceed decoding next postings. We could also check `it.err == nil` in the loop condition, but the happy case is to not fail, and nobody should  call Next() on something that returned false anyway.

tsdb/index/labelvalues.go
aknuds1: The reason for implementing `checkIntersection` instead of using `Intersect` is to avoid having to allocate an iterator. I tried benchmarking your change, the mean results vs before the change are as follows:

* sec/op    => +10.82%
* B/op      => +26.29%
* allocs/op => +50.28%

I don't think some simplification of the code is worth this drop in performance.

tsdb/index/labelvalues.go
dimitarvdimitrov: Do you know how this change impacted benchmarks? It's doing significantly more copying now

aknuds1: @dimitarvdimitrov haven't gauged this yet, but can do so after updating this PR with the latest changes (from the mother PR, as it looks pretty ready to me).

tsdb/index/labelvalues.go
aknuds1: Good catch. I opted instead to do as I do elsewhere, i.e. return `false` immediately in `Next` if `it.err` is non-nil.

tsdb/index/labelvalues.go
aknuds1: Renaming to `intersect`.

storage/interface.go
```diff
@@ -439,3 +439,19 @@ type ChunkIterable interface {
 	// chunks of the series, sorted by min time.
 	Iterator(chunks.Iterator) chunks.Iterator
 }
+
+// LabelValues is an iterator over label values.
+type LabelValues interface {
```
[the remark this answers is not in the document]
aknuds1: Referering to `index` from `storage` will cause a dependency cycle at least.

tsdb/index/index.go
aknuds1: It means to avoid allocating, but I dropped this optimization.

tsdb/head_read_test.go
```diff
@@ -552,3 +553,18 @@ func TestMemSeries_chunk(t *testing.T) {
 		})
 	}
 }
+
+func TestHeadIndexReader_LabelValuesFor(t *testing.T) {
+	t.Run("empty postings", func(t *testing.T) {
```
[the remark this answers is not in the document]
aknuds1: Yes, I think we should.

storage/interface.go
```diff
@@ -439,3 +439,19 @@ type ChunkIterable interface {
...
+	// Next tries to advance the iterator and returns true if it could, false otherwise.
+	Next() bool
+	// At returns the current label value.
+	At() string
+	// Err is the error that iteration eventually failed with.
+	// When an error occurs, the iterator cannot continue.
+	Err() error
+	// Warnings is a collection of warnings that have occurred during iteration.
+	// Warnings could be non-empty even if iteration has not failed with an error.
+	Warnings() annotations.Annotations
```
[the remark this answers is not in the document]
aknuds1: Isn't it best to keep the ability to return warnings, to mirror the `LabelValues` method?

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+		} else {
+			it.d.Skip(it.skip)
+		}
+
+		// Label value
+		v := yoloString(it.d.UvarintBytes())
+
+		postingsOff := it.d.Uvarint64()
+		// Read from the postings table
+		d2 := encoding.NewDecbufAt(it.b, int(postingsOff), castagnoliTable)
```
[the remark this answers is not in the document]
aknuds1: Done.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+	"golang.org/x/exp/slices"
+
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/tsdb/encoding"
+	"github.com/prometheus/prometheus/util/annotations"
+)
+
+// LabelValuesFor returns LabelValues for the given label name in the series referred to by postings.
+func (r *Reader) LabelValuesFor(postings Postings, name string) storage.LabelValues {
+	if r.version == FormatV1 {
```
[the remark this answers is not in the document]
aknuds1: Done.

tsdb/index/labelvalues.go
aknuds1: Done.

tsdb/index/labelvalues.go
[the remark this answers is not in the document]
aknuds1: Done.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,332 @@
...
+
+		it.exhausted = v == it.lastVal
+
+		isMatch := intersect(curPostings, it.postings.Clone())
+		if it.inverted {
+			isMatch = !isMatch
+		}
+		if isMatch {
+			// Make sure to allocate a new string
+			it.cur = string(vb)
```
colega, line 181: There's no need for v/vb, because `string(val) == it.lastVal` doesn't allocate.
So we can do just this: (I called the variable `val` as opposed to `lastVal`)

```suggestion
		val := it.d.UvarintBytes()
		postingsOff := int(it.d.Uvarint64())
		// Read from the postings table
		postingsDec := encoding.NewDecbufAt(it.b, postingsOff, castagnoliTable)
		_, curPostings, err := it.dec.Postings(postingsDec.Get())
		if err != nil {
			it.err = fmt.Errorf("decode postings: %w", err)
			return false
		}

		it.exhausted = string(val) == it.lastVal

		isMatch := intersect(curPostings, it.postings.Clone())
		if it.inverted {
			isMatch = !isMatch
		}
		if isMatch {
			// Make sure to allocate a new string
			it.cur = string(val)
```

tsdb/index/labelvalues.go
colega: Thank you, it works but we can do slightly better, I posted a suggestion in https://github.com/grafana/mimir-prometheus/pull/576/files#r1452532101

Resolving this thread.

tsdb/block.go
```diff
@@ -106,6 +106,12 @@ type IndexReader interface {
 	// storage.ErrNotFound is returned as error.
 	LabelValueFor(ctx context.Context, id storage.SeriesRef, label string) (string, error)
 
+	// LabelValuesFor returns LabelValues for the given label name in the series referred to by postings.
+	LabelValuesFor(p index.Postings, name string) storage.LabelValues
+
+	// LabelValuesNotFor returns LabelValues for the given label name in the series *not* referred to by postings.
+	LabelValuesNotFor(p index.Postings, name string) storage.LabelValues
```
colega, line 113: I find `NotFor` really confusing, like "I want label values for something else, haha!".
I'd prefer it to be `LabelValuesForNot`...postings. Or maybe `LabelValuesExcluding(p, name)`?

IDK, naming is hard. 
This is not a blocker comment.

aknuds1: Yeah I also thought it could be a bit confusing. It would have been easier to disambiguate if we had kept the old name of `LabelValuesFor` :)

tsdb/index/labelvalues.go
colega: @dimitarvdimitrov note that there's (almost) no copying happening here: NewListPostings just stores the slice header provided.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,332 @@
...
+		return storage.EmptyLabelValues()
+	}
+
+	// With thread safety in mind and due to random key ordering in map, we have to construct the array in memory
+	vals := make([]string, 0, len(e))
+	candidates := make([]Postings, 0, len(e))
+	for val, srs := range e {
+		vals = append(vals, val)
+		candidates = append(candidates, NewListPostings(srs))
+	}
```
colega, line 233: There's one allocation happening for each label value, and we can easily avoid that:

```suggestion
	candidates := make([]Postings, 0, len(e))
	// Allocate a slice for all needed ListPostings, so no need to allocate them one by one.
	lps := make([]ListPostings, 0, len(e))
	for val, srs := range e {
		vals = append(vals, val)
		lps = append(lps, ListPostings{list: srs})
		candidates = append(candidates, &lps[len(lps)-1])
	}
```

Since `ListPostings` is part of this package, it should be fine to instantiate it without using the constructor.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+
+		// Filter the values, keeping only those without intersecting postings
+		keep := make([]string, 0, len(vals)-len(indexes))
+		for i, val := range vals {
+			if _, ok := indexes[i]; !ok {
+				keep = append(keep, val)
+			}
+		}
+		vals = keep
```
colega, line 265: I think we don't need the TODO and we don't need the `findIntersectingPostingsMap` function:

```suggestion
		indexes, err := FindIntersectingPostings(postings, candidates)
		p.mtx.RUnlock()
		if err != nil {
			return storage.ErrLabelValues(err)
		}

		if len(vals) == len(indexes) {
			return storage.EmptyLabelValues()
		}

		slices.Sort(indexes)
		keep := make([]string, 0, len(vals)-len(indexes))
		idx := 0
		for i, val := range vals {
			for idx < len(indexes) && indexes[idx] < i {
				idx++
			}
			if idx == len(indexes) || indexes[idx] != i {
				keep = append(keep, val)
			}
		}
		vals = keep
```

tsdb/index/labelvalues.go
colega: This thread is outdated. Using postings cloner was good idea here, @aknuds1 👍  thank you.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,245 @@
...
+			it.d.Skip(it.skip)
+		}
+
+		// Label value
+		v := yoloString(it.d.UvarintBytes())
+
+		postingsOff := it.d.Uvarint64()
+		// Read from the postings table
+		d2 := encoding.NewDecbufAt(it.b, int(postingsOff), castagnoliTable)
+		if _, err := it.dec.PostingsInPlace(d2.Get(), &it.curPostings); err != nil {
```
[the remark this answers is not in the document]
colega: Not asking for a change, but I have to say that I liked the idea of `PostingsInPlace`: otherwise this performs one allocation per each label value, even if it doesn't end up in the result set.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,332 @@
...
+			return false
+		}
+
+		it.exhausted = v == it.lastVal
+
+		isMatch := intersect(curPostings, it.postings.Clone())
+		if it.inverted {
+			isMatch = !isMatch
+		}
+		if isMatch {
```
colega, line 179: The fact that `inverted` is negative, makes it difficult to shorten the path here.

I would consider replacing `inverted` by `includeMatches` instead (switching the instantiation) and then we can just do:

```suggestion
		if isMatch == it.includeMatches {
```

tsdb/index/labelvalues.go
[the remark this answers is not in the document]
aknuds1: I have `PostingsInPlace` in a different branch, for the next iteration :)

tsdb/block.go
aknuds1: Renamed to `LabelValuesExcluding`.

tsdb/index/labelvalues.go
aknuds1: Thanks for the suggestion - I changed `inverted` booleans to `includeMatches`, does simplify the code.

tsdb/index/labelvalues.go
aknuds1: @colega for my understanding - `string(val) == it.lastVal` doesn't allocate unlike `s := string(val)`? Otherwise, why would you need `yoloString`, right?

aknuds1: I applied your suggestion manually btw. Can't commit the suggestion directly due to changes in the meantime.

tsdb/index/labelvalues.go
aknuds1: Thanks, nice observation.

tsdb/head_read_test.go
aknuds1: Done.

tsdb/head_read_test.go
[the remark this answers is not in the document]
aknuds1: Done.

tsdb/index/labelvalues.go
colega: `string(val) == it.lastVal` doesn't allocate because the compiler optimizes that into just comparing the underlying values.

colega (approved):
LGTM, great job!

This is a dense review, so I'd wait for at least one more review before merging.

tsdb/index/labelvalues_test.go
```diff
@@ -0,0 +1,195 @@
...
+	require.NoError(t, iw.AddSymbol("a"))
+	require.NoError(t, iw.AddSymbol("b"))
+
+	// Postings lists are only written if a series with the respective
+	// reference was added before.
+	require.NoError(t, iw.AddSeries(1, series[0]))
+	require.NoError(t, iw.AddSeries(2, series[1]))
+	require.NoError(t, iw.AddSeries(3, series[2]))
+	require.NoError(t, iw.AddSeries(4, series[3]))
+	require.NoError(t, iw.AddSeries(5, series[4]))
```
dimitarvdimitrov, line 43: can we use a loop instead or did you intentionally have it enumerated like this?

tsdb/index/labelvalues_test.go
```diff
@@ -0,0 +1,195 @@
...
+	iw, err := NewWriter(ctx, fn)
+	require.NoError(t, err)
+
+	series := []labels.Labels{
+		labels.FromStrings("a", "1", "b", "1"),
+		labels.FromStrings("a", "1", "b", "2"),
+		labels.FromStrings("a", "1", "b", "3"),
+		labels.FromStrings("a", "1", "b", "4"),
+		labels.FromStrings("a", "2", "b", "5"),
+	}
```
dimitarvdimitrov, line 27: do you plan to add more involved tests in a later PR? I'm thinking about how extensive the [tests for PostingsForMatchers](https://github.com/grafana/mimir-prometheus/blob/b7a7dad1e1d80bb35d63364b57d55f752086e555/tsdb/querier_test.go#L2730) are and  `TestReader_LabelValuesFor` looks simpler than that. Maybe I should defer to a prometheus maintainer (@jesusvazquez?) to decide, but I think some more involved testing now that we don't rely on PostingsForMatchers would be useful.

tsdb/index/postings.go
dimitarvdimitrov, line 1: Are there any functional changes in this file? I couldn't spot any

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+		vals = keep
+	}
+
+	slices.Sort(vals)
+	return storage.NewListLabelValues(vals)
+}
+
+// intersect returns whether p1 and p2 have at least one series in common.
+func intersect(p1, p2 Postings) bool {
```
dimitarvdimitrov, line 266: internal implementations aren't usually tested in Prometheus, right? So no need to test this?

dimitarvdimitrov (commented):
I have a doubt about how `LabelValuesExcluding` is implemented, I couldn't spot anything else major

tsdb/index/labelvalues_test.go
```diff
@@ -0,0 +1,195 @@
...
+	mp.Add(1, labels.FromStrings("a", "1"))
+	mp.Add(1, labels.FromStrings("b", "1"))
+	mp.Add(2, labels.FromStrings("a", "1"))
+	mp.Add(2, labels.FromStrings("b", "2"))
+	mp.Add(3, labels.FromStrings("a", "1"))
+	mp.Add(3, labels.FromStrings("b", "3"))
+	mp.Add(4, labels.FromStrings("a", "1"))
+	mp.Add(4, labels.FromStrings("b", "4"))
+	mp.Add(5, labels.FromStrings("a", "2"))
+	mp.Add(5, labels.FromStrings("b", "5"))
```
colega, line 175: We're adding a labelset for each series, not appending individual labels to series (that's not a valid operation, and it only casually works because of current implementation of the method).

```suggestion
	mp.Add(1, labels.FromStrings("a", "1", "b", "1"))
	mp.Add(2, labels.FromStrings("a", "1", "b", "2"))
	mp.Add(3, labels.FromStrings("a", "1", "b", "3"))
	mp.Add(4, labels.FromStrings("a", "1", "b", "4"))
	mp.Add(5, labels.FromStrings("a", "2", "b", "5"))
```

Can you also review the rest of the test cases please?

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,295 @@
...
+		for i, val := range vals {
+			for idx < len(indexes) && indexes[idx] < i {
+				idx++
+			}
+			if idx == len(indexes) || indexes[idx] != i {
+				keep = append(keep, val)
+			}
+		}
+		vals = keep
+	}
```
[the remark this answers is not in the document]
colega: I find it hard to follow the conversation with references to implementations in other repos. Can we have a test in this repo that would fail?

I tried modifying `TestMemPostings_LabelValuesExcluding` by adding `mp.Add(6, labels.FromStrings("b", "6"))` and the result correctly included `6` as that's where `a!="1"` also applies.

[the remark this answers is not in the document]
dimitarvdimitrov: here's a diff on this branch which causes TestMemPostings_LabelValuesExcluding to fail

```diff
diff --git a/tsdb/index/labelvalues_test.go b/tsdb/index/labelvalues_test.go
index 9f0a3ea58..470d5b893 100644
--- a/tsdb/index/labelvalues_test.go
+++ b/tsdb/index/labelvalues_test.go
@@ -173,6 +173,7 @@ func TestMemPostings_LabelValuesExcluding(t *testing.T) {
 	mp.Add(4, labels.FromStrings("b", "4"))
 	mp.Add(5, labels.FromStrings("a", "2"))
 	mp.Add(5, labels.FromStrings("b", "5"))
+	mp.Add(6, labels.FromStrings("a", "1", "b", "5"))
 	p := mp.Get("a", "1")
 
 	it := mp.LabelValuesExcluding(p, "b")
```

storage/labelvalues.go
```diff
@@ -0,0 +1,65 @@
...
+
+var emptyLabelValues = errLabelValues{}
+
+// EmptyLabelValues returns an empty LabelValues.
+func EmptyLabelValues() LabelValues {
+	return emptyLabelValues
+}
+
+// ListLabelValues is an iterator over a slice of label values.
+type ListLabelValues struct {
```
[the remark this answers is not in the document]
aknuds1: I'll try.

tsdb/head_read_test.go
```diff
@@ -552,3 +553,90 @@ func TestMemSeries_chunk(t *testing.T) {
 		})
 	}
 }
+
+func TestHeadIndexReader_LabelValuesFor(t *testing.T) {
+	mp := index.NewMemPostings()
+	mp.Add(1, labels.FromStrings("a", "1"))
+	mp.Add(1, labels.FromStrings("b", "1"))
```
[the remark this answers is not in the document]
aknuds1: I'm pretty sure I took inspiration from `TestMemPostingsStats`, which follows the same setup pattern. I guess that should be fixed up too, along the same lines?

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,295 @@
...
+
+func (it *intersectLabelValuesV1) Warnings() annotations.Annotations {
+	return nil
+}
+
+func (it *intersectLabelValuesV1) Close() error {
+	return nil
+}
+
+type intersectLabelValues struct {
```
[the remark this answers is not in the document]
aknuds1: Feel free to disagree, but my thinking was that v1 is legacy and v2 is current. Therefore it just felt superfluous to me to append "V2".

tsdb/index/labelvalues.go
aknuds1: I have to admit I'm not sure, but generally my philosophy is to not add tests for helper functions unless there's some real reason for it. Testing public behaviour is more important, and one wants to avoid nailing down implementation details.

tsdb/head_read_test.go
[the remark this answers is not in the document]
aknuds1: Done, thanks.

tsdb/head_read_test.go
```diff
@@ -552,3 +553,90 @@ func TestMemSeries_chunk(t *testing.T) {
...
+	it := r.LabelValuesFor(p, "b")
+
+	var vals []string
+	for it.Next() {
+		vals = append(vals, it.At())
+	}
+	require.NoError(t, it.Err())
+	require.Empty(t, it.Warnings())
+	// Note that b=5 is filtered out, since it has a=2, while our postings have a=1
+	require.Equal(t, []string{"1", "2", "3", "4"}, vals)
```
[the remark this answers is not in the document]
aknuds1: Done.

storage/interface.go
```diff
@@ -465,3 +465,19 @@ type ChunkIterable interface {
 	// chunks of the series, sorted by min time.
 	Iterator(chunks.Iterator) chunks.Iterator
 }
+
+// LabelValues is an iterator over label values.
```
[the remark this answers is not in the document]
aknuds1: Yep, sorted. Will specify this.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,295 @@
...
+		values:         vals,
+		postings:       NewPostingsCloner(postings),
+		b:              r.b,
+		dec:            r.dec,
+		includeMatches: includeMatches,
+	}
+}
+
+type intersectLabelValuesV1 struct {
+	seriesRefs     map[string]uint64
```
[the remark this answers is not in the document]
aknuds1: Nice catch, you're right. Thanks!

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,295 @@
...
+	candidates := make([]Postings, 0, len(e))
+	// Allocate a slice for all needed ListPostings, so no need to allocate them one by one.
+	lps := make([]ListPostings, 0, len(e))
+	for val, srs := range e {
+		vals = append(vals, val)
+		lps = append(lps, ListPostings{list: srs})
+		candidates = append(candidates, &lps[len(lps)-1])
+	}
+
+	indexes, err := FindIntersectingPostings(postings, candidates)
```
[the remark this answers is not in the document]
aknuds1: I haven't seen see any write latency increase with label values streaming in the cloud, but now that you've made me aware of it, I could try to look more closely.

tsdb/index/labelvalues_test.go
aknuds1: I believe I just copied this approach from `TestIndexRW_Postings`.

aknuds1: I'm converting into a loop though, I see it makes sense.

tsdb/index/labelvalues_test.go
aknuds1: I have to admit I haven't planned more tests currently. Might you have ideas on what kind of cases should be covered? @jesusvazquez?

tsdb/index/postings.go
aknuds1: I'll drop these. One change is just a basic renaming to reflect the move from heap to loser tree, whereas the other is supposed to be a performance improvement.

tsdb/index/labelvalues.go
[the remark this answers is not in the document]
aknuds1: Checked - I don't see any worsening in the write latency versus before enabling label values streaming.

tsdb/index/labelvalues.go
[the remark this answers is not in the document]
aknuds1: Thanks @dimitarvdimitrov, I'll try your test extension for myself.

colega:
(I've revoked my approval until the testcase supplied by @dimitarvdimitrov is solved)

aknuds1:
> (I've revoked my approval until the testcase supplied by @dimitarvdimitrov is solved)

@colega yep, looking into it now.

tsdb/head_read_test.go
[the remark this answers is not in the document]
dimitarvdimitrov: > I guess that should be fixed up too, along the same lines?

yes, I think that test is just wrong

tsdb/index/labelvalues.go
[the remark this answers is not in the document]
aknuds1: Thanks for catching this corner case @dimitarvdimitrov! I've added tests and fixed it. The updated definition I came up with for `LabelValuesExcluding` is that it should return label values for which the postings are not a strict _subset_ of the passed `postings` iterator, i.e. label values that have also non-excluded postings.

I added a corresponding test case to `TestLabelValuesWithMatchers`, to make sure that the non-streaming `LabelValues` method handles this corner case the same.

PTAL

tsdb/index/labelvalues_test.go
aknuds1: Added tests for your [comment](https://github.com/grafana/mimir-prometheus/pull/576#discussion_r1457806974) above. Will look into the others.

tsdb/block.go
```diff
@@ -106,6 +106,12 @@ type IndexReader interface {
 	// storage.ErrNotFound is returned as error.
 	LabelValueFor(ctx context.Context, id storage.SeriesRef, label string) (string, error)
 
+	// LabelValuesFor returns LabelValues for the given label name in the series referred to by postings.
+	LabelValuesFor(p index.Postings, name string) storage.LabelValues
+
+	// LabelValuesExcluding returns LabelValues for the given label name in the series *not* referred to by postings.
```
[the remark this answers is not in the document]
aknuds1: Done.

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,295 @@
...
+
+func (it *intersectLabelValuesV1) Next() bool {
+	if it.err != nil {
+		return false
+	}
+
+	// Look for a value with intersecting postings
+	for len(it.values) > 0 {
+		val := it.values[0]
+		it.values = it.values[1:]
```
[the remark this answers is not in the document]
aknuds1: Done.

aknuds1:
@dimitarvdimitrov @colega could you please review my fix for the bug Dimitar [discovered](https://github.com/grafana/mimir-prometheus/pull/576#discussion_r1457806974) and corresponding tests?

tsdb/index/labelvalues.go
```diff
@@ -0,0 +1,319 @@
...
+		if !p1.Next() {
+			// Exhausted p1, with only matches against p2
+			break
+		}
+
+		cur = p1.At()
+	}
+
+	return false
+}
```
colega, line 319: Can we rename `p1, p2` into meaningful `subset` and `set`? Otherwise the reader has to refer to the comment each time.

Also, I think we can do shorter:
```suggestion
// notSubset returns whether subset is not a subset of sub.
func notSubset(subset, set Postings) bool {
	// Look for a value in subset which is not in set.
	for subset.Next() {
		if cur := subset.At(); !set.Seek(cur) || set.At() != cur {
			// Cur is not in set, so subset is not a subset of set.
			return true
		}
	}

	// Couldn't find any value in subset which is not in set.
	return false
}
```

Also, I'd swap the order of params to match the ones in stdlib's `strings.Contains(str, substr)`.

Said that, I'd actually rewrite this method in a positive form:
```go
// contains returns whether subp is contained in p.
func contains(p, subp Postings) bool {
	for subp.Next() {
		if needle := subp.At(); !p.Seek(needle) || p.At() != needle {
			return true
		}
	}

	// Couldn't find any value in subp which is not in set.
	return false
}
```
And then use it as `!contains(it.postings.Clone(), curPostings)`

tsdb/index/postings.go
```diff
@@ -834,6 +834,41 @@ func FindIntersectingPostings(p Postings, candidates []Postings) (indexes []int,
 	return indexes, nil
 }
 
+// findNonIntersectingPostings checks the intersection of p and candidates[i] for each i in candidates,
+// if intersection is empty, i is added to the indexes returned.
```
colega, line 838: > if intersection is empty, i is added to the indexes returned.

Does this really do that?
```
		if p.At() != h.at() {
			indexes = append(indexes, h.popIndex())
```
I think it adds if it's not contained.

tsdb/index/labelvalues.go
aknuds1: I tried your `contains` function @colega, but then the following fails:

```
go test ./tsdb/index -run TestReader_Label
```

tsdb/index/postings.go
aknuds1: Thanks, I will look the definition of this function over again. Maybe I was a bit quick. You don't see a bug in there though, do you?

tsdb/index/labelvalues.go
aknuds1: Committed your suggestion, since it works AFAICT.

tsdb/index/postings.go
aknuds1: After reviewing the function again, I renamed it to `findNonSubsetPostings`, as what it should do is find postings in `candidates` which are not subsets (i.e. contained in) of `p`.  I think we're now on the same page here?

tsdb/index/labelvalues.go
colega: Of course my suggestion didn't work, I didn't reverse the `return true`/`return false` conditions :)

aknuds1: Ah it's that simple. I didn't have the time to analyze :P

colega:
It took me a while to understand why are we looking for the non-subset postings, I think it's not trivial and would appreciate a comment on that.

tsdb/index/labelvalues.go
colega: I'm unblocking the review, but I'd still rewrite the function in a positive form as that usually makes understanding the code easier.

colega (approved):

tsdb/index/labelvalues.go
aknuds1: OK, I can give it another shot.

bboreham:
> yes, the plan is to upstream 

I recommend posting a proposal issue first, to allow discussion before you show up with code.

aknuds1:
Good point @bboreham - making a note of it.

tsdb/index/labelvalues_test.go
dimitarvdimitrov: can you add the same tests for the head reader and mempostings?

dimitarvdimitrov (commented):
I only have more comment about testing, otherwise this looks good

tsdb/index/labelvalues_test.go
```diff
@@ -0,0 +1,194 @@
...
+	for i, s := range series {
+		require.NoError(t, iw.AddSeries(storage.SeriesRef(i+1), s))
+	}
+	require.NoError(t, iw.Close())
+
+	ir, err := NewFileReader(fn)
+	require.NoError(t, err)
+
+	// Obtain postings where a=1
+	p, err := ir.Postings(ctx, "a", "1")
```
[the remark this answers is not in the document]
aknuds1: You are right @dimitarvdimitrov - can't believe I missed this. Fixed.

tsdb/index/labelvalues.go
aknuds1: Converted to your positive form @colega.

tsdb/index/labelvalues_test.go
aknuds1: I added more cases to `Test{Reader,MemPostings}_LabelValues{For,Excluding}` @dimitarvdimitrov, PTAL.

dimitarvdimitrov (approved):
LGTM, nicely done @aknuds1 :)

tsdb/index/labelvalues.go
```diff
@@ -96,9 +97,11 @@ func (it *intersectLabelValuesV1) Next() bool {
 
 		isMatch := false
 		if it.includeMatches {
-			isMatch = intersect(curPostings, it.postings.Clone())
+			isMatch = intersect(it.postings.Clone(), curPostings)
 		} else {
-			isMatch = notSubset(curPostings, it.postings.Clone())
+			// We only want to include this value if curPostings is not fully contained
+			// by the postings iterator (which is to be excluded).
+			isMatch = !contains(it.postings.Clone(), curPostings)
```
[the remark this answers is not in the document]
dimitarvdimitrov: i realized that's not true. We still short-circuit on the first posting which isn't in `it.postings`

PR #577, opened 2023-12-20
Fix FastRegexMatcher to skip nested capture groups

pracucci:
`FastRegexMatcher` ignores capture group by design. This is done calling `clearCapture()`. However, `clearCapture()` removes a single capture group. This means that when a regexp has nested capture groups (e.g. `((a|b))`) the optimization doesn't trigger because we remove the 1st capture group but not the 2nd (nested) one.

This PR fixes it. To clarify, this is not a logic bug, but we're missing the opportunity to optimize regexp with nested capture groups.


bboreham (approved):
lgtm

PR #463, opened 2023-03-29
Optimize long alternate lists

Logiraptor:
We've seen that very long regex matchers of the form `abc|def|...` consume a large amount of CPU both in `NewFastRegexMatcher` and while matching label values. This PR uses a slice or map to optimize both operations as a special case when the regex is one long alternate of literal values. In that special case, the regex parsing is skipped entirely. As an example, this can occur when using Grafana dashboard template variables with the `All` option selected.

Unfortunately, this introduces a constant performance cost in `NewFastRegexMatcher` for other regex queries, but It's not clear that this is actually significant in practice. For example, cpu for a simple case:

```
NewFastRegexMatcher/^foo-10                                2.396µ ± 2%   2.634µ ± 2%   +9.94% (p=0.002 n=6)
```

This is a 10% increase, but in concrete terms less than 0.3µs on my laptop. 

Here are the results for the specific edge case this PR addresses:

```
goos: darwin
goarch: arm64
pkg: github.com/prometheus/prometheus/model/labels
                                                        │    before     │                after                │
                                                        │    sec/op     │    sec/op     vs base               │
NewFastRegexMatcher/foo-10                                1978.00n ± 2%   82.61n ±  6%  -95.82% (p=0.002 n=6)
NewFastRegexMatcher/zQPbMkNO|NNSPdvMi|iWuuSoAl|qbvKM-10   139.687µ ± 2%   5.787µ ±  1%  -95.86% (p=0.002 n=6)
NewFastRegexMatcher/jyyfj00j0061|jyyfj00j0062|jyyfj9-10    5726.0µ ± 1%   132.0µ ±  4%  -97.70% (p=0.002 n=6)
FastRegexMatcher/zQPbMkNO|NNSPdvMi|iWuuSoAl|qbvKM-10        740.5n ± 2%   204.2n ±  6%  -72.42% (p=0.002 n=6)
FastRegexMatcher/jyyfj00j0061|jyyfj00j0062|jyyfj9-10       3033.5n ± 1%   217.5n ±  6%  -92.83% (p=0.002 n=6)
geomean                                                     925.9n        749.7n        -19.03%

                                                        │      before      │                  after                   │
                                                        │       B/op       │     B/op      vs base                    │
NewFastRegexMatcher/foo-10                                   3240.0 ± 0%       176.0 ± 0%   -94.57% (p=0.002 n=6)
[...]

PR #578, opened 2024-01-02
Fix regexp set matches for literal matchers

Logiraptor:
While testing a feature for Adaptive Metrics, I was surprised to find out that posting lookups for `foo=~"(bar|bar)"` are faster than `foo=~"bar"`. It turns out I introduced a performance regression in https://github.com/grafana/mimir-prometheus/pull/463.

When we added the `optimizeAlternatingLiterals` function, we subtly broke one edge case. A regexp matcher which matches a single literal, like `foo=~"bar"` used to return `bar` from `SetMatches()`, but currently does not. The implication is that the tsdb will first do a LabelValues call to get all values for `foo`, then match them against the regexp `bar`. This PR restores the previous behavior which is able to directly lookup postings for `foo="bar"` instead. In a random sample of 1M queries from a large dedicated cell, about 11% of queries had at least one matcher in this form.

I've added a new benchmark to show the difference for a label with high cardinality:

```
goos: darwin
goarch: arm64
                                                    │   before.txt    │              after.txt              │
                                                    │     sec/op      │   sec/op     vs base                │
Querier/Head/PostingsForMatchers/i_times_n=~"0"-10    7136391.5n ± 4%   362.0n ± 1%  -99.99% (p=0.000 n=10)
Querier/Block/PostingsForMatchers/i_times_n=~"0"-10     7578.47µ ± 0%   49.29µ ± 1%  -99.35% (p=0.000 n=10)
geomean                                                   7.354m        4.224µ       -99.94%

                                                    │   before.txt    │              after.txt              │
                                                    │      B/op       │    B/op     vs base                 │
Querier/Head/PostingsForMatchers/i_times_n=~"0"-10    7372864.00 ± 0%   64.00 ± 0%  -100.00% (p=0.000 n=10)
Querier/Block/PostingsForMatchers/i_times_n=~"0"-10   7372896.00 ± 0%   80.00 ± 0%  -100.00% (p=0.000 n=10)
geomean                                                  7.031Mi        71.55       -100.00%

                                                    │ before.txt │             after.txt              │
                                                    │ allocs/op  │ allocs/op   vs base                │
Querier/Head/PostingsForMatchers/i_times_n=~"0"-10    4.000 ± 0%   3.000 ± 0%  -25.00% (p=0.000 n=10)
Querier/Block/PostingsForMatchers/i_times_n=~"0"-10   6.000 ± 0%   4.000 ± 0%  -33.33% (p=0.000 n=10)
geomean                                               4.899        3.464       -29.29%

```

### DIFF
--- a/model/labels/regexp.go
+++ b/model/labels/regexp.go
@@ -366,7 +366,7 @@ func optimizeAlternatingLiterals(s string) (StringMatcher, []string) {
 	// If there are no alternates, check if the string is a literal
 	if estimatedAlternates == 1 {
 		if regexp.QuoteMeta(s) == s {
-			return &equalStringMatcher{s: s, caseSensitive: true}, nil
+			return &equalStringMatcher{s: s, caseSensitive: true}, []string{s}
 		}
 		return nil, nil
 	}
--- a/model/labels/regexp_test.go
+++ b/model/labels/regexp_test.go
@@ -342,6 +342,14 @@ func TestFindSetMatches(t *testing.T) {
 			matches, actualCaseSensitive := findSetMatches(parsed)
 			require.Equal(t, c.expMatches, matches)
 			require.Equal(t, c.expCaseSensitive, actualCaseSensitive)
+
+			if c.expCaseSensitive {
+				// When the regexp is case sensitive, we want to ensure that the
+				// set matches are maintained in the final matcher.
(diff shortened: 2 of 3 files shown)


aknuds1 (approved):
LGTM

PR #579, opened 2024-01-03
Add multiple encoding test

fionaliao:
Add a test for when a series has samples with multiple encodings, including when the samples are out of order. Note that the behaviour is a bit weird when multiple encodings are committed together - float samples are always processed before histogram ones so sometimes histograms are marked as OOO when you might not expect them to be - see test comments for more details. I don't think this behaviour is worth changing as it shouldn't happen very often.


carrieedwards (approved):

carrieedwards (approved):
LGTM!

PR #581, opened 2024-01-05
Fix OOO counter reset test

fionaliao:
Test was failing due to ` 814 observations found in buckets, but the Count field is 1000: histogram's observation count should equal the number of observations found in the buckets (in absence of NaN)`. Updated test to generate a valid histogram for each sample rather than just updating the count.


carrieedwards (approved):
LGTM!

PR #580, opened 2024-01-05
Sync upstream Prometheus

pracucci:
In this PR I'm synching upstream Prometheus up until today (commit https://github.com/prometheus/prometheus/commit/6d2e0a74240c5f8cdcd618e4730089bc5c1265c6).

## Notes on the sync

The `git merge` had conflicts on the following files (so pay extra care when reviewing):

- `go.mod`
- `go.sum`
- `rules/manager_test.go`
- `tsdb/compact.go`
- `tsdb/compact_test.go`
- `tsdb/db.go`
- `tsdb/head.go`
- `tsdb/head_test.go`
- `tsdb/querier_test.go`

Other fixing conflicts (mostly related to changes to errors creation, moving from `github.com/pkg/errors` to `errors` in https://github.com/prometheus/prometheus/pull/13155), I had do some some notable changes:

- `AllowOverlappingCompaction` config option was been removed in https://github.com/prometheus/prometheus/pull/13282. I've reintroduced it back because we actually use it in our fork (we need 100% sure we never compact overlapping blocks in ingesters). I will open a PR to propose to re-introduce it back, with the actual logic we use.

Other changes I've done:

- [Fix linter issues](https://github.com/grafana/mimir-prometheus/pull/580/commits/7c8e9a2a76fc729e9078889782928b2fdfe240e9) due to new rules. These changes have been done on code existing in mimir-prometheus but not prometheus.
- [Changed `approxBytesPerEntry` const](https://github.com/grafana/mimir-prometheus/pull/580/commits/397179e39480d2275dea29d002f4c06d2d3296d4) (used by `SeriesHashCache`) because the computed value using golang 1.21 (which I've upgraded to in this PR) is higher than the one we computed with golang 1.20

## Notes on my self review

I've reviewed all changes. I think it's all good. What's worth to note...

- Packages whose changes don't affect Mimir (can skip the review):
  - `cmd/`, `discovery/`, `docs/`, `documentation/`, `scrape/`, `scripts/`, `tsdb/agent/`, `web/ui/`
- Countless changes just to switch from `github.com/pkg/errors` to `errors` (bah...)
- New PromQL functions (we'll have to check whether they can be safely supported by query sharding in Mimir):
  - mad_over_time
  - sort_by_label
  - sort_by_label_desc
- `PostingsCardinalityStats()` got a fix in https://github.com/prometheus/prometheus/pull/12653, but we don't use that function.
[...]

PR #583, opened 2024-01-16
Sync prometheus

pracucci:
This PR is a follow up of https://github.com/grafana/mimir-prometheus/pull/580, synching more commits merged in Prometheus after https://github.com/grafana/mimir-prometheus/pull/580.


tsdb/compact.go
```diff
@@ -170,9 +193,18 @@ func NewLeveledCompactorWithChunkSize(ctx context.Context, r prometheus.Register
 	if l == nil {
 		l = log.NewNopLogger()
 	}
+	mergeFunc := opts.MergeFunc
 	if mergeFunc == nil {
 		mergeFunc = storage.NewCompactingChunkSeriesMerger(storage.ChainedSeriesMerge)
 	}
+	var maxBlockChunkSegmentSize int64
```
pracucci, line 200: This is buggy

tsdb/compact.go
```diff
@@ -170,9 +193,18 @@ func NewLeveledCompactorWithChunkSize(ctx context.Context, r prometheus.Register
...
+	if opts.MaxBlockChunkSegmentSize == 0 {
+		maxBlockChunkSegmentSize = chunks.DefaultChunkSegmentSize
+	}
+	var pe index.PostingsEncoder
```
pracucci, line 204: This is buggy

tsdb/db.go
```diff
@@ -505,9 +505,7 @@ func (db *DBReadOnly) FlushWAL(dir string) (returnErr error) {
 		nil,
 		db.logger,
 		ExponentialBlockRanges(DefaultOptions().MinBlockDuration, 3, 5),
-		chunkenc.NewPool(),
-		nil,
-		false,
```
pracucci, line 510: This was false

tsdb/compact.go
```diff
@@ -181,8 +213,9 @@ func NewLeveledCompactorWithChunkSize(ctx context.Context, r prometheus.Register
 		ctx:                         ctx,
 		maxBlockChunkSegmentSize:    maxBlockChunkSegmentSize,
 		mergeFunc:                   mergeFunc,
+		postingsEncoder:             pe,
+		enableOverlappingCompaction: opts.EnableOverlappingCompaction,
```
krajorama, line 217: nice touch to reduce diff :+1:

tsdb/db.go
krajorama: Doesn't this (removal of setting enable overlapping blocks = false) mean that compaction works slightly differently? I cannot wrap my head around what this means on the high level. https://github.com/grafana/mimir-prometheus/blob/fc40f7cff852bf3946a6c84869539f52fe9e371e/tsdb/compact.go#L279

pracucci: Yes, it's different, but this function is `DBReadOnly.FlushWAL()` and in Mimir we never use `DBReadOnly`, so it doesn't affect us.

krajorama: ah, ok, now I see, in my head the option change and DBReadOnly was separate issues, approving

krajorama (approved):
LGTM

pracucci:
I'm going to merge it to move forward with my work, but if anyone else has any post-merge comment please ping me directly on Slack and I will address them immediately. Thanks!

PR #584, opened 2024-01-16
Add conditional read after write for rules

pracucci:
In this PR:

- Cherry pick https://github.com/prometheus/prometheus/pull/12946
- Add `RuleDetail.Independent`
  - To get it done, I've split the `RuleConcurrencyController` into `RuleConcurrencyController` and `RuleDependencyController`

### DIFF
--- a/rules/origin.go
+++ b/rules/origin.go
@@ -28,6 +28,10 @@ type RuleDetail struct {
 	Query  string
 	Labels labels.Labels
 	Kind   string
+
+	// Independent holds whether this rule depends on the result of other rules
+	// within the same rule group or not.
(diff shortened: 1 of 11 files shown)


rules/origin.go
```diff
@@ -28,6 +28,10 @@ type RuleDetail struct {
 	Query  string
 	Labels labels.Labels
 	Kind   string
+
+	// Independent holds whether this rule depends on the result of other rules
+	// within the same rule group or not.
```
dimitarvdimitrov, line 33: this forgets to mention that if other rules depend on this rule `Independent` is also true

PR #587, opened 2024-01-19
rules.Manager: check whether a Rule has dependents or dependencies and expose it through RuleDetail

pracucci:
This is a simplified version of https://github.com/grafana/mimir-prometheus/pull/584 which doesn't include the support to concurrently evaluate rules, but just analyse whether a Rule has dependents or dependencies and expose the info through RuleDetail. The content of `rules/dependency.go` and `rules/dependency_test.go` is a copy-paste from [Danny's PR](https://github.com/prometheus/prometheus/pull/12946).

Notes:
- The changes done in this PR have been designed to be as much isolated as possible from Prometheus code, in order to reduce the future conflicts when synching.
- Why did I call the fields / functions "**no** dependent/cy rules" (negative) instead of a positive "dependent/cy rules"? Reason is that I wanted the zero value to be the safest option (so zero value means "yes, there are dependant/cy"). I don't feel strong about this tho.

### DIFF
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -327,7 +327,7 @@ func (r *AlertingRule) SetNoDependentRules(noDependentRules bool) {
 	r.noDependentRules.Store(noDependentRules)
 }
 
-func (r *AlertingRule) GetNoDependentRules() bool {
+func (r *AlertingRule) NoDependentRules() bool {
 	return r.noDependentRules.Load()
 }
 
@@ -335,7 +335,7 @@ func (r *AlertingRule) SetNoDependencyRules(noDependencyRules bool) {
 	r.noDependencyRules.Store(noDependencyRules)
 }
 
-func (r *AlertingRule) GetNoDependencyRules() bool {
+func (r *AlertingRule) NoDependencyRules() bool {
 	return r.noDependencyRules.Load()
 }
 
--- a/rules/alerting.go
+++ b/rules/alerting.go
@@ -142,6 +142,9 @@ type AlertingRule struct {
 	active map[uint64]*Alert
 
 	logger log.Logger
+
+	noDependentRules  *atomic.Bool
+	noDependencyRules *atomic.Bool
 }
 
 // NewAlertingRule constructs a new AlertingRule.
@@ -168,6 +171,9 @@ func NewAlertingRule(
 		evaluationTimestamp: atomic.NewTime(time.Time{}),
 		evaluationDuration:  atomic.NewDuration(0),
 		lastError:           atomic.NewError(nil),
+
+		noDependentRules:  atomic.NewBool(false),
+		noDependencyRules: atomic.NewBool(false),
 	}
 }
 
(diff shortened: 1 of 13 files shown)


rules/alerting_test.go
```diff
@@ -920,3 +920,45 @@ func TestAlertingEvalWithOrigin(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, detail, NewRuleDetail(rule))
 }
+
+func TestAlertingRule_SetNoDependentRules(t *testing.T) {
```
dimitarvdimitrov, line 924: nit: I'm not 100% convinced we need _that_ granular tests (same for `TestRecordingRule_SetNo*`

rules/alerting.go
```diff
@@ -142,6 +142,9 @@ type AlertingRule struct {
 	active map[uint64]*Alert
 
 	logger log.Logger
+
+	noDependentRules  *atomic.Bool
```
dimitarvdimitrov, line 146: wouldn't this always be `true`? In Danny's PR it says that nothing can depend on an alerting rule

rules/rule.go
```diff
@@ -62,4 +62,12 @@ type Rule interface {
...
 	// NOTE: Used dynamically by rules.html template.
 	GetEvaluationTimestamp() time.Time
+
+	// SetNoDependentRules sets whether there's no other rule in the rule group that depends on this rule.
+	SetNoDependentRules(bool)
+	GetNoDependentRules() bool
+
+	// SetNoDependencyRules sets whether this rule doesn't depend on the output of any rule in the rule group.
+	SetNoDependencyRules(bool)
+	GetNoDependencyRules() bool
```
dimitarvdimitrov, line 72: maybe clarify that if these return `true`, then the result is guaranteed. But if they return `false`, then there still may be no rules which depend on this rule/on which this rule depends.

dimitarvdimitrov (approved):
these changes make more sense for Mimir. LGTM

rules/alerting.go
pracucci: Yes, but I don't think it's `AlertingRule` responsibility to decide it. It's the "dependency controller" to decide it and the `AlertingRule` (which simply implements the `Rule` interface) will obey that what it gets told.

rules/rule.go
pracucci: Done in https://github.com/grafana/mimir-prometheus/pull/587/commits/6b67569e97cdc1d8d66aa340b2f2fedda9582b54

rules/alerting.go
```diff
@@ -317,6 +323,22 @@ func (r *AlertingRule) Restored() bool {
 	return r.restored.Load()
 }
 
+func (r *AlertingRule) SetNoDependentRules(noDependentRules bool) {
+	r.noDependentRules.Store(noDependentRules)
+}
+
+func (r *AlertingRule) GetNoDependentRules() bool {
+	return r.noDependentRules.Load()
+}
```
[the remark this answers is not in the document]
pracucci: Done in https://github.com/grafana/mimir-prometheus/pull/587/commits/d3e31231a30455fb2c9fbfffac8fbb4a4726e871

rules/alerting_test.go
pracucci: I skip this nit. Given the similar name between "no dependent" and "no dependency" it helped me to make sure I wired the setters and getters correctly.

PR #590, opened 2024-01-23
fix: increase function results do not reflect normal trends when ther…

shichanglin5:
**When there is out-of-order data (counter types of time series may occasionally have non-incremental data), the results returned by using increase queries do not reflect the actual trend.**

**The time series is directly queried here, as follows：**
image

**When queried using the increase function, the result is the following:**
image

**I call the query_range api directly, and the query finds that the sample value at a certain point in time is smaller than the previous point in time.**
image

**After modifying the relevant calculation logic of functions.go, the query results are in line with expectations.**
image


bboreham:
The behaviour appears to be in line with the definition of `increase`, `rate`, etc. - values must be monotonitcally increasing.

bboreham:
I'm going to close this since it's not something we would take as a difference from upstream.

PR #597, opened 2024-02-08
Add timely head compaction support

jhalterman:
This PR introduces a new TSDB and head option: `TimelyCompaction`. This option, which defaults to `false`, allows compaction to happen if the min block range is no longer appendable, without requiring the head to exceed 1.5 times the chunk range, as the current behavior requires. When this option is not enabled, the current behavior applies. 

The goal of this setting is to allow compactions which might normally occur at a certain time, such as because of a previous early compaction, to still occur. For example: Consider a head with:

- minT=1:55
- maxT=3:01

If head compaction is attempted at 3:01, the current logic will prevent it since the head does not contain 1.5 times the block size. With `TimelyCompaction` enabled, the min-compactable block, which is from 1:00-2:00, is compactable since it ends more than .5 blocks (1h) prior to the maxT, so compaction will be allowed.

Partially addresses https://github.com/grafana/mimir/issues/7306

### Testing

I ran this for a day with heavy use of early compaction, and did not observe any blocks that were uploaded late, as they would be prior to this change.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -211,6 +211,10 @@ type Options struct {
 	// EnableSharding enables query sharding support in TSDB.
 	EnableSharding bool
 
+	// Timely compaction allows head compaction to happen if the min block in the head is appendable, without requiring
+	// the head to exceed 1.5 times the chunk range.
+	TimelyCompaction bool
+
 	// HeadPostingsForMatchersCacheTTL is the TTL of the postings for matchers cache in the Head.
 	// If it's 0, the cache will only deduplicate in-flight requests, deleting the results once the first request has finished.
 	HeadPostingsForMatchersCacheTTL time.Duration
@@ -939,6 +943,7 @@ func open(dir string, l log.Logger, r prometheus.Registerer, opts *Options, rngs
 	headOpts.OutOfOrderTimeWindow.Store(opts.OutOfOrderTimeWindow)
 	headOpts.OutOfOrderCapMax.Store(opts.OutOfOrderCapMax)
 	headOpts.EnableSharding = opts.EnableSharding
+	headOpts.TimelyCompaction = opts.TimelyCompaction
 	headOpts.PostingsForMatchersCacheTTL = opts.HeadPostingsForMatchersCacheTTL
 	headOpts.PostingsForMatchersCacheMaxItems = opts.HeadPostingsForMatchersCacheMaxItems
 	headOpts.PostingsForMatchersCacheMaxBytes = opts.HeadPostingsForMatchersCacheMaxBytes
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -211,8 +211,8 @@ type Options struct {
 	// EnableSharding enables query sharding support in TSDB.
 	EnableSharding bool
 
-	// Timely compaction allows head compaction to happen if the min block in the head is compactable, without requiring
-	// the head to exceed 1.5 times the chunk range.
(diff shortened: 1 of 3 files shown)


tsdb/db.go
```diff
@@ -211,6 +211,10 @@ type Options struct {
 	// EnableSharding enables query sharding support in TSDB.
 	EnableSharding bool
 
+	// Timely compaction allows head compaction to happen if the min block in the head is appendable, without requiring
+	// the head to exceed 1.5 times the chunk range.
+	TimelyCompaction bool
```
jhalterman, line 216: As ever, naming is hard. I'm open to alternative suggestions.

tsdb/db.go
```diff
@@ -211,6 +211,10 @@ type Options struct {
 	// EnableSharding enables query sharding support in TSDB.
 	EnableSharding bool
 
+	// Timely compaction allows head compaction to happen if the min block in the head is appendable, without requiring
```
pstibrany, line 214: "min block in the head is appendable" -- I don't quite follow this description. Shouldn't it be the other way around, ie. "min block in the head is no longer appendable"?

jhalterman: It's a good point. I updated the description. The description on `compactable()` is more specific, but the description on the settings just says "when the min block is compactable". The meaning of compactable seems a bit too low level to put in those descriptions, and later, into a flag in Mimir.

jhalterman: Does the possibility for OOO samples make that somewhat confusing, where we can still append those to the (OOO) head? I think it's more about how far in the "past" the min block is.

pstibrany: But this flag only applies to "normal" (non-OOO) Head, right? We can be explicit about it.

jhalterman: Updated similar to the suggestion.

pstibrany (approved):
Love how tiny the code change is. I wonder if Prometheus would be interested in this too, since it results in more predictable compaction times.

pracucci (approved):
LGTM, but I would be glad to hear @jesusvazquez or @bboreham opinion. Not a blocker to get you testing this out in dev, but please ping them for a feedback. I want to make sure we're not missing anything.

bboreham:
No objections to trying this out.

However, I think the change should be made without an option; I think the current behaviour is wrong.

I looked back at the history of this code:
* The 1.5x calculation was added very early, in this commit: https://github.com/prometheus-junkyard/tsdb/commit/3901b6e70bcb4e93152de79be86a183fc0ec135a#diff-ac94c91f8ea84762d55b6932ef1466a2165984591d3e646360c5bb31b2130900R343
  * The comment says "The 0.5 acts as a buffer of the appendable window."
* That code was factored out in https://github.com/prometheus-junkyard/tsdb/pull/542
* `appendableMinValidTime()` was added a year later in https://github.com/prometheus/prometheus/pull/8212
  * This also adds 0.5 of a block range, but to the current max time rather than the start time plus one block.

I think `compactable()` could have been modified to use `appendableMinValidTime()` as you do in this PR at that point.

pstibrany:
Personally I'd keep the flag for now, merge it and then enable on our prod cells. Once we know it's working as expected and don't see problems, we can remove the flag and switch behaviour.

Flag allows us to quickly revert to old behaviour by simple restart, instead of rebuilding everything.

jhalterman:
Confirmed Bryan is ok to merge (Jesus is away).

PR #604, opened 2024-03-11
merge upstream at b4d4dcd9f6b5d76a72d71c4a5db0681538c3b070

krajorama:
Most important according to @krajorama :
https://github.com/prometheus/prometheus/pull/13725
https://github.com/prometheus/prometheus/pull/13681


cmd/prometheus/main.go
```diff
@@ -99,7 +100,7 @@ var (
 )
 
 func init() {
-	prometheus.MustRegister(version.NewCollector(strings.ReplaceAll(appName, "-", "_")))
+	prometheus.MustRegister(versioncollector.NewCollector(strings.ReplaceAll(appName, "-", "_")))
```
pstibrany, line 103: This may affect Mimir and GEM too.

tsdb/compact.go
```diff
@@ -121,9 +121,12 @@ func newCompactorMetrics(r prometheus.Registerer) *CompactorMetrics {
...
 	m.Duration = prometheus.NewHistogram(prometheus.HistogramOpts{
-		Name:    "prometheus_tsdb_compaction_duration_seconds",
-		Help:    "Duration of compaction runs",
-		Buckets: prometheus.ExponentialBuckets(1, 2, 14),
+		Name:                            "prometheus_tsdb_compaction_duration_seconds",
+		Help:                            "Duration of compaction runs",
+		Buckets:                         prometheus.ExponentialBuckets(1, 2, 14),
+		NativeHistogramBucketFactor:     1.1,
+		NativeHistogramMaxBucketNumber:  100,
+		NativeHistogramMinResetDuration: 1 * time.Hour,
```
pstibrany, line 129: Mimir exports sum of these histograms as a metric: https://github.com/grafana/mimir/blob/6f421c77000846aafb299a7cf1dad5ec4a1c259b/pkg/ingester/metrics.go#L753

I wonder if it can handle native histograms too. (I realize that we still produce "classic" histogram here.)

pstibrany (approved):

tsdb/compact.go
krajorama: no it cannot, I'll have to fix that (shouldn't break on the other hand)

cmd/prometheus/main.go
krajorama: not sure what you mean, but I'll keep an eye out

pstibrany: Mimir and GEM both use `version` package from Prometheus. We may need to switch to new `versioncollector` package.

PR #605, opened 2024-03-12
TSDB: Don't compact the head block when empty

56quarters:
Don't compact the Head block if there have not yet been any samples appended. Adds methods for checking if the minTime or maxTime haven't yet been set instead of relying on comparisons to default values in several different places.


tsdb/head.go
```diff
@@ -1665,10 +1675,15 @@ func (h *Head) MaxOOOTime() int64 {
 // Else the head has a compactable range when the head time range is 1.5 times the chunk range.
 // The 0.5 acts as a buffer of the appendable window.
 func (h *Head) compactable() bool {
+	if h.IsMinTimeUnset() || h.IsMaxTimeUnset() {
```
jhalterman, line 1678: Do we need to check maxTime?

tsdb/head.go
```diff
@@ -1643,11 +1643,21 @@ func (h *Head) MinTime() int64 {
 	return h.minTime.Load()
 }
 
+// IsMinTimeUnset returns true if the head does not yet have a MinTime set, false otherwise.
+func (h *Head) IsMinTimeUnset() bool {
```
jhalterman, line 1647: Looking at the comment on `AppendableMinValidTime`, maybe we could call this `isInitialized`?

tsdb/head.go
56quarters: I'd rather check both to ensure both have valid values, even if they're updated together today. We do math with both of their values below and there's nothing code or type-system wise preventing people from introducing a bug that causes only one to be updated.

tsdb/head.go
56quarters: Sure, I'd be OK with combining the min and max checks into a single method called something like that.

56quarters:
I'm going to close this PR in favor of the upstream one since this seems like something that should be fixed upstream and is not unique to our timely compaction logic.

pstibrany:
Upstream PR: https://github.com/prometheus/prometheus/pull/13755

PR #607, opened 2024-03-14
Merge upstream prometheus/prometheus at d0ecf9b

charleskorn:
This includes the following changes, none of which seem to be worthy of concern for Mimir:

Performance improvements:
* https://github.com/prometheus/prometheus/pull/13742: performance improvement to chunk checksum checks
* https://github.com/prometheus/prometheus/pull/13673: performance improvement to `Labels.String()`

Bug fixes:
* https://github.com/prometheus/prometheus/pull/13667: bug fixes for query statistics when native histograms are queried
* https://github.com/prometheus/prometheus/pull/13777
* https://github.com/prometheus/prometheus/pull/9633

Refactoring and tidy up:
* https://github.com/prometheus/prometheus/pull/13713: introduce `promql.QueryEngine` interface  and allow using it for rule evaluation
* https://github.com/prometheus/prometheus/pull/13763
* https://github.com/prometheus/prometheus/pull/13764

Changes that don't affect Mimir:
* https://github.com/prometheus/prometheus/pull/13772
* https://github.com/prometheus/prometheus/pull/12933
* https://github.com/prometheus/prometheus/pull/13099

Non-production changes (tests, comments, docs etc.):
* https://github.com/prometheus/prometheus/pull/13751: changes some tests to use `http.StatusXXX` constants instead of literals
* https://github.com/prometheus/prometheus/pull/13743
* https://github.com/prometheus/prometheus/pull/13728: upgrades prometheus/common in an example app
* https://github.com/prometheus/prometheus/pull/13752, https://github.com/prometheus/prometheus/pull/13770, https://github.com/prometheus/prometheus/pull/13768, https://github.com/prometheus/prometheus/pull/13783: add automation to publish container image readme to Docker Hub and Quay.io
* https://github.com/prometheus/prometheus/pull/13759: fix import path comments
* https://github.com/prometheus/prometheus/pull/13771


.github/workflows/container_description.yml
```diff
@@ -0,0 +1,55 @@
+---
+name: Push README to Docker Hub
```
aknuds1, line 2: Do we want this workflow in the fork? This would be the only workflow shared with upstream. AFAICT, you make this workflow essentially no-op, by checking that the repository is prometheus/prometheus. Why not follow the pattern of the other workflows for consistency, and leave them out?

aknuds1 (commented):
Generally LGTM, just wondering about why this new GitHub workflow is included (while we don't share any other workflows with upstream).

.github/workflows/container_description.yml
charleskorn: We don't want this workflow in any fork - https://github.com/prometheus/prometheus/pull/13768 will disable it for forks.

To me, it seems easier to take the workflow file as-is including the condition so that it won't run in our fork, rather than removing the workflow altogether. Then we don't have to worry about conflicts etc. in the future.

If you're OK with this approach, I'll wait until that PR is merged and then update this PR to include all changes up to and including it.

aknuds1: Thanks for the explanation, sounds good to me @charleskorn.

aknuds1 (approved):
LGTM, thanks!

PR #615, opened 2024-04-29
Remove OpEmptyMatch from regex concatenations

colega:
OpEmptyMatch is regex's nop operation, but it breaks some optimizations which aren't designed for a concatenation of multiple operations.

I saw a `pod=~"().*"` matcher in the wild and wondered whether it would be optimized, and was suprised to see that it wasn't.

### DIFF
--- a/model/labels/regexp.go
+++ b/model/labels/regexp.go
@@ -291,18 +291,12 @@ func clearCapture(regs ...*syntax.Regexp) {
 	}
 }
 
-// clearCapture removes capture operation as they are not used for matching.
+// removeEmptyMatches returns the slice with syntax.OpEmptyMatch regexps removed.
+// Note: it modifies the input slice (the returned slice is a sub-slice of the input).
 func removeEmptyMatches(regs []*syntax.Regexp) []*syntax.Regexp {
-	for i := 0; i < len(regs); i++ {
-		if regs[i].Op == syntax.OpEmptyMatch {
-			regs = slices.Delete(regs, i, i+1)
-		}
-		// Don't remove the last one.
-		if len(regs) == 1 {
-			return regs
-		}
-	}
-	return regs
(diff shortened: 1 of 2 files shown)


model/labels/regexp.go
```diff
@@ -291,6 +291,20 @@ func clearCapture(regs ...*syntax.Regexp) {
 	}
 }
 
+// clearCapture removes capture operation as they are not used for matching.
```
pracucci, line 294: Looks being copied from the previous function. Can you update it, please?

pracucci (approved):
LGTM (modulo a couple of minor comments)

model/labels/regexp.go
```diff
@@ -291,6 +291,20 @@ func clearCapture(regs ...*syntax.Regexp) {
...
+func removeEmptyMatches(regs []*syntax.Regexp) []*syntax.Regexp {
+	for i := 0; i < len(regs); i++ {
+		if regs[i].Op == syntax.OpEmptyMatch {
+			regs = slices.Delete(regs, i, i+1)
+		}
+		// Don't remove the last one.
```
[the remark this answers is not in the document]
colega: I added testcase for `()` (https://github.com/grafana/mimir-prometheus/pull/615/commits/c68bcaf36ce8a8d72d3c7f899f87c7bcd88d61b7) and I also removed the check to keep one last element completely, it didn't make sense as we're checking for empty `re.Sub` right after calling it.

model/labels/regexp.go
colega: Updated in https://github.com/grafana/mimir-prometheus/pull/615/commits/513a847634c9f12ecd5011e08ae4f5adb5993ced

colega:
Thank you, @pracucci, do you want to take another look?

pracucci (approved):
LGTM!

PR #617, opened 2024-04-30
Use matrix Github job to test all buildtags

colega:
Instead of defining a new job each time, we can just use a matrix. Useful since now we have the dedupelabels build aswell.

This also adds a test job for `dedupelabels` tag, which wasn't present before.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         buildtags:
-          - ""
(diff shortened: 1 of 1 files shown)


pstibrany (approved):
nice. (I'd call it something like `default` instead of `slicelabels`, but it doesn't matter).

colega:
I think `default` would be misleading as the _default_ thing we run in production is `stringlabels` now.

colega:
I see, I've edited the branch protection rules to require the `test (stringlabels)` job instead of the plain `test` job.

PR #618, opened 2024-04-30
Use setup-go 

colega:
Using the pre-defined action should be faster than installing golang from scrach each time.


colega:
Removed the golang upgrade from this PR, as it causes some tests to fail and I'll address that separately.

aknuds1 (approved):

PR #621, opened 2024-04-30
GitHub Workflows: Upgrade to checkout v4.1.4

aknuds1:
Upgrade to actions/checkout v4.1.4 in our GitHub workflows. I noticed we were using inconsistent versions of it (v2, v3, v4), so why not upgrade to the latest.


pstibrany (approved):
shall we just use `@v4`?

colega:
Yes, please. Depending on a commit here feels weird.

PR #626, opened 2024-05-07
Update upstream prometheus to: fix `parser.VectorSelector.String()` with empty name matcher (#14015)

colega:
This brings the only new commit there's in upstream Prometheus:

> Fix `parser.VectorSelector.String()` with empty name matcher (#14015)


pracucci (approved):
Remember to merge with merge commit, thanks.

PR #633, opened 2024-05-13
Check for context cancel before regex matching postings

krajorama:
Regex matching can be heavy if the regex takes a lot of cycles to evaluate.

### DIFF
--- a/tsdb/index/index.go
+++ b/tsdb/index/index.go
@@ -1816,7 +1816,12 @@ func (r *Reader) postingsForLabelMatchingV1(ctx context.Context, name string, ma
 	}
 
 	var its []Postings
+	count := 1
 	for val, offset := range e {
+		if count%1000 == 0 && ctx.Err() != nil {
+			return ErrPostings(ctx.Err())
+		}
+		count++
 		if !match(val) {
 			continue
 		}
--- a/tsdb/index/index.go
+++ b/tsdb/index/index.go
@@ -1818,7 +1818,7 @@ func (r *Reader) postingsForLabelMatchingV1(ctx context.Context, name string, ma
 	var its []Postings
 	count := 1
 	for val, offset := range e {
-		if count%1000 == 0 && ctx.Err() != nil {
+		if count%100 == 0 && ctx.Err() != nil {
 			return ErrPostings(ctx.Err())
 		}
 		count++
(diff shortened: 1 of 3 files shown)


aknuds1:
Having a look. Should this PR not be upstreamed though?

aknuds1 (commented):
Generally looks good to me, but I don't think the context should be checked in the callback to `r.traversePostingOffsets`.

tsdb/index/index.go
```diff
@@ -1791,7 +1791,12 @@ func (r *Reader) PostingsForLabelMatching(ctx context.Context, name string, matc
 
 	lastVal := e[len(e)-1].value
 	var its []Postings
+	count := 1
 	if err := r.traversePostingOffsets(ctx, e[0].off, func(val string, postingsOff uint64) (bool, error) {
+		if count%1000 == 0 && ctx.Err() != nil {
```
pracucci, line 1796: I think it's already covered by `traversePostingOffsets()`. The context is checked for every iteration there, isn't it?

krajorama: yep, got to same conclusion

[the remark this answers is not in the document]
krajorama: yes

pracucci (commented):
I would also do it in `labelValuesWithMatchers()`.

pracucci (approved):
LGTM (modulo doing it in `labelValuesWithMatchers()` too).

aknuds1 (approved):
Looks good from my side.

pracucci:
On slack we agreed to change 1000 to 100.

bboreham (approved):
lgtm

56quarters:
All review concerns appear to be addressed, merging

tsdb/index/index.go
```diff
@@ -1816,7 +1816,12 @@ func (r *Reader) postingsForLabelMatchingV1(ctx context.Context, name string, ma
 	}
 
 	var its []Postings
+	count := 1
 	for val, offset := range e {
+		if count%100 == 0 && ctx.Err() != nil {
```
colega, line 1821: May I send some late feedback here?
First: I feel like `100` is too often. How often do we want to check? I'd say every 10ms or even 100ms would be enough. For 10ms, this would mean that each check takes 100 microseconds, and that sounds like a lot (do we have numbers?)
Second: Can we do `%128` here? We don't really care if it's every 100 or every 128, but it's way cheaper for the CPU to do the latter operation. (I did the benchmarks, and there's a noticeable difference when switching to 128. Interestingly, same difference as checking `count == count>>7<<7` but for some reason, checking `count & (1<<7-1) == 0` is less performant)

PR #636, opened 2024-05-27
Add +~ and *~ operators that keep name

colega:
These are the modified versions of + and * (ADD and MUL) that keep the metric name on the series instead of dropping it.


colega:
Alternative implementation using a modifier: https://github.com/grafana/mimir-prometheus/pull/637

PR #637, opened 2024-05-27
Implement keeping_name modifier

colega:
This implements a modifier for binary operations called `keeping_name`
that prevents the metric name from being dropped.


colega:
Not needed anymore.

PR #639, opened 2024-05-30
Merge upstream prometheus/prometheus at `37b408c`

gotjosh:
Note to reviewers:

Please pay close attention to this diff, in particular to the following files:

```git
Unmerged paths:
  (use "git add/rm <file>..." as appropriate to mark resolution)
        deleted by us:   .github/workflows/ci.yml
        both modified:   model/rulefmt/rulefmt.go
        both modified:   promql/engine_test.go
        both modified:   rules/alerting.go
        both modified:   rules/group.go
        both modified:   rules/manager.go
        both modified:   rules/manager_test.go
        both modified:   rules/recording.go
        both modified:   rules/rule.go
        both modified:   tsdb/index/index_test.go
```

The change that affected the `tsdb/index/index_test.go` and `promql/engine_test.go` is https://github.com/prometheus/prometheus/pull/14071 -- I'll need an in-depth review of those two files and that PR as I modified to the best of my ability with what I understood.

# Relevant

- https://github.com/prometheus/prometheus/pull/14061 (the upstreaming of "rule evaluation delay" which is now called "query offset")
- https://github.com/prometheus/prometheus/pull/14075 (a small tsdb optimisation, not worth of a changelog)
- https://github.com/prometheus/prometheus/pull/14000 (query logger internals, not worth of a changelog)
- https://github.com/prometheus/prometheus/pull/14079 (seems changelog worthy, TSDB experts, help me out?)
- https://github.com/prometheus/prometheus/pull/13218 (maybe, need to double check in Mimir)
- https://github.com/prometheus/prometheus/pull/14118 (optimisation, not worth of a changelog)
- https://github.com/prometheus/prometheus/pull/14071 (test change, because we have non-upstreamed changes in this area)
- https://github.com/prometheus/prometheus/pull/14090 (optimisation, not worth a changelog)
- https://github.com/prometheus/prometheus/pull/14091 (optimisations (?), not worth a changelog)
- https://github.com/prometheus/prometheus/pull/14096 (optimisation, not worth a changelog)
- https://github.com/prometheus/prometheus/pull/14068 (bugfix (?), will ask oleg, not sure if we have enabled support for utf-8 already in Mimir)
- https://github.com/prometheus/prometheus/pull/13669 (optimisation, not worth a changelog)

# Not Relevant
- https://github.com/prometheus/prometheus/pull/13999 (test change)
- https://github.com/prometheus/prometheus/pull/13932 (OTel translator)
- https://github.com/prometheus/prometheus/pull/13758 (doc change)
- https://github.com/prometheus/prometheus/pull/14103 (bugfix for code we never imported)
- https://github.com/prometheus/prometheus/pull/14062 (prometheus mixin)
- https://github.com/prometheus/prometheus/pull/14100 (test change)
- https://github.com/prometheus/prometheus/pull/13970 (service discovery)
- https://github.com/prometheus/prometheus/pull/14121 (struct made public, not relevant but might trigger a code change)
- https://github.com/prometheus/prometheus/pull/13638 (test change)
- https://github.com/prometheus/prometheus/pull/14120 (doc change)
- https://github.com/prometheus/prometheus/pull/14116 (API message when limit is hit, we don't use them in Mimir IIRC)
- https://github.com/prometheus/prometheus/pull/14106 (doc string change)
- https://github.com/prometheus/prometheus/pull/14129 (doc change)
- https://github.com/prometheus/prometheus/pull/12339 (doc change)
- https://github.com/prometheus/prometheus/pull/14130 (dockerfile)
- https://github.com/prometheus/prometheus/pull/14149 (doc change)
- https://github.com/prometheus/prometheus/pull/14154 (linter bump)
- https://github.com/prometheus/prometheus/pull/14153 (linter bump)
- https://github.com/prometheus/prometheus/pull/14148 (new go metrics, not relevant for mimir as they are not imported but we might want to add them if we don't have them?)


bboreham:
> * https://github.com/prometheus/prometheus/pull/14096 (optimisation, not worth a changelog)

This came out of a major incident, so it is quite important, but also it has already been merged here as #633.

tsdb/index/index_test.go
```diff
@@ -764,3 +639,46 @@ func TestReader_PostingsForLabelMatchingHonorsContextCancel(t *testing.T) {
 	require.Error(t, p.Err())
 	require.Equal(t, failAfter, ctx.Count())
 }
+
+// createFileReader creates a temporary index file. It writes the provided input to this file.
+// It returns a Reader for this file, the file's name, and the symbol map.
+func createFileReaderWithOptions(ctx context.Context, tb testing.TB, input indexWriterSeriesSlice, withCache bool) (*Reader, string, map[string]struct{}) {
```
aknuds1, line 645: Can you retain the `createFileReader` function, as a wrapper for `createFileReaderWithOptions`, that passes `false` for `withCache`? That would simplify the code (not having to pass `false` for `withCache`), and be more idiomatic for the reader.

The doc comment is now out of sync in any case, since it refers to `createFileReader`.

tsdb/index/index_test.go
```diff
@@ -764,3 +639,46 @@ func TestReader_PostingsForLabelMatchingHonorsContextCancel(t *testing.T) {
...
+
+	var ir *Reader
+	if withCache {
+		ir, err = NewFileReaderWithOptions(fn, hashcache.NewSeriesHashCache(1024*1024*1024).GetBlockCacheProvider("test"))
+		require.NoError(tb, err)
+	} else {
+		ir, err = NewFileReader(fn)
+		require.NoError(tb, err)
+	}
+
```
aknuds1, line 682: Why don't you retain the following here? 

```
	tb.Cleanup(func() {
		require.NoError(tb, ir.Close())
	})
```

tsdb/index/index_test.go
```diff
@@ -242,18 +219,12 @@ func TestIndexRW_Postings(t *testing.T) {
...
 	for _, cacheEnabled := range []bool{false, true} {
-		t.Run(fmt.Sprintf("ShardedPostings() cache enabled: %v", cacheEnabled), func(t *testing.T) {
-			var cache ReaderCacheProvider
-			if cacheEnabled {
-				cache = hashcache.NewSeriesHashCache(1024 * 1024 * 1024).GetBlockCacheProvider("test")
-			}
-
-			ir, err := NewFileReaderWithOptions(fn, cache)
-			require.NoError(t, err)
+		t.Run(fmt.Sprintf("ShardedPostings() cache enabled: %v", cacheEnabled), func(innerT *testing.T) {
```
aknuds1, line 222: Why do you call it `innerT`? The standard idiom is to just call it `t`, thus shadowing the parent `t` (which is what you want, otherwise failures will be confusing). It's also a change from `main`, which I don't see why we would want.
```suggestion
		t.Run(fmt.Sprintf("ShardedPostings() cache enabled: %v", cacheEnabled), func(t *testing.T) {
```

tsdb/index/index_test.go
```diff
@@ -733,28 +619,17 @@ func TestChunksTimeOrdering(t *testing.T) {
...
-	}
-	require.NoError(t, idx.AddSymbol("__name__"))
-
-	for i := 1; i <= seriesCount; i++ {
-		require.NoError(t, idx.AddSeries(storage.SeriesRef(i), labels.FromStrings("__name__", fmt.Sprintf("%4d", i)),
-			chunks.Meta{Ref: 1, MinTime: 0, MaxTime: 10},
-		))
+	const seriesCount = 1000
+	var input indexWriterSeriesSlice
+	for i := 1; i < seriesCount; i++ {
```
aknuds1, line 624: I think the `<=` should be maintained:
```suggestion
	for i := 1; i <= seriesCount; i++ {
```

I think I might have made a mistake in my upstream PR, I'll try to rectify it in upstream.

aknuds1 (commented):
I think there might be some mistakes in index_test.go, please see comments.

tsdb/index/index_test.go
gotjosh: This fails with `invalid argument` on an specific test: `TestIndexRW_Postings` - I spent a few cycles but I couldn't figure out why it was failing on the first place. Do you have any ideas or do I need to go hunting more deeply?

tsdb/index/index_test.go
gotjosh: It was an attempt to fix the above - let me change it. I thought the shadowing of `t` was conflicting with the call to `t.Cleanup` but it turned out to not to be that problem.

tsdb/index/index_test.go
gotjosh: This feels potentially harmless? I would opt for leaving it as is and then just fixing it by syncing with upstream again.

tsdb/index/index_test.go
gotjosh: Done in https://github.com/grafana/mimir-prometheus/pull/639/commits/86722a6ef758dc731aa158c1792b9f4a10d3e828

tsdb/index/index_test.go
gotjosh: Done in https://github.com/grafana/mimir-prometheus/pull/639/commits/86722a6ef758dc731aa158c1792b9f4a10d3e828

tsdb/index/index_test.go
aknuds1: OK, should be harmless. I just merged my fix upstream.

tsdb/index/index_test.go
aknuds1: Not sure, I'll check tomorrow.

aknuds1: Checking.

aknuds1: I could fix it. The problem is that `ir` is also closed at the end of `TestIndexRW_Postings`. Just remove that line and you're good.

tsdb/index/index_test.go
```diff
@@ -764,3 +648,79 @@ func TestReader_PostingsForLabelMatchingHonorsContextCancel(t *testing.T) {
 	require.Error(t, p.Err())
 	require.Equal(t, failAfter, ctx.Count())
 }
+
+// createFileReader creates a temporary index file. It writes the provided input to this file.
+// It returns a Reader for this file, the file's name, and the symbol map.
+func createFileReader(ctx context.Context, tb testing.TB, input indexWriterSeriesSlice) (*Reader, string, map[string]struct{}) {
```
aknuds1, line 652: Do you mind just refactoring this function as follows (that's what I had in mind during my last review)?

```golang
func createFileReader(ctx context.Context, tb testing.TB, input indexWriterSeriesSlice) (*Reader, string, map[string]struct{}) {
	return createFileReaderWithOptions(ctx, tb, input, false)
}
```

Then you can also avoid conditional logic in the tests that have a `cacheEnabled` variable, and just call `return createFileReaderWithOptions`.

gotjosh: I'm happy to do it, but I don't think that will make it better. if I make `createFileReader` something different to what's in upstream it'll get conflicts _everytime_. The code duplication is basically there to ensure whatever is upstream matches what we have here.

aknuds1: Won't there only be a conflict when `createFileReader` should eventually change upstream? I don't understand why there would be a conflict _every time_. Isn't that better than `createFileReader` and `createFileReaderWithCache` silently drift apart, when the former changes?

gotjosh: Done in https://github.com/grafana/mimir-prometheus/pull/639/commits/06f16518c1957d1f4f06e89398c9bf409c521bb0

aknuds1: Thanks!

aknuds1 (approved):
Looks good from my side.

Maybe you could get input from @bboreham on https://github.com/prometheus/prometheus/pull/14079?

bboreham:
What is your question?

bboreham:
Yes, put it in the changelog.

model/rulefmt/rulefmt.go
```diff
@@ -136,13 +136,48 @@ func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
...
+	EvaluationDelay *model.Duration `yaml:"evaluation_delay,omitempty"`
+	QueryOffset     *model.Duration `yaml:"query_offset,omitempty"`
+
+	Limit                         int        `yaml:"limit,omitempty"`
+	Rules                         []RuleNode `yaml:"rules"`
+	SourceTenants                 []string   `yaml:"source_tenants,omitempty"`
+	AlignEvaluationTimeOnInterval bool       `yaml:"align_evaluation_time_on_interval,omitempty"`
+}
+
+func (g *RuleGroup) UnmarshalYAML(value *yaml.Node) error {
```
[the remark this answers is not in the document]
colega: I like the way you did it, but I'd fail if both fields are found.

model/rulefmt/rulefmt.go
```diff
@@ -136,13 +136,48 @@ func (g *RuleGroups) Validate(node ruleGroups) (errs []error) {
...
+
+func (g *RuleGroup) MarshalYAML() (interface{}, error) {
+	type Plain RuleGroup
+	aux := Plain(*g)
+
+	if aux.EvaluationDelay != nil && g.QueryOffset == nil {
+		aux.QueryOffset = g.EvaluationDelay
+	}
+
+	aux.EvaluationDelay = nil
```
dimitarvdimitrov, line 178: will this break clients that expect eval delay to be there? I'm thinking about grafana specifically

dimitarvdimitrov: old versions of mimirtool might also [expect](https://github.com/grafana/mimir/blob/5dd3c308090530bd3b6dd5f49aa4785897642f95/pkg/mimirtool/rules/rwrulefmt/rulefmt.go#L12-L17) evaluation_delay to be set

model/rulefmt/rulefmt.go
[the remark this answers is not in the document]
dimitarvdimitrov: as for being backwards compatible I think it's enough. 

Maybe just always set them to the same value so that the code can use either (I haven't reviewed the rest of this PR); this also unblocks rollbacks from newer ruler to older ruler.

config/config.go
```diff
@@ -397,6 +398,8 @@ type GlobalConfig struct {
 	ScrapeProtocols []ScrapeProtocol `yaml:"scrape_protocols,omitempty"`
 	// How frequently to evaluate rules by default.
 	EvaluationInterval model.Duration `yaml:"evaluation_interval,omitempty"`
+	// Offset the rule evaluation timestamp of this particular group by the specified duration into the past to ensure the underlying metrics have been received.
+	RuleQueryOffset model.Duration `yaml:"rule_query_offset"`
```
pracucci, line 402: [non blocking for Mimir, since we don't use it, but relevant for Prometheus] Is it intentional that's missing the `omitempty`? If yes, why? If no, should be fixed.

model/rulefmt/rulefmt.go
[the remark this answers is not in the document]
pracucci: > Maybe just always set them to the same value so that the code can use either

I agree on this. Looks safer, even if once we'll upgrade mimir-prometheus in mimir we'll remove the usage of `EvaluationDelay` (but better safe than sorry).

model/rulefmt/rulefmt.go
pracucci: I agree with Dimitar here. We need to keep exporting `evaluation_delay` otherwise old clients break.

rules/manager_test.go
```diff
@@ -192,8 +192,8 @@ func TestAlertingRule(t *testing.T) {
 }
 
 func TestForStateAddSamples(t *testing.T) {
-	for _, evalDelay := range []time.Duration{0, time.Minute} {
-		t.Run(fmt.Sprintf("evalDelay %s", evalDelay.String()), func(t *testing.T) {
+	for _, queryOffset := range []time.Duration{0, time.Minute} {
+		t.Run("queryOffset "+queryOffset.String(), func(t *testing.T) {
```
pracucci, line 196: Am I missing something or this is different than upstream? Checking prometheus `main` I can see:

```suggestion
		t.Run(fmt.Sprintf("queryOffset %s", queryOffset.String()), func(t *testing.T) {
```

rules/manager_test.go
```diff
@@ -353,176 +353,182 @@ func sortAlerts(items []*Alert) {
 }
 
 func TestForStateRestore(t *testing.T) {
-	storage := promqltest.LoadedStorage(t, `
+	for _, queryOffset := range []time.Duration{0, time.Minute} {
+		t.Run("queryOffset "+queryOffset.String(), func(t *testing.T) {
```
pracucci, line 357: Same here. Not sure what I'm missing 🤔 

```suggestion
		t.Run(fmt.Sprintf("queryOffset %s", queryOffset.String()), func(t *testing.T) {
```

rules/manager_test.go
```diff
@@ -792,7 +838,7 @@ func TestUpdate(t *testing.T) {
 	// Change group rules and reload.
 	for i, g := range rgs.Groups {
 		for j, r := range g.Rules {
-			rgs.Groups[i].Rules[j].Expr.SetString(fmt.Sprintf("%s * 0", r.Expr.Value))
```
pracucci, line 795: I don't understand this too. Upstream main looks to be `rgs.Groups[i].Rules[j].Expr.SetString(fmt.Sprintf("%s * 0", r.Expr.Value))`

tsdb/head_read_test.go
```diff
@@ -130,7 +129,7 @@ func TestBoundedChunk(t *testing.T) {
 		},
 	}
 	for _, tc := range tests {
-		t.Run(fmt.Sprintf("name=%s", tc.name), func(t *testing.T) {
```
pracucci, line 133: Upstream man looks to be `t.Run(fmt.Sprintf("name=%s", tc.name), func(t *testing.T) {`

tsdb/ooo_head_read_test.go
```diff
@@ -828,7 +828,7 @@ func TestOOOHeadChunkReader_Chunk(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		t.Run(fmt.Sprintf("name=%s", tc.name), func(t *testing.T) {
+		t.Run("name="+tc.name, func(t *testing.T) {
```
pracucci, line 831: Upstream is `t.Run(fmt.Sprintf("name=%s", tc.name), func(t *testing.T) {`

_Same comment applies below._

rules/manager_test.go
gotjosh: You're right - I fixed all the linter failures blindly and these were one of them (it basically complains of using `fmt.Sprintf` instead of string concatenation when possible).

I took a brief look and I can see that our setup in CI is slightly different from the one in Prometheus. `golangcilint` configuration files are definitely the same but the way they are used is different - I suspect the problem with the CI being skipped/not-check lies in upstream Prometheus and not here. 

EDIT: I figured it out, the upgrade of golangcilint broke this linter: https://github.com/prometheus/prometheus/pull/14215

I can do either of two things:

1. I can fix it in Prometheus and then come back.
2. I can upgrade the version of `golangcilint` here so that it matches prometheus and then I can safely remove the changes.

config/config.go
gotjosh: Great catch! https://github.com/prometheus/prometheus/pull/14216

rules/manager_test.go
pracucci: >I can upgrade the version of golangcilint here so that it matches prometheus and then I can safely remove the changes.

I would do this. We aim to be on par with prometheus as much as possible.

gotjosh: I have:

1. Reverted the linter fixes with https://github.com/grafana/mimir-prometheus/pull/639/commits/3836920aa9d20dfd14cef050346ea23d855c5d3d and https://github.com/grafana/mimir-prometheus/pull/639/commits/2e6cc9b06eeac9822c7f83f01c12240b8474bab1
2. Upgraded `golangcilint` to match Prometheus with https://github.com/grafana/mimir-prometheus/pull/639/commits/9cce125267604599fa927479e23235c471d25241
3. and finally, fixed 4 lints from changes that are not in Prometheus https://github.com/grafana/mimir-prometheus/pull/639/commits/199a070f2bfe78c131d45274706c18a2f152e1b0

config/config.go
gotjosh: The above has been merged, I'll avoid making this PR larger and not bring this change in now - I'll re-update mimir-prometheus in another PR.

pracucci: Sure, LGTM.

promql/query_logger.go
```diff
@@ -83,6 +87,23 @@ func logUnfinishedQueries(filename string, filesize int, logger log.Logger) {
...
 
+type mmapedFile struct {
+	f io.Closer
+	m mmap.MMap
+}
+
+func (f *mmapedFile) Close() error {
+	err := f.m.Unmap()
+	if err != nil {
+		err = fmt.Errorf("mmapedFile: unmapping: %w", err)
```
krajorama, line 98: Note : there was a fix to this later: https://github.com/prometheus/prometheus/pull/13948

krajorama: Actually fix is further down to an existing code path, but line numbers didn't line up

model/rulefmt/rulefmt.go
gotjosh: Good shout, I've made the change to preserve both in https://github.com/grafana/mimir-prometheus/pull/639/commits/a8dec7c316155c3f60955e10dad468f163a71738

model/rulefmt/rulefmt.go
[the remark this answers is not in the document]
gotjosh: I've made the change to preserve both in https://github.com/grafana/mimir-prometheus/pull/639/commits/a8dec7c316155c3f60955e10dad468f163a71738

> I like the way you did it, but I'd fail if both fields are found.

I think failing on both sounds sensible but seems more of a Mimir validation than a Prometheus one. We do have a function to validate rule groups for Mimir at an API level and I think this validation belongs there: https://github.com/grafana/mimir/blob/main/pkg/mimirtool/rules/rules.go#L255

> Maybe just always set them to the same value so that the code can use either (I haven't reviewed the rest of this PR); this also unblocks rollbacks from newer ruler to older ruler.

I'm not sure you meant set them always to the same value if neither of them exist which is the alternative I had in mind but this is what I did:

1. If either `evaluation_delay` or `query_offset` is set, set the other to the same value.
2. If both of them have values (technically not possible with the validation in Mimir) output them as is
3. In mimir, we'll only use `query_offset`

Let me know what you think.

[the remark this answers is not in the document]
krajorama: I think there would be two typical use cases:
1) Always load rules from code. In which case the user will set the delay and we correctly unmarshal into the offset now.
2) GET the rule from API, update and POST. The offset may or may not be returned on the POST, but both should work as expected I think

LGTM

promql/query_logger.go
krajorama: Let's get this PR merged first

krajorama (approved):

model/rulefmt/rulefmt.go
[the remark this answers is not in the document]
dimitarvdimitrov: > 2. If both of them have values (technically not possible with the validation in Mimir) output them as is

this is the bit I had in mind - if they conflict, then just choose one and set both to it. It's more an argument in principle, than solving for a specific problem right now. A group with different offset and delay is an invalid group. Not feeling strongly about it and I don't need to look at it again.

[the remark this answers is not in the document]
dimitarvdimitrov: perhaps there can be some confusion if
1. the tenant gets a rule group which has both evalution_delay and query_offset
2. They set only evaluation_delay in the json/yaml we returned.
3. post back the same json without removing query_offset

if the client only pays attention to evaluation_delay, they won't ever understand why the delay is not being changed in Mimir. If we set them both to the same value, then the client would at least notice that the changes aren't reflected in Mimir. 

Somewhat contrived, but I imagine that's how some javascript clients might work. Don't want to block this further, sorry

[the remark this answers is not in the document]
pracucci: In Mimir, we have to make it work with older clients not supporting `query_offset`. For this reason, we can't stop exporting `evaluation_delay`, but we also want to export `query_offset`. I want a client to be able to successfully upload and exported config. For this reason, I'm not convinced the following is an option:

>> I like the way you did it, but I'd fail if both fields are found.
> I think failing on both sounds sensible but seems more of a Mimir validation than a Prometheus one

I would keep it simple: export both fields, just fail the validation if they're both set **but** to different values.

gotjosh:
Thank you all for the help 🙏

PR #643, opened 2024-06-10
[weekly-r290] `MemPostings.PostingsForLabelMatching()`: don't hold the mutex while matching

colega:
Backporting https://github.com/prometheus/prometheus/pull/14286 to the `weekly-r290` branch here.

### DIFF
--- a/tsdb/index/postings_test.go
+++ b/tsdb/index/postings_test.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/grafana/regexp"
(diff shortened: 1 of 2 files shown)


tsdb/index/postings_test.go
```diff
@@ -999,117 +1001,6 @@ func TestMemPostings_Delete(t *testing.T) {
 	require.Empty(t, expanded, "expected empty postings, got %v", expanded)
 }
 
-func TestPostingsCloner(t *testing.T) {
```
pracucci, line 1002: Why was this deleted?

jesusvazquez, line 1002: Why are you removing this test?

colega: Recovered it rebasing again and force-pushing.

colega: Bad rebase, sorry, recovered it now.

PR #644, opened 2024-06-10
Improve context cancellation handling in PostingsForMatchersCache

pracucci:
The `PostingsForMatchersCache` has a comment that says:

```
// FIXME: do we need to cancel the call to postingsForMatchers if all the callers waiting for the result have
// cancelled their context?
```

During the weekend we learned the hard way that, yes, we should. The reason is that the underlying `PostingsForMatchers()` honor context cancellation, but when it's called through the single flight promise used by the cache, we run it with a new `context.Background()` and so it never gets canceled.

If the execution of the underlying `PostingsForMatchers()` takes a very long time (e.g. due to an insanely complex regexp), then it's undesirable keep it running if the request has been canceled. We prefer to cancel the `PostingsForMatchers()` execution.

This PR addresses it.

### DIFF
--- a/tsdb/postings_for_matchers_cache.go
+++ b/tsdb/postings_for_matchers_cache.go
@@ -83,8 +83,13 @@ type PostingsForMatchersCache struct {
 
 	// timeNow is the time.Now that can be replaced for testing purposes
 	timeNow func() time.Time
+
 	// postingsForMatchers can be replaced for testing purposes
 	postingsForMatchers func(ctx context.Context, ix IndexPostingsReader, ms ...*labels.Matcher) (index.Postings, error)
+
+	// onPromiseExecutionDoneBeforeHook is used for testing purposes. It allows to hook at the
+	// beginning of onPromiseExecutionDone() execution.
+	onPromiseExecutionDoneBeforeHook func()
 
 	tracer trace.Tracer
 	// Preallocated for performance
@@ -159,40 +164,46 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Contex
 		callersCtxTracker: promiseCallersCtxTracker,
 	}
 
-	// Add the caller context to the ones tracked by the new promise. It's important to do it here
-	// so that if the promise will be stored, then the caller's context is already tracked. Otherwise,
-	// if the promise will not be stored (because there's another in-flight promise for the same label
-	// matchers) then it's not a problem, and resources will be released.
-	promise.callersCtxTracker.add(ctx)
+	// Add the caller context to the ones tracked by the new promise.
+	//
+	// It's important to do it here so that if the promise will be stored, then the caller's context is already
+	// tracked. Otherwise, if the promise will not be stored (because there's another in-flight promise for the
+	// same label matchers) then it's not a problem, and resources will be released.
+	//
+	// Skipping the error checking because it can't happen here.
+	_ = promise.callersCtxTracker.add(ctx)
 
 	key := matchersKey(ms)
 
 	if oldPromiseValue, loaded := c.calls.LoadOrStore(key, promise); loaded {
 		// The new promise hasn't been stored because there's already an in-flight promise
 		// for the same label matchers. We should just wait for it.
-		span.AddEvent("using cached postingsForMatchers promise", trace.WithAttributes(
-			attribute.String("cache_key", key),
-		))
-
-		oldPromise := oldPromiseValue.(*postingsForMatcherPromise)
-
-		// Add the caller context to the ones tracked by the old promise (currently in-flight).
-		// It will be a no-op if the tracker is already closed because the promise isn't in-flight,
-		// and it has been loaded from the cache.
-		//
-		// However, there's a race condition here happening when the "loaded" promise
-		// execution was just canceled. It could happen that "loaded" promise execution was just
-		// canceled, its result not cached (because we don't cache results for canceled executions),
-		// but we didn't remove the promise from c.calls yet, so it was loaded anyway.
-		//
-		// In this scenario, this caller will get context.Canceled error, which is bad because it
-		// should have triggered a new execution instead. We believe this race condition to be
-		// rare enough to not worry about it.
-		oldPromise.callersCtxTracker.add(ctx)
 
 		// Release the resources created by the new promise, that will not be used.
 		close(promise.done)
 		promise.callersCtxTracker.close()
+
+		oldPromise := oldPromiseValue.(*postingsForMatcherPromise)
+
+		// Add the caller context to the ones tracked by the old promise (currently in-flight).
+		if err := oldPromise.callersCtxTracker.add(ctx); err != nil && errors.Is(err, errContextsTrackerCanceled{}) {
+			// We've hit a race condition happening when the "loaded" promise execution was just canceled,
+			// but it hasn't been removed from map of calls yet, so the old promise was loaded anyway.
+			//
+			// We expect this race condition to be infrequent. In this case we simply skip the cache and
+			// pass through the execution to the underlying postingsForMatchers().
+			span.AddEvent("looked up in-flight postingsForMatchers promise, but the promise was just canceled due to a race condition: skipping the cache", trace.WithAttributes(
+				attribute.String("cache_key", key),
+			))
+
+			return func(ctx context.Context) (index.Postings, error) {
+				return c.postingsForMatchers(ctx, ix, ms...)
+			}
+		}
+
+		span.AddEvent("using cached postingsForMatchers promise", trace.WithAttributes(
+			attribute.String("cache_key", key),
+		))
 
 		return oldPromise.result
 	}
@@ -283,6 +294,11 @@ func (c *PostingsForMatchersCache) evictHead() {
 func (c *PostingsForMatchersCache) onPromiseExecutionDone(ctx context.Context, key string, ts time.Time, sizeBytes int64, err error) {
 	span := trace.SpanFromContext(ctx)
 
+	// Call the registered hook, if any. It's used only for testing purposes.
+	if c.onPromiseExecutionDoneBeforeHook != nil {
+		c.onPromiseExecutionDoneBeforeHook()
+	}
+
 	// Do not cache if cache is disabled.
 	if c.ttl <= 0 {
 		span.AddEvent("not caching promise result because configured TTL is <= 0")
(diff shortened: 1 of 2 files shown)


tsdb/postings_for_matchers_cache.go
```diff
@@ -123,6 +124,11 @@ func (c *PostingsForMatchersCache) PostingsForMatchers(ctx context.Context, ix I
 type postingsForMatcherPromise struct {
 	done chan struct{}
 
+	// TODO comment
+	callersCtxTracker *contextsTracker
+
+	// The result of the promise is stored either in cloner or err (only of the two is valued).
+	// Do not access these fields until the done channel is closed.
```
colega, line 131: I think you should move `done` to be [a hat](https://dmitri.shuralyov.com/idiomatic-go/entries/2) here.

tsdb/postings_for_matchers_cache.go
```diff
@@ -147,39 +153,62 @@ func (p *postingsForMatcherPromise) result(ctx context.Context) (index.Postings,
...
+		// The new promise hasn't been stored because there's already an in-flight promise
+		// for the same label matchers. We should just wait for it.
 		span.AddEvent("using cached postingsForMatchers promise", trace.WithAttributes(
 			attribute.String("cache_key", key),
 		))
+
+		oldPromise := oldPromiseValue.(*postingsForMatcherPromise)
+
+		// Add the caller context to the ones tracked by the old promise (currently in-flight).
+		// TODO what if it returns false?
```
krajorama: If I understand this correctly, if return is false, we have not connected the state of our context with the state of the existing promise.  So when we execute the "result" promise, it will look at two things:

Our context, which was cancelled or not, which is fine as it's ours.
The "done" channel of the promise which is probably done, so we'll check error and result which also seems fine.

So nothing to be done?

tsdb/postings_for_matchers_cache.go
```diff
@@ -147,39 +154,69 @@ func (p *postingsForMatcherPromise) result(ctx context.Context) (index.Postings,
...
 		))
+
+		oldPromise := oldPromiseValue.(*postingsForMatcherPromise)
+
+		// Add the caller context to the ones tracked by the old promise (currently in-flight).
+		//
+		// There's an infrequent race condition here. We may try to add the caller's context
+		// on a contextsTracker which has already been closed. In that case, if the stored promise
+		// execution successfully completed, then we're good, but if it was canceled than this
+		// caller will get context.Canceled error. We believe the race condition to be so infrequent
```
krajorama, line 185: Why would this caller get context cancel? This only return false if we did nothing, in which case we did not connect this context with the promise's context?

tsdb/postings_for_matchers_cache.go
pracucci: Done in https://github.com/grafana/mimir-prometheus/pull/644/commits/02ef5a673def9e3109cd2839189ec7e2cceeeba0

tsdb/postings_for_matchers_cache.go
pracucci: When we're inside this `if` branch it means we've "loaded" an already existing promise and we're waiting for its result. Since we didn't connected our context (due to the race condition) then our context is not taken in account when checking whether it's safe to cancel the execution context (the execution context was already canceled due the race condition).

At this point there are 2 conditions:
1. The execution succeeded, so all good, we're just returning the result
2. The execution was canceled, so we're returning `context.Canceled` for this caller too, which is bad

krajorama: ah, I see

pracucci: I've rephrased my comment: https://github.com/grafana/mimir-prometheus/pull/644/commits/b4bb990d6d2c3514c07648343b7344620106da2d

tsdb/postings_for_matchers_cache.go
```diff
@@ -147,39 +153,69 @@ func (p *postingsForMatcherPromise) result(ctx context.Context) (index.Postings,
...
 			attribute.String("cache_key", key),
 		))
+
+		oldPromise := oldPromiseValue.(*postingsForMatcherPromise)
+
+		// Add the caller context to the ones tracked by the old promise (currently in-flight).
+		//
+		// There's an infrequent race condition here. We may try to add the caller's context
+		// on a contextsTracker which has already been closed. In that case, if the stored promise
+		// execution successfully completed, then we're good, but if it was canceled than this
```
colega:
```suggestion
		// execution successfully completed, then we're good, but if it was canceled then this
```

tsdb/postings_for_matchers_cache.go
colega: I don't think that returning a `context.Canceled` to a client who didn't cancel the context is a negligible side effect: the client will always mis-interpret this (because it will asume *their* context is canceled) and unexpected stuff will happen (like returning a 499).

I do think that if it's a rare race condition there's no need to build huge workrarounds for this, so I think that we need to explicitly prevent a `context.Canceled` from being stored in `promise.err` below or to return a different error in `result()` if `errors.Is(promise.err, context.Canceled)`. We can return something generic and internal, like `errors.New("promise error canceled, should be retried")` and rely on the frontend's retry mechanisms.

tsdb/postings_for_matchers_cache.go
```diff
@@ -147,39 +153,74 @@ func (p *postingsForMatcherPromise) result(ctx context.Context) (index.Postings,
...
+		// and it has been loaded from the cache.
+		//
+		// However, there's a race condition here happening when the "loaded" promise
+		// execution was just canceled. It could happen that "loaded" promise execution was just
+		// canceled, its result not cached (because we don't cache results for canceled executions),
+		// but we didn't remove the promise from c.calls yet, so it was loaded anyway.
+		//
+		// In this scenario, this caller will get context.Canceled error, which is bad because it
+		// should have triggered a new execution instead. We believe this race condition to be
+		// rare enough to not worry about it.
```
krajorama, line 190: If we're paranoid we can add a counter here. It's very hard to say what happens in reality. The tracker close and tracker add are synchronized on the same lock, so the sequence tracker.close followed by tracker.add would be very likely if we had many queries being added.

colega: We can also just call `c.postingsForMatchersPromise(...)` recursively.

tsdb/postings_for_matchers_cache.go
pracucci: I think I've found an easy to gracefully handle it. Now the code can detect the race conditon. When it happens, the cache is skipped and the underlying PostingsForMatchers() executed: https://github.com/grafana/mimir-prometheus/pull/644/commits/fa2b5a99aa6a32c3f411af7f5a29a900335cbb86

tsdb/postings_for_matchers_cache.go
pracucci: I want to avoid an unlimited recursion. I thought about a limited recursion, with a fallback in case it consistently fails after N tries, but looked an overcomplication. I'm open to feedback tho.

pracucci: > The tracker close and tracker add are synchronized on the same lock, so the sequence tracker.close followed by tracker.add would be very likely if we had many queries being added.

I had a very hard time to reproduce it in a test (I had to introduce a hook to reproduce the race condition). I expect the PostingsForMatchers() to take order of magnitude more time than the locking, so I'm not fully convinced that the lock will make it "likely" to happen.

tsdb/postings_for_matchers_cache.go
```diff
@@ -147,39 +158,80 @@ func (p *postingsForMatcherPromise) result(ctx context.Context) (index.Postings,
...
 	}
 
+	// Add the caller context to the ones tracked by the new promise.
+	//
+	// It's important to do it here so that if the promise will be stored, then the caller's context is already
+	// tracked. Otherwise, if the promise will not be stored (because there's another in-flight promise for the
+	// same label matchers) then it's not a problem, and resources will be released.
+	//
+	// Skipping the error checking because it can't happen here.
+	_ = promise.callersCtxTracker.add(ctx)
```
colega, line 174: Nit: I think that we can add `ctx` to the constructor `newContextTracker()` and avoid this extra comment/logic here.

tsdb/postings_for_matchers_cache.go
```diff
@@ -298,3 +365,105 @@ func (ir indexReaderWithPostingsForMatchers) PostingsForMatchers(ctx context.Con
 }
 
 var _ IndexReader = indexReaderWithPostingsForMatchers{}
+
+// errContextsTrackerClosed is the reason to identify contextsTracker has been explicitly closed by calling close().
+//
+// This error is a struct instead of a globally generic error so that postingsForMatcherPromise computed size is smaller
+// (this error is referenced by contextsTracker, which is referenced by postingsForMatcherPromise).
```
colega, line 372: Thank you for this comment, it's not obvious.

tsdb/postings_for_matchers_cache_test.go
```diff
@@ -338,30 +341,250 @@ func TestPostingsForMatchersCache(t *testing.T) {
...
+
+		go func() {
+			defer wg.Done()
+
+			_, err := c.PostingsForMatchers(reqCtx2, indexForPostingsMock{}, true, matchers...)
+			require.ErrorIs(t, err, context.Canceled)
+		}()
+
+		// Give some time to let the 2nd request attach to the 1st one (we have no better way in tests to detect it).
+		time.Sleep(100 * time.Millisecond)
```
colega, line 513: Ugh, this sounds flaky. What if we try to find the promise (there should be only one in the sync.Map?) and inspect that it already has two subscribers?

tsdb/postings_for_matchers_cache.go
colega: Thank you, that looks great!

tsdb/postings_for_matchers_cache_test.go
pracucci: Good idea. Done in: https://github.com/grafana/mimir-prometheus/pull/644/commits/c1770fabc3d78c7626965f0281a968cfea72dd1e

tsdb/postings_for_matchers_cache.go
pracucci: I thought about it, but I'm not super convinced. We will pass a context to `newContextsTracker()` which than returns a context. I've the feeling it makes it harder to read the code, rather than explicitly calling `.add()`.

I would prefer to skip this.

colega (approved):

PR #645, opened 2024-06-10
Sync upstream prometheus to bring the MemPostings improvements.

colega:
This brings 4 last commits from upstream, I'm mostly interested in these two:
- https://github.com/prometheus/prometheus/pull/13286
- https://github.com/prometheus/prometheus/pull/14286

Solved minor conflicts in `postings_test.go`


colega:
`BenchmarkMemPostings_Delete` was added in the same place where we previously added `TestPostingsCloner`, and you know git doesn't like that because _OMG THEY'RE SHARING THE CLOSING BRACE_.

PR #646, opened 2024-06-11
prometheusremotewrite: Support OTel resource attribute promotion

aknuds1:
Add support to OTLP ingestion for promoting OTel resource attributes to labels.

We agreed to test this functionality in our Prom fork before proposing it upstream.

### DIFF
--- a/storage/remote/otlptranslator/prometheusremotewrite/helper.go
+++ b/storage/remote/otlptranslator/prometheusremotewrite/helper.go
@@ -66,14 +66,14 @@ type bucketBoundsData struct {
 	bound float64
 }
 
-// byBucketBoundsData enables the usage of sort.Sort() with a slice of bucket bounds
+// byBucketBoundsData enables the usage of sort.Sort() with a slice of bucket bounds.
 type byBucketBoundsData []bucketBoundsData
 
 func (m byBucketBoundsData) Len() int           { return len(m) }
 func (m byBucketBoundsData) Less(i, j int) bool { return m[i].bound < m[j].bound }
 func (m byBucketBoundsData) Swap(i, j int)      { m[i], m[j] = m[j], m[i] }
 
-// ByLabelName enables the usage of sort.Sort() with a slice of labels
+// ByLabelName enables the usage of sort.Sort() with a slice of labels.
 type ByLabelName []prompb.Label
 
 func (a ByLabelName) Len() int           { return len(a) }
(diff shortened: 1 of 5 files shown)


storage/remote/otlptranslator/prometheusremotewrite/helper.go
```diff
@@ -115,14 +115,24 @@ var seps = []byte{'\xff'}
...
 // if logOnOverwrite is true, the overwrite is logged. Resulting label names are sanitized.
-func createAttributes(resource pcommon.Resource, attributes pcommon.Map, externalLabels map[string]string,
+// If settings.PromoteResourceAttributes is not empty, it's a set of resource attributes that should be promoted to labels.
+func createAttributes(resource pcommon.Resource, attributes pcommon.Map, settings Settings,
 	ignoreAttrs []string, logOnOverwrite bool, extras ...string) []prompb.Label {
 	resourceAttrs := resource.Attributes()
 	serviceName, haveServiceName := resourceAttrs.Get(conventions.AttributeServiceName)
 	instance, haveInstanceID := resourceAttrs.Get(conventions.AttributeServiceInstanceID)
 
+	promotedAttrs := make([]prompb.Label, 0, len(settings.PromoteResourceAttributes))
```
ying-jeanne, line 125: I think this could get more efficient if we sort the configured promote attributes when read from configuration once, then range on settings.PromoteResourceAttributes, append when key exist in the settings. It can avoid the sort in every request. also this seems better controlled (take assumption we are the only one could change config)

something like, after sorting the config.PromoteResourceAttributes
for _, lb := range config.PromoteResourceAttributes {
  if value, exists := resourceAttrs.Get(lb); exists{
    promotedAttrs = append(promotedAttrs, prompb.Label{Name: lb, Value: value.AsString()})
  }
}

aknuds1:
> Wonder you thought about this already, I think this could get more efficient if we sort the configured promote attributes when read from configuration once, then range on settings.PromoteResourceAttributes, append when incoming attribute exist in the settings.

@ying-jeanne what you gain by sorting the configured attributes to promote beforehand is to avoid sorting `promotedAttrs`? You also have to make sure that the list of attributes to promote is unique though, this is why I range over `resourceAttrs` instead. There should be relatively few attributes to promote (assuming a limit prevents the user from promoting too many), so I guess iterating over this list, after making sure it's unique, should be fast. By the same argument (relatively few attributes to promote), though, I don't think pre-sorting gives you anything.

aknuds1:
> You also have to make sure that the list of attributes to promote is unique though, this is why I range over resourceAttrs instead.

@ying-jeanne in retrospect I realize that it doesn't matter if the list of attributes has duplicates, since labels stemming from promoted attributes are only added if the label isn't already in the set. This is effectively deduplication.

I've pushed a revision optimizing the promotion loop, PTAL.

storage/remote/otlptranslator/prometheusremotewrite/helper.go
```diff
@@ -115,14 +115,23 @@ var seps = []byte{'\xff'}
...
+	for _, name := range settings.PromoteResourceAttributes {
+		if value, ok := resourceAttrs.Get(name); ok {
+			promotedAttrs = append(promotedAttrs, prompb.Label{Name: name, Value: value.AsString()})
+		}
+	}
+	sort.Stable(ByLabelName(promotedAttrs))
```
ying-jeanne, line 131: if the config is sorted already by name, I think we can remove this sort here.

aknuds1:
Closing in favour of [upstream PR](https://github.com/prometheus/prometheus/pull/14200) (already merged).

PR #648, opened 2024-06-11
Cherry pick #644 into r291

pracucci:
Cherry pick #644 into r291


pracucci:
I think the linter in CI has failed because I initially opened again `main` instead of `weekly-r291`. I will re-run it.

colega (approved):

PR #650, opened 2024-06-11
[WIP] Otlp: Add info metrics identifiers labels in open metrics specification

ying-jeanne:


beorn7:
Was this discussed anywhere? Is this meant as a PoC to propose a change in OpenMetrics?

I think we should write down some thoughts before implementing anything (but having some PoC code is fine – that's why I'm asking if there is maybe some written down thoughts or a design doc anywhere).

My random thoughts:

- OpenMetrics protobuf already has the distinction between identifying and data labels. There isn't any implementation using OM proto yet (to my knowledge), but we can use the proto structure as inspiration for how to design the text format and what the parser should yield. (E.g. I could imagine to not return the data labels at all as part of the usual labels-returning functions but have a separate call for them. In this PR, it looks like all labels will be returned, and there is a separate call for the identifying labels, which would be a very different approach.)
- Adding another metadata line in OM text format is probably the way to go, but why `IDENS` as the keyword? Maybe just `ID` as the shortest recognizable keyword?
- We need to specify a default behavior (if the ID line is missing). I guess that "all the labels are data labels" is a good default, and nicely matches an empty ID line.
- Comma separation is problematic in light of the current plans to support all of UTF-8 in label names (as OTel does). We needed some quoting rules. Or maybe we can just require every identifying label to be on its own repeated `ID` line? I believe it will be rare to have an identifying label at all in the exposition (they usually come from target labels upon scrape), not to speak about having two or more. But I'm not sure if OM has a problem with repeated metadata lines.

aknuds1:
@beorn7 Ying, Jesus, and I just discussed - we think there should be a design doc for hashing out the change :)

PR #651, opened 2024-06-17
Update to upstream 64a9abb8be (`LabelValuesFor` signature change)

colega:
This brings the changes from upstream up to `64a9abb8be` which includes changes to `LabelValuesFor` method: https://github.com/prometheus/prometheus/pull/14280

This intentionally doesn't update to latest main, which will have more changes because of the compactor interface refactor: https://github.com/prometheus/prometheus/pull/14143


colega:
This introduces a performance regression in `LabelNamesFor` fixed later in https://github.com/prometheus/prometheus/commit/5efc8dd27b6e68d5102b77bc708e52c9821c5101

So this shouldn't go into the Mimir's weekly release without that.

model/labels/regexp.go
```diff
@@ -832,13 +834,51 @@ func (m *equalMultiStringMapMatcher) setMatches() []string {
...
+		s = toNormalisedLower(s)
 	}
 
 	_, ok := m.values[s]
 	return ok
 }
 
+// toNormalisedLower normalise the input string using "Unicode Normalization Form D" and then convert
+// it to lower case.
+func toNormalisedLower(s string) string {
```
aknuds1, line 846: Should this be mentioned in the PR description as a fix?

pracucci (approved):

model/labels/regexp.go
colega: Are we doing that? 

The PR will be a merge commit, and it has a list of all the commits we brought from upstream. I'm not sure what's the utility of copying that to the description.

aknuds1: Normally we tend to mention changes relevant to Mimir in these sync PRs.

colega: > Normally we tend to mention changes relevant to Mimir in these sync PRs.

I think that is a mistake, because it's hard to tell here which changes are relevant to Mimir and which are not.

We should mention the changes that are relevant to Mimir when we're vendoring the used modules in Mimir.

> Makes it easier for reviewers, and to know what should make it into the Mimir changelog when the time comes.

Not sure why it's easier, it's double work, we still have to review the changes when vendoring to Mimir.

PR #652, opened 2024-06-17
Merge upstream Prometheus

pstibrany:
This PR merges upstream Prometheus, with following PRs:

* https://github.com/prometheus/prometheus/pull/14094
* https://github.com/prometheus/prometheus/pull/14143
* https://github.com/prometheus/prometheus/pull/14300
* https://github.com/prometheus/prometheus/pull/14302

PR https://github.com/prometheus/prometheus/pull/14143 has conflicts related to our sharding compactor. I've adapted `Compact` method to match updated description of this method -- specifically, it now returns `nil` if resulting block doesn't have any samples. Previously this returned empty `ulid.ULID`.

### DIFF
--- a/tsdb/blockwriter.go
+++ b/tsdb/blockwriter.go
@@ -105,12 +105,17 @@ func (w *BlockWriter) Flush(ctx context.Context) (ulid.ULID, error) {
 	if err != nil {
 		return ulid.ULID{}, fmt.Errorf("create leveled compactor: %w", err)
 	}
-	id, err := compactor.Write(w.destinationDir, w.head, mint, maxt, nil)
+	ids, err := compactor.Write(w.destinationDir, w.head, mint, maxt, nil)
 	if err != nil {
 		return ulid.ULID{}, fmt.Errorf("compactor write: %w", err)
 	}
 
-	return id, nil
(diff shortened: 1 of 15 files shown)


tsdb/blockwriter.go
```diff
@@ -105,12 +105,17 @@ func (w *BlockWriter) Flush(ctx context.Context) (ulid.ULID, error) {
...
 		return ulid.ULID{}, fmt.Errorf("compactor write: %w", err)
 	}
 
-	return id, nil
+	// No block was produced. Caller is responsible to check empty
+	// ulid.ULID based on its use case.
+	if len(ids) == 0 {
+		return ulid.ULID{}, nil
+	}
+	return ids[0], nil
```
colega, line 118: What if there are more than 1 here?

pstibrany: This comes from upstream PR: https://github.com/prometheus/prometheus/pull/14143/files#diff-d1db4fe319c4c13342b55c33af552d4b418d41b8c9d956ac2fab440cbae3cb54

Prometheus always return one or no block when calling `compactor.Write`.

colega (approved):
Thank you!

PR #654, opened 2024-06-24
OTLP: Check context error during translation

aknuds1:
In the OTLP endpoint's translation layer, add checks for context cancellation/timeout every 128 times, similarly to what we do in e.g. `tsdb/index.MemPostings` code.

The motivation is that we've found that Mimir OTLP writes timing out continue the translation process, beyond the timeout. I suggest we test this improvement in Mimir first, and then eventually upstream it.

### DIFF
--- a/storage/remote/otlptranslator/prometheusremotewrite/context.go
+++ b/storage/remote/otlptranslator/prometheusremotewrite/context.go
@@ -28,7 +28,7 @@ func (e *everyNTimes) checkContext(ctx context.Context) error {
 	}
 
 	e.i++
-	if e.i%e.n == 0 {
+	if e.i >= e.n {
 		e.i = 0
 		e.err = ctx.Err()
 	}
--- a/storage/remote/otlptranslator/prometheusremotewrite/context.go
+++ b/storage/remote/otlptranslator/prometheusremotewrite/context.go
@@ -1,0 +1,37 @@
+// Copyright 2024 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package prometheusremotewrite
+
+import "context"
+
+// everyNTimes supports checking for context error every n times.
+type everyNTimes struct {
+	n   int
+	i   int
+	err error
+}
+
+// checkContext calls ctx.Err() every e.n times and returns an eventual error.
+func (e *everyNTimes) checkContext(ctx context.Context) error {
+	if e.err != nil {
+		return e.err
+	}
+
+	e.i++
+	if e.i%e.n == 0 {
+		e.i = 0
+		e.err = ctx.Err()
+	}
+
+	return e.err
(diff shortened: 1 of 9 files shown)


storage/remote/otlptranslator/prometheusremotewrite/context.go
```diff
@@ -0,0 +1,39 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package prometheusremotewrite
+
+import "context"
+
+// everyNIterationsContext wraps a context.Context, while its Err implementation only calls the wrapped
+// context's Err every n times.
+type everyNIterationsContext struct {
```
seizethedave, line 19: I think hiding/reusing the check-1-in-N-times logic is appealing, but I'm not sure this is better than adding the modulo check in multiple spots:

* if I have what looks like a `Context` and I call `Err()` on it, I will be surprised if it is hiding the fact that it's only going to check internally every N times. (principle of least astonishment, etc.) Imagine we add a step that is known to take 1 second- we'll want to definitely check Err before we do that work.
* If I have what looks like a `Context` I would expect `Err()` to be goroutine-safe.

the second one is fixable; I don't think the first one is. But let me hear those thoughts!

aknuds1: You're making some good points @seizethedave, thanks. I'll see if it can be solved in a better way.

aknuds1: @seizethedave in order to encapsulate the state of how many times the context has been checked for an error, I opted to keep it in a struct `everyNTimes`. The purpose is not to hide/re-use the logic, but to make the check happen for every 128th operation and not for every 128th iteration of each loop which can be nested on varying levels. Does that make sense to you?

seizethedave: Sounds good - I'll take a look.

storage/remote/otlptranslator/prometheusremotewrite/context.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// checkContext calls ctx.Err() every e.n times and returns an eventual error.
+func (e *everyNTimes) checkContext(ctx context.Context) error {
+	if e.err != nil {
+		return e.err
+	}
+
+	e.i++
+	if e.i%e.n == 0 {
```
seizethedave, line 31: Optional, just something that piqued my attention.

i%128 is as fast as i>=128, but for other values of `i`, modulo is quite a bit slower. Since `i` can be customized, we just want to use `>=`.

```
BenchmarkEvery100WithModulo-11          	 1810951	       660.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkEvery100WithGreaterEqual-11    	 3408360	       355.1 ns/op	       0 B/op	       0 allocs/op
->
BenchmarkEvery128WithModulo-11          	 3325652	       354.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkEvery128WithGreaterEqual-11    	 3353599	       354.7 ns/op	       0 B/op	       0 allocs/op
```

seizethedave (approved):
LGTM!

storage/remote/otlptranslator/prometheusremotewrite/context.go
aknuds1: TIL. Thanks!

seizethedave: that was supposed to say "we _may_ just want to use". 😸

aknuds1: Done.

PR #658, opened 2024-07-04
Sync upstream prometheus

pracucci:
Synching upstream prometheus because I want to get https://github.com/prometheus/prometheus/pull/13843

What's in this sync:
- https://github.com/prometheus/prometheus/pull/13843
- https://github.com/prometheus/prometheus/pull/12503
- https://github.com/prometheus/prometheus/pull/14396


promql/parser/parse.go
```diff
@@ -447,6 +447,10 @@ func (p *parser) newAggregateExpr(op Item, modifier, args Node) (ret *AggregateE
 
 	desiredArgs := 1
 	if ret.Op.IsAggregatorWithParam() {
+		if !EnableExperimentalFunctions && (ret.Op == LIMITK || ret.Op == LIMIT_RATIO) {
+			p.addParseErrf(ret.PositionRange(), "limitk() and limit_ratio() are experimental and must be enabled with --enable-feature=promql-experimental-functions")
```
colega, line 451: Will we have this feature enabled in Mimir? If not, we should modify this message (maybe make it configurable?)

colega (approved):
LGTM, left one minor comment.

promql/parser/parse.go
pracucci: Good point. I just customised the message, adding a comment to explain why (in case of future conflicts to resolve).

colega: Thank you, lgtm.

PR #661, opened 2024-07-17
Merge upstream at a28d197

charleskorn:
This PR brings in the following changes from upstream:

# Major changes

* https://github.com/prometheus/prometheus/pull/14438
* https://github.com/prometheus/prometheus/pull/14109
* https://github.com/prometheus/prometheus/pull/9138

# Minor changes

* https://github.com/prometheus/prometheus/pull/14469

# Not relevant for Mimir

* https://github.com/prometheus/prometheus/pull/14455 (which contains https://github.com/prometheus/prometheus/pull/14446)
* https://github.com/prometheus/prometheus/pull/14427

Dependency updates for dependencies not used in Mimir:

* https://github.com/prometheus/prometheus/pull/14459
* https://github.com/prometheus/prometheus/pull/14458
* https://github.com/prometheus/prometheus/pull/14390
* https://github.com/prometheus/prometheus/pull/14386
* https://github.com/prometheus/prometheus/pull/14468
* https://github.com/prometheus/prometheus/pull/14385
* https://github.com/prometheus/prometheus/pull/14375
* https://github.com/prometheus/prometheus/pull/14374
* https://github.com/prometheus/prometheus/pull/14379
* https://github.com/prometheus/prometheus/pull/14383

# Non-production changes (docs, CI pipeline, tests etc.)

* https://github.com/prometheus/prometheus/pull/14463
* https://github.com/prometheus/prometheus/pull/14467
* https://github.com/prometheus/prometheus/pull/14420

# Dependency updates

* https://github.com/prometheus/prometheus/pull/14384


pracucci:
@charleskorn I pushed a commit to change `TestSeriesHashCache_MeasureApproximateSizePerEntry` to use powers of 2 and then I've updated `approxBytesPerEntry` accordingly.

Why? Based on some research and experimentation:

- The allocated memory is driven by the map pre-allocation (in the test we preallocate based on the number of entries we'll write). Actually writing the entries or not doesn't make a different in allocate memory, because the map is preallocated.
- The internal map data structures are not pre-allocated to hold exactly the number of entries given, but it's rounded to powers of 2. This means that the actual allocated memory changes depending on the actual number of `numEntries` entries used in the test. Setting `numEntries` to a power of 2 yields consistent results (e.g. the estimated average size running the test with `2<<10`, `2<<11`, `2<<12`, etc... always returns 38, but the same is not true if you run the test with `1_000`, `10_000`, `100_000`, `1_000_000`, etc...).
- The actual memory reported by `TotalAlloc` is the allocated mspans. New mspans are allocated when go runtime can't fit new objects in the existing ones. For example, if I just allocate `_ = make([]byte, 10)` the result of `after.TotalAlloc-before.TotalAlloc` is `0` because the 10 bytes have been allocated in a already existing mspan. For this reason, it's important to run the test with a large number of `numEntries` to make sure new mspan are allocated.
- mspan memory fragmentation is another factor here, given `TotalAlloc` measures the allocated mspans and not actual objects.

Why was `TestSeriesHashCache_MeasureApproximateSizePerEntry` failing in this PR? I have no idea. I also run the test locally on main, and I get 28 in main (which is the value reported by this PR before my changes). I hope that using powers of 2 we just get more stable results.

pracucci (approved):

PR #666, opened 2024-08-05
[matalabel] Fix querier for tests

ying-jeanne:


ying-jeanne:
previously tests were failing because of this line https://github.com/grafana/mimir-prometheus/pull/666/files#diff-d1e13350906ab5a0f7e0c2b6419103539c97d8ece926aca61e1e147887900e52L160, when no meta label in the query, us ms instead to keep the logics untouched.

ying-jeanne:
after introduce the activate flag, all tests should pass

PR #667, opened 2024-08-05
metalabels: move label matching in separate metadata.go

ying-jeanne:
Separate metalabel filtering from the querier and head appender, placing it into its own file.

### DIFF
--- a/tsdb/metadata.go
+++ b/tsdb/metadata.go
@@ -1,0 +1,43 @@
+package tsdb
+
+import (
+	"github.com/grafana/regexp"
+
+	"github.com/prometheus/prometheus/model/labels"
+)
+
+const metaDataPrefix = `^__metadata__`
+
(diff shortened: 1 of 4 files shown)


jesusvazquez (approved):
LGTM

tsdb/metadata.go
```diff
@@ -0,0 +1,43 @@
...
+		} else {
+			normalMatchers = append(normalMatchers, m)
+		}
+	}
+	return metaMatchers, normalMatchers
+}
+
+func seperateMetaLabels(lbs labels.Labels) (labels.Labels, labels.Labels) {
+	// here we get the posting matches metadata
+	re := regexp.MustCompile(metaDataPrefix)
```
jesusvazquez, line 30: This should be done outside of this method else you're compiling the regex on every call. No need to do this on this PR i'll take care on the next one.

PR #669, opened 2024-08-06
Merge from upstream 73abc8a8bee3

bboreham:
* Particularly to bring in prometheus/prometheus#14584
* Also prometheus/prometheus#14575


56quarters (approved):
LGTM

PR #670, opened 2024-08-07
Fix: panic: runtime error: index out of range [4] with length 4

krajorama:
Port of https://github.com/prometheus/prometheus/pull/14598 From
https://github.com/krajorama/prometheus/commit/98ecdf35891cfd65d7cf01dc228c3d5d0dc82f6b


krajorama:
just used to run CI

PR #675, opened 2024-08-08
Metalabels working with stringlabels

codesome:
There is a slight difference in how the methods for normal and stringlabels work. This PR fixes queries when stringlabels is used.


jesusvazquez (approved):

tsdb/head_read.go
```diff
@@ -204,7 +204,9 @@ func (h *headIndexReader) Series(ref storage.SeriesRef, builder *labels.ScratchB
 		h.head.metrics.seriesNotFound.Inc()
 		return storage.ErrNotFound
 	}
-	builder.Assign(s.labels())
+	s.labels().Range(func(l labels.Label) {
```
codesome, line 207: Main fix

PR #677, opened 2024-08-10
TSDB: improve debug logging in compaction

bboreham:
While diagnosing https://github.com/prometheus/prometheus/pull/14584, the logging of chunks found to be out-of-order during ingester head compaction was lacking some useful information such as series labels.

If the chunks are `safeHeadChunk` it would print labels, but in the case I analyzed they were ordinary mmapped chunks.
So, this PR makes it log the labels that `PopulateBlock` has.

Refactor slightly to avoid logging blanks where information is unavailable.

(This function is only in mimir-prometheus, not upstream)

### DIFF
--- a/tsdb/compact.go
+++ b/tsdb/compact.go
@@ -939,7 +939,7 @@ func (c *LeveledCompactor) write(dest string, outBlocks []shardedBlock, blockPop
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):
LGTM

PR #689, opened 2024-09-05
OTLP: Insert created timestamp if starttimestamp is within interval

jesusvazquez:
Relates to https://opentelemetry.io/docs/specs/otel/metrics/data-model/#resets-and-gaps and https://github.com/open-telemetry/opentelemetry-specification/issues/4184

While testing [our previously introduced feature](https://github.com/grafana/mimir-prometheus/pull/684) to introduce 0's at `ts-1` when an OTLP sample had `StartTimeUnixNano == sample.timestamp` we noticed that it was not working for some usages of the OTel SDK and connectors like the spanmetrics connector.

Later we found out the reason is that it's unclear to what value `StartTImeUnixNano` should be set to when there is a reset in a cumulative stream. The links at the beginning give you more context on what the spec says and then a reasonable ask by Arthur asking to have a StartTimeUnixNano initialized per series and to the first observation.

Unfortunately it will take time to get there.

So what this PR does is to slightly modify our existing workaround of adding those zeros. Instead of writing the zero if the sample had `StartTimeUnixNano == sample.timestamp`, now it will only insert the zero if `the sample's StartTimeUnixNano is within 2 minutes of sample.Timestamp`

The downside of this approach is that we can't know the scrape interval of samples so in the case of more than 1 sample sent per minute, this workaround could lead to duplicate zero's for the same timestamp until that minute has passed. This incurs an increase of duplicate data in memory and an increase of DPM per series.
- This duplicate data in memory should be removed during compaction.

### DIFF
--- a/storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
+++ b/storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
@@ -21,7 +21,10 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.uber.org/multierr"
@@ -42,6 +45,12 @@ type Settings struct {
 	EnableCreatedTimestampZeroIngestion bool
 }
 
+type StartTsAndTs struct {
+	StartTs int64
+	Ts      int64
+	Labels  []prompb.Label
+}
+
 // PrometheusConverter converts from OTel write format to Prometheus remote write format.
 type PrometheusConverter struct {
 	unique    map[uint64]*prompb.TimeSeries
@@ -57,7 +66,7 @@ func NewPrometheusConverter() *PrometheusConverter {
 }
 
 // FromMetrics converts pmetric.Metrics to Prometheus remote write format.
-func (c *PrometheusConverter) FromMetrics(ctx context.Context, md pmetric.Metrics, settings Settings) (annots annotations.Annotations, errs error) {
+func (c *PrometheusConverter) FromMetrics(ctx context.Context, md pmetric.Metrics, settings Settings, logger log.Logger) (annots annotations.Annotations, errs error) {
 	c.everyN = everyNTimes{n: 128}
 	resourceMetricsSlice := md.ResourceMetrics()
 	for i := 0; i < resourceMetricsSlice.Len(); i++ {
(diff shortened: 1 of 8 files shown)


storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
```diff
@@ -166,6 +177,21 @@ func (c *PrometheusConverter) FromMetrics(ctx context.Context, md pmetric.Metric
 		addResourceTargetInfo(resource, settings, mostRecentTimestamp, c)
 	}
 
+	// TODO(jesus.vazquez) Temporary debug log to understand the behavior of start ts and ts. We're only tracking counters
+	for _, v := range c.startTsInfoPerSeries {
```
bboreham, line 181: Why store the info?  Why not log it out straight away when `trackStartTimestampForSeries` is called?

storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
```diff
@@ -42,22 +45,30 @@ type Settings struct {
...
+}
+
 // PrometheusConverter converts from OTel write format to Prometheus remote write format.
 type PrometheusConverter struct {
-	unique    map[uint64]*prompb.TimeSeries
-	conflicts map[uint64][]*prompb.TimeSeries
-	everyN    everyNTimes
+	unique               map[uint64]*prompb.TimeSeries
+	conflicts            map[uint64][]*prompb.TimeSeries
+	startTsInfoPerSeries map[uint64]StartTsAndTs
```
bboreham, line 58: nit: if this is for debugging only, I would comment that.

bboreham (approved):
Seems reasonable.

storage/remote/otlptranslator/prometheusremotewrite/number_data_points.go
```diff
@@ -65,14 +66,16 @@ func (c *PrometheusConverter) addGaugeNumberDataPoints(ctx context.Context, data
...
-	resource pcommon.Resource, metric pmetric.Metric, settings Settings, name string) error {
+	resource pcommon.Resource, metric pmetric.Metric, settings Settings, name string, logger log.Logger) error {
 	for x := 0; x < dataPoints.Len(); x++ {
 		if err := c.everyN.checkContext(ctx); err != nil {
 			return err
 		}
 
 		pt := dataPoints.At(x)
+		timestamp := convertTimeStamp(pt.Timestamp())
 		startTimestampNs := pt.StartTimestamp()
```
aknuds1, line 77: Could we just get rid of this variable now?

aknuds1 (approved):
I think it looks good, except I think maybe a `startTimestampNs` variable should be removed for consistency (see comment).

storage/remote/otlptranslator/prometheusremotewrite/number_data_points.go
jesusvazquez: Done!

jesusvazquez:
Note: I'm not updating the changelog because the existing entry is enough.

> * [ENHANCEMENT] OTLP receiver: If the feature flag `--created-timestamp-zero-ingestion` is true, convert OTel start timestamps to Prometheus zero samples. #14759

storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
jesusvazquez: You are right, dropped this map and just logged right away.

storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
jesusvazquez: Removed this map.

PR #695, opened 2024-09-17
OTEL logging: add log when adding extra sample, log metric type too.

pstibrany:
This PR enhances logging in OTEL translation code:
* it always logs a message when adding new zero-sample
* it includes metric type
* it now logs start_ts and sample_ts for all incoming samples and histograms.

The last point can generate a ton of messages when enabled.


jesusvazquez (approved):
LGTM

PR #698, opened 2024-09-18
merge upstream at prometheus/prometheus@14e3c05

krajorama:
Merge upstream at https://github.com/prometheus/prometheus/commit/14e3c05ce80a9e73b8a50e80448274e1fd3ff0ae

This is to add out-of-order native histogram support https://github.com/prometheus/prometheus/pull/14546

It also includes 3.0 beta

Notes: on my own review:
- excluding docs, discovery, UI changes there's something like 80 files to check

Read path:
- the interval boundary change [PR13904](https://github.com/prometheus/prometheus/pull/13904) is very noticeable from tests , but of course those are synthetic so that's to be expected

Write path:
- unsure if UTF8 label value validation affects us directly from here - but can be set differently if yes
- the other change is the out-of-order histograms, but they should be safe and are behind flag


krajorama:
outdated

PR #699, opened 2024-09-20
OTLP: Convert Exponential histograms start timestamps into zero samples

jesusvazquez:
Follow up PR to https://github.com/grafana/mimir-prometheus/pull/684 and https://github.com/grafana/mimir-prometheus/pull/689 where we add the same logic but for exponential histograms.

Unit tests included.

### DIFF
--- a/storage/remote/otlptranslator/prometheusremotewrite/helper.go
+++ b/storage/remote/otlptranslator/prometheusremotewrite/helper.go
@@ -617,7 +617,7 @@ func (c *PrometheusConverter) handleHistogramStartTime(startTs, sampleTs int64, 
 		return
 	}
 
-	ts.Histograms = append(ts.Histograms, prompb.Histogram{})
+	ts.Histograms = append(ts.Histograms, prompb.Histogram{Timestamp: startTs})
 }
 
 // addResourceTargetInfo converts the resource to the target info metric.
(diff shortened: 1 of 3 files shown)


storage/remote/otlptranslator/prometheusremotewrite/helper.go
```diff
@@ -599,6 +599,27 @@ func (c *PrometheusConverter) handleStartTime(startTs, ts int64, labels []prompb
...
+	if startTs <= 0 || startTs == sampleTs || startTs > sampleTs {
+		return
+	}
+
+	// The difference between the start and the actual timestamp is more than a reasonable time, so we skip this sample.
+	if sampleTs-startTs > validIntervalForStartTimestamps {
+		return
+	}
+
+	ts.Histograms = append(ts.Histograms, prompb.Histogram{Timestamp: startTs})
```
jesusvazquez, line 620: I'd like to know if this is correct cc @krajorama

krajorama: yes, empty histogram is valid because schema 0 is defined, and it's ok to have no buckets

krajorama (approved):
LGTM, I was a little unsure about setting the type of the Count field and leaving it nil, but it should work as far as I can tell and we spare an alloc

PR #701, opened 2024-09-20
[BUGFIX] TSDB: Only query chunks up to truncation time (#14948)

leizor:
https://raintank-corp.slack.com/archives/C07NBLRRK9A

Picking this fix into r307:

> If the query overlaps the range currently undergoing compaction, we should only fetch chunks up to that time. Need to store that min time in `HeadAndOOOIndexReader`.


krajorama (approved):
LGTM, this is the fix

PR #703, opened 2024-09-25
backport out-of-order native histograms pr14546 from upstream

krajorama:
There was one conflict because https://github.com/prometheus/prometheus/pull/14868 is not picked, but it only affects comment string. 

Includes:
- https://github.com/prometheus/prometheus/pull/14850  (small prepwork PR)
- https://github.com/prometheus/prometheus/pull/14546


jesusvazquez (approved):
LGTM

It's been mostly a re-read on the Prometheus one. DIdn't spot anything wrong related to mimir-prometheus. I think it should work!

PR #706, opened 2024-10-02
OTLP: Increase valid created timestamp threshold to 5 minutes

jesusvazquez:
Increasing valid threshold for inserting zero's at StartTimeUnixNano on OTLP Write requests.

The reason why we want to do this is that Grafana's $__rate_interval is typically 4 times the scrape interval so assuming an average DPM of 1 we could have the zero sample 4 minutes before the first sample. 

This is a small change that might help with usecases where samples are slightly delayed.

Also we've made this threshold configurable now. It defaults to 5 minutes but it can be overiden for special cases.

### DIFF
--- a/storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
+++ b/storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
@@ -35,15 +35,15 @@ import (
 )
 
 type Settings struct {
-	ExternalLabels                      map[string]string
-	Namespace                           string
-	PromoteResourceAttributes           []string
-	ValidIntervalForStartTimestamps     time.Duration
-	DisableTargetInfo                   bool
-	ExportCreatedMetric                 bool
-	AddMetricSuffixes                   bool
-	SendMetadata                        bool
-	EnableCreatedTimestampZeroIngestion bool
+	ExternalLabels                             map[string]string
+	Namespace                                  string
+	PromoteResourceAttributes                  []string
+	DisableTargetInfo                          bool
+	ExportCreatedMetric                        bool
+	AddMetricSuffixes                          bool
+	SendMetadata                               bool
+	EnableCreatedTimestampZeroIngestion        bool
+	ValidIntervalCreatedTimestampZeroIngestion time.Duration
 }
 
 type StartTsAndTs struct {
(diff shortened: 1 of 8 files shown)


storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go
```diff
@@ -34,20 +35,21 @@ import (
 )
 
 type Settings struct {
-	Namespace                           string
 	ExternalLabels                      map[string]string
+	Namespace                           string
+	PromoteResourceAttributes           []string
```
krajorama, line 40: If we're going to change the order , we should keep the original order of the upstream parameters AND leave an empty line before ours to avoid conflicts, see https://github.com/prometheus/prometheus/blob/199d2fb492322af138ef85d08ba022066fea784c/storage/remote/otlptranslator/prometheusremotewrite/metrics_to_prw.go#L34

krajorama: discussed IRL

krajorama (approved):
LGTM

web/web.go
```diff
@@ -244,27 +244,28 @@ type Options struct {
...
+	EnableLifecycle                            bool
+	EnableAdminAPI                             bool
+	PageTitle                                  string
+	RemoteReadSampleLimit                      int
+	RemoteReadConcurrencyLimit                 int
+	RemoteReadBytesInFrame                     int
+	EnableRemoteWriteReceiver                  bool
+	EnableOTLPWriteReceiver                    bool
+	EnableCreatedTimestampZeroIngestion        bool
+	ValidIntervalCreatedTimestampZeroIngestion time.Duration
```
krajorama, line 266: nit: these two cause conflicts when updating from upstream due to the indenting. Would be nice to eventually clean these up by adding our additions at the end after empty line

krajorama (approved):

web/web.go
jesusvazquez: You mean something like:
```
	RemoteReadBytesInFrame                     int
	EnableRemoteWriteReceiver                  bool
	EnableOTLPWriteReceiver                    bool

        // Our additions
	EnableCreatedTimestampZeroIngestion        bool
	ValidIntervalCreatedTimestampZeroIngestion time.Duration
```

krajorama (approved):
thanks

PR #708, opened 2024-10-03
Wrap ErrNotFound in some cases to get more information

krajorama:
Update tests to use ErrorIs and not equal. This could go upstream as the query code uses ErrorIs anyway so the tests are inconsistent with the production code.


krajorama:
not needed anymore here

PR #714, opened 2024-10-11
bugfix: Fix concurrent access for `seriesInPreviousEval` slice.

gotjosh:
With the introduction of rule evaluation concurrency, there's now a very obvious potential race whilst trying to access the `seriesInPreviousEval` slice.

This fixes it by adding a lock around the slice.

### DIFF
--- a/rules/group.go
+++ b/rules/group.go
@@ -45,20 +45,23 @@ import (
 
 // Group is a set of rules that have a logical relation.
 type Group struct {
-	name                 string
-	file                 string
-	interval             time.Duration
-	queryOffset          *time.Duration
-	limit                int
-	rules                []Rule
-	sourceTenants        []string
-	seriesInPreviousEval []map[string]labels.Labels // One per Rule.
-	staleSeries          []labels.Labels
-	opts                 *ManagerOptions
-	mtx                  sync.Mutex
-	evaluationTime       time.Duration
-	lastEvaluation       time.Time // Wall-clock time of most recent evaluation.
-	lastEvalTimestamp    time.Time // Time slot used for most recent evaluation.
+	name          string
+	file          string
+	interval      time.Duration
+	queryOffset   *time.Duration
+	limit         int
+	rules         []Rule
+	sourceTenants []string
+
+	seriesInPreviousEvalMtx sync.Mutex
+	seriesInPreviousEval    []map[string]labels.Labels // One per Rule.
+
+	staleSeries       []labels.Labels
+	opts              *ManagerOptions
+	mtx               sync.Mutex
+	evaluationTime    time.Duration
+	lastEvaluation    time.Time // Wall-clock time of most recent evaluation.
+	lastEvalTimestamp time.Time // Time slot used for most recent evaluation.
 
 	shouldRestore bool
 
@@ -209,6 +212,20 @@ func (g *Group) Interval() time.Duration { return g.interval }
 // Limit returns the group's limit.
 func (g *Group) Limit() int { return g.limit }
 
+// seriesInPreviousEvalForRuleIndex returns the series for the rule at the given index.
+func (g *Group) seriesInPreviousEvalForRuleIndex(i int) map[string]labels.Labels {
+	g.seriesInPreviousEvalMtx.Lock()
+	defer g.seriesInPreviousEvalMtx.Unlock()
+	return g.seriesInPreviousEval[i]
+}
+
+// setSeriesInPreviousEvalForRule sets the series for the rule at the given index.
+func (g *Group) setSeriesInPreviousEvalForRule(i int, series map[string]labels.Labels) {
+	g.seriesInPreviousEvalMtx.Lock()
+	defer g.seriesInPreviousEvalMtx.Unlock()
+	g.seriesInPreviousEval[i] = series
+}
+
 // SourceTenants returns the source tenants for the group.
 // If it's empty or nil, then the owning user/tenant is considered to be the source tenant.
 func (g *Group) SourceTenants() []string { return g.sourceTenants }
@@ -244,6 +261,7 @@ func (g *Group) run(ctx context.Context) {
 			return
 		}
 		go func(now time.Time) {
+			g.seriesInPreviousEvalMtx.Lock()
 			for _, rule := range g.seriesInPreviousEval {
 				for _, r := range rule {
 					g.staleSeries = append(g.staleSeries, r)
@@ -251,6 +269,7 @@ func (g *Group) run(ctx context.Context) {
 			}
 			// That can be garbage collected at this point.
 			g.seriesInPreviousEval = nil
+			g.seriesInPreviousEvalMtx.Unlock()
 			// Wait for 2 intervals to give the opportunity to renamed rules
 			// to insert new series in the tsdb. At this point if there is a
 			// renamed rule, it should already be started.
@@ -458,7 +477,9 @@ func (g *Group) CopyState(from *Group) {
 			continue
 		}
 		fi := indexes[0]
-		g.seriesInPreviousEval[i] = from.seriesInPreviousEval[fi]
+
+		g.setSeriesInPreviousEvalForRule(i, from.seriesInPreviousEvalForRuleIndex(fi))
+
 		ruleMap[nameAndLabels] = indexes[1:]
 
 		ar, ok := rule.(*AlertingRule)
@@ -481,7 +502,7 @@ func (g *Group) CopyState(from *Group) {
 		nameAndLabels := nameAndLabels(fromRule)
 		l := ruleMap[nameAndLabels]
 		if len(l) != 0 {
-			for _, series := range from.seriesInPreviousEval[fi] {
+			for _, series := range from.seriesInPreviousEvalForRuleIndex(fi) {
 				g.staleSeries = append(g.staleSeries, series)
 			}
 		}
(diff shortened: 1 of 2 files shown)


rules/manager_test.go
```diff
@@ -2357,6 +2357,45 @@ func TestUpdateWhenStopped(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestGroup_Eval_RaceCondition(t *testing.T) {
```
gotjosh, line 2360: I managed to confirm the race with this test and `go test -race ./rules/... -run "TestGroup_Eval_RaceCondition" -count 10`

I don't get failures anymore.

pracucci (commented):
I honestly can't understand how this PR could fix the following panic (which is what we experienced):

```
panic: runtime error: index out of range [11] with length 0

goroutine 240581 [running]:
github.com/prometheus/prometheus/rules.(*Group).Eval.func1(0xb, {0x3f519c0, 0xc016a58fc0}, 0xc0512a3860)
	/drone/src/vendor/github.com/prometheus/prometheus/rules/group.go:560 +0x227f
created by github.com/prometheus/prometheus/rules.(*Group).Eval in goroutine 7829
	/drone/src/vendor/github.com/prometheus/prometheus/rules/group.go:642 +0x34c
```

Looking at the panic we think that `g.seriesInPreviousEval` was accessed with index 11 but its length was 0. `g.seriesInPreviousEval` gets initialised in `NewGroup()` with the number of rules in the group. We don't expect the rules in the group to change (if they do, we expect the group to be re-created), so how is it possible that we accessed it with index 11 but the slice was empty?

In this PR you mention that you see a race on `g.seriesInPreviousEval` but where exactly? In `Eval()` we do have goroutines concurrently accessing it, but each goroutine is expected to access it at a different index. The `g.seriesInPreviousEval` slice length is never expected to change, so I can't see any race there (if there is please tell me more).

So where is the problem we've see in the panic? My theory is that the issue is `Group.run()`, specifically here:

```
// That can be garbage collected at this point.
g.seriesInPreviousEval = nil
```

rules/manager_test.go
```diff
@@ -2357,6 +2357,45 @@ func TestUpdateWhenStopped(t *testing.T) {
...
+	for _, group := range groups {
+		group := group
+
+		for i := 0; i < 100; i++ {
+			wg.Add(1)
+			go func() {
+				group.Eval(ctx, time.Now())
+				wg.Done()
+			}()
+		}
```
pracucci, line 2393: Isn't `Eval()` contract that will be never invoked concurrently?

pracucci:
I'm wondering if the panic we experienced is due to this cleanup we do in a defer in `Group.run()`:

```
// That can be garbage collected at this point.
g.seriesInPreviousEval = nil
``

In theory there should be no Eval() running when run ☝️ , but I'm wondering if we missed an edge case (e.g. when the rule group evaluation is getting terminated?).

**Comment by pracucci** [comment]
> In theory there should be no Eval() running when run ☝️ , but I'm wondering if we missed an edge case (e.g. when the rule group evaluation is getting terminated?).

Look what we do in `Group.Eval()` when a group is signalled to be terminated:

```
	for i, rule := range g.rules {
		select {
		case <-g.done:
			return
		default:
		}
```

We don't wait for in-flight evaluations to be completed! I think that's the root cause of the panic.

**Comment by gotjosh** [comment]
Thanks, @pracucci! I figured out that the cause of the panic was indeed the access to `seriesInPreviousEval,` and I thought that a catch-all of ensuring thread-safety would have been sufficient. 

Turns out I was completely wrong, I think your theory is 100% correct and what I ended up settling on was to make sure we wait for the group to finish before we return.

This would retain the original behaviour at the expense of maybe making updates to this particular manager a bit slower if we hit update in the middle of a concurrent evaluation. What do you think? The other option is to refactor the usage of `seriesInPreviousEval` to give it a snapshot of the current value for the goroutine.

File: rules/group.go [hunk 3]
```diff
@@ -501,6 +501,10 @@ func (g *Group) Eval(ctx context.Context, ts time.Time) {
 	for i, rule := range g.rules {
 		select {
 		case <-g.done:
+			// There's a chance that the group is asked to return early. In that case, we should
+			// wait for any in-flight rules to finish evaluating before returning so that we can preserve the same semantics.
+			// At the time of writing, the main reason for this was to make sure we don't clear seriesInPreviousEval before we're done using it.
+			wg.Wait()
```
pracucci, line 507: I was wondering it's more correct doing the `wg.Wait()` and `return` here, or alternatively break the `for` loop and let the rest of the function run (which does a `wg.Wait()` as first thing and then track a metric and calls `cleanupStaleSeries()`).

I'm trying to understand how `cleanupStaleSeries()` works, but I can't understand why it's called at the end of `Eval()`. In which case `g.staleSeries` is valued at the end of `Eval()`?  `g.staleSeries` is only valued by `g.run` in a `defer`, so how could `g.staleSeries` being ever values when `Eval()` runs?

> I was wondering it's more correct doing the `wg.Wait()` and `return` here, or alternatively break the `for` loop and let the rest of the function run (which does a `wg.Wait()` as first thing and then track a metric and calls `cleanupStaleSeries()`).

Back to us. Based on what I'm seeing I think there's no difference related to `cleanupStaleSeries()`. The only thing we miss doing the `wg.Wait()` is the metric tracking, but I have mixed feeling whether it's good or bad (we would track a partial metric, given it will be based on partial rule evaluations).

rules/manager_test.go
```diff
@@ -2357,6 +2357,39 @@ func TestUpdateWhenStopped(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestGroup_Eval_RaceCondition(t *testing.T) {
+	storage := teststorage.New(t)
+	t.Cleanup(func() { storage.Close() })
+
+	// ctx, cancel := context.WithCancel(context.Background())
+	// t.Cleanup(cancel)
```
pracucci, line 2365: [nit] Remove this, please.

rules/manager_test.go
```diff
@@ -2357,6 +2357,39 @@ func TestUpdateWhenStopped(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestGroup_Eval_RaceCondition(t *testing.T) {
```
pracucci, line 2360: [nit] I would clarify which race

```suggestion
func TestGroup_Eval_RaceConditionOnStoppingGroupEvaluationWhileRulesAreEvaluatedConcurrently(t *testing.T) {
```

rules/manager_test.go
```diff
@@ -2357,6 +2357,39 @@ func TestUpdateWhenStopped(t *testing.T) {
...
+	}()
+	<-ruleManager.block
+
+	// Update the group a decent number of times to simulate start and stopping in the middle of an evaluation.
+	for i := 0; i < 500; i++ {
+		err := ruleManager.Update(time.Second, files, labels.EmptyLabels(), "", nil)
+		require.NoError(t, err)
+		err = ruleManager.Update(time.Second, files2, labels.EmptyLabels(), "", nil)
+		require.NoError(t, err)
+	}
```
pracucci, line 2388: I struggled to reproduce the race on my local machine. I think that's because the next `Update()` may be called before we even start running the 1st query in the group if the machine is fast. With the following change, the race is reproduced consistently to me.

```suggestion
	for i := 0; i < 10; i++ {
		err := ruleManager.Update(time.Second, files, labels.EmptyLabels(), "", nil)
		require.NoError(t, err)

		// Wait half of the query execution duration and then change the rule groups loaded by the manager
		// so that the previous rule group will be interrupted while the query is executing.
		time.Sleep(artificialDelay / 2)

		err = ruleManager.Update(time.Second, files2, labels.EmptyLabels(), "", nil)
		require.NoError(t, err)
	}
```

pracucci (approved):

rules/manager_test.go
gotjosh: Nice! I had to resort to using `-count` to test it, but this is much nicer. Thanks!

rules/manager_test.go
gotjosh: I actually made the commit yesterday, but forgot to push it - It's in [eac3b8d](https://github.com/grafana/mimir-prometheus/pull/714/commits/eac3b8d31a1e0489071523461096726dca3c7a26).

Because of this, github places it before your comment (just an FYI).

rules/manager_test.go
gotjosh: Done in https://github.com/grafana/mimir-prometheus/pull/714/commits/e829903745c2ee2f4a697045370f67946fc353a7

rules/manager_test.go
gotjosh: Done in https://github.com/grafana/mimir-prometheus/pull/714/commits/e07d041c417e507488925adf7e0ac75600a3ad3b

rules/group.go
gotjosh: > I'm trying to understand how cleanupStaleSeries() works, but I can't understand why it's called at the end of Eval(). In which case g.staleSeries is valued at the end of Eval()? g.staleSeries is only valued by g.run in a defer, so how could g.staleSeries being ever values when Eval() runs?

Upon inspection - I agree with you this is confusing, and there's no reason why this is valued to be called at the end - it could perfectly be called at any point during `Eval()`

The caveat here is [the usage of markStale](https://github.com/grafana/mimir-prometheus/blob/d341d3f0f910bac49bcac36bfb2971daacf8e793/rules/manager.go#L264) which would indicate that only when the group is the same we need to rely on the next evaluation to actually cleanup the stale series: https://github.com/grafana/mimir-prometheus/blob/main/rules/manager.go#L243C5-L244 - because the defer as we stop that group wouldn't trigger. So when the copy of the old group is run next we need to rely on it cleaning up the series.

All in all, returning here means we can stop the old group faster and let the new group finish its evaluation and eventually (if needed) _actually_ call stale series after the next evaluation so I guess this is fine?

> Back to us. Based on what I'm seeing I think there's no difference related to cleanupStaleSeries(). The only thing we miss doing the wg.Wait() is the metric tracking, but I have mixed feeling whether it's good or bad (we would track a partial metric, given it will be based on partial rule evaluations).

Great shout, after looking at it - I have to side with you, I would rather not track the metric at all as the gauge would suffer from sudden drops that might not make more sense (e.g. stopping after just 1 rule on a group of 30 rules).

PR #716, opened 2024-10-15
Implement index.MaybeRecyclePostings

colega:
This method would call an optional Recycle() method on the postings, that would reuse the reference by putting it into a pool.

The heuristic I've followed is that it should be called by the method that performs the Next() check, as it's that method who knows that postings can't be used anymore.

### DIFF
--- a/tsdb/index/postings.go
+++ b/tsdb/index/postings.go
@@ -447,6 +447,7 @@ func (p *MemPostings) labelValues(name string) []string {
 
 // ExpandPostings returns the postings expanded as a slice.
 func ExpandPostings(p Postings) (res []storage.SeriesRef, err error) {
+	defer MaybeRecyclePostings(p)
 	for p.Next() {
 		res = append(res, p.At())
 	}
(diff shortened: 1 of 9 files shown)


tsdb/index/postings.go
```diff
@@ -1002,3 +1069,16 @@ func (h *postingsWithIndexHeap) Pop() interface{} {
...
 }
+
+func MaybeRecyclePostings(p Postings) {
+	if r, ok := p.(recycler); ok {
+		r.Recycle()
+	}
+}
+
+// recycler is an interface for types that can be recycled.
+// it's not exported to avoid creating a contract on its name.
```
dimitarvdimitrov, line 1080: aren't contracts based on the methods, not the names ❓ 

i'm not sure if you meant to have the method as private too - i think _that_ prevents other packages from implementing this interface

colega: Everything exported is a contract, I didn't want to create a contract of having an `index.Recycler`

And yes, I explicitly left the method exported so anyone can implement the interface (I can imagine usages of this in Mimir)

colega:
Closing in favour of a newer version built on top of r313 commit: https://github.com/grafana/mimir-prometheus/pull/727

PR #727, opened 2024-10-21
Implement index.MaybeRecyclePostings (r313)

colega:
This method would call an optional Recycle() method on the postings, that would reuse the reference by putting it into a pool.

The heuristic I've followed is that it should be called by the method that performs the Next() check, as it's that method who knows that postings can't be used anymore.


colega:
While I saw a decrease in allocations, I didn't see a decrease in memory or cpu usage that would justify this complexity.

I also saw some allocations of bigendian postings that were never recycled.

PR #730, opened 2024-10-29
Fix `MemPostings.Add` and `MemPostings.Get` data race

aknuds1:
Cherry-pick https://github.com/prometheus/prometheus/pull/15141.

When references are not added in order, we are sorting in place the tail of the postings list held by MemPostings, however the array backing that slice might already being used by some readers.


colega (approved):
LGTM, however I think that we should just bring the entire latest main.

aknuds1:
> LGTM, however I think that we should just bring the entire latest main.

I'll make a dedicated PR for that. I just want to get the race condition fixes in first :)

PR #733, opened 2024-10-29
Do not call span.AddEvent() with PostingsForMatchersCache lock taken

pracucci:
We shouldn't call `span.AddEvent()` with `PostingsForMatchersCache` lock taken. It keeps the lock for longer, for no good reason.


56quarters (approved):
LGTM

PR #734, opened 2024-10-29
Reduce PostingsForMatchersCache.expire() pressure on mutex

pracucci:
Currently, `PostingsForMatchersCache.expire()` is called for every single `PostingsForMatchers()` call. If the cache is full, it's expected that there's always something to expire from the cache (1 in, 1 out). However, in a high concurrency environment, there's no need to call `expire()` **concurrently**. We just need 1 goroutine to cleanup stale or over-the-capacity cached entries.

In this PR I propose a simple solution to only 1 `expire()` execution at a time.

I've added a new benchmark:

```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                                                          │    sec/op    │   sec/op     vs base                │
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11   1653.0n ± 4%   335.1n ± 8%  -79.73% (p=0.000 n=10)

                                                          │ before.txt  │              after.txt              │
                                                          │    B/op     │    B/op      vs base                │
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11   1446.5 ± 0%   1004.0 ± 0%  -30.59% (p=0.000 n=10)

                                                          │ before.txt │             after.txt              │
                                                          │ allocs/op  │ allocs/op   vs base                │
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11   29.00 ± 0%   20.00 ± 0%  -31.03% (p=0.000 n=10)
```

### DIFF
--- a/tsdb/postings_for_matchers_cache.go
+++ b/tsdb/postings_for_matchers_cache.go
@@ -14,6 +14,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/atomic"
 
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/tsdb/index"
@@ -49,8 +50,9 @@ type IndexPostingsReader interface {
 // If `force` is true, then all requests go through cache, regardless of the `concurrent` param provided to the PostingsForMatchers method.
 func NewPostingsForMatchersCache(ttl time.Duration, maxItems int, maxBytes int64, force bool) *PostingsForMatchersCache {
 	b := &PostingsForMatchersCache{
-		calls:  &sync.Map{},
-		cached: list.New(),
+		calls:            &sync.Map{},
+		cached:           list.New(),
+		expireInProgress: atomic.NewBool(false),
 
 		ttl:      ttl,
 		maxItems: maxItems,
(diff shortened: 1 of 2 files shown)


56quarters (approved):
Changes LGTM. It seems like this is a case where `RWLock.TryRLock()` could be used but I don't feel strongly about using it instead of the atomic here.

pracucci:
> Changes LGTM. It seems like this is a case where `RWLock.TryRLock()` could be used but I don't feel strongly about using it instead of the atomic here.

I tried it, but it looks slower. Here is the comparison between this PR vs `Try...Lock`:

```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                                                          │ before.txt  │              after.txt               │
                                                          │   sec/op    │   sec/op     vs base                 │
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11   303.8n ± 2%   795.8n ± 2%  +161.93% (p=0.000 n=10)

                                                          │ before.txt  │              after.txt              │
                                                          │    B/op     │    B/op      vs base                │
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11   1016.0 ± 0%   1371.5 ± 0%  +34.99% (p=0.000 n=10)

                                                          │ before.txt │             after.txt              │
                                                          │ allocs/op  │ allocs/op   vs base                │
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11   21.00 ± 0%   27.00 ± 0%  +28.57% (p=0.000 n=10)
```

Note: it's still faster than `main`, but not as fast as this PR.

56quarters:
Makes sense, thanks!

PR #751, opened 2024-11-15
Build 24h blocks for older OOO data

codesome:
For https://github.com/grafana/mimir/issues/9911

In order to reduce the number of blocks when a long OOO window is configured.

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -1309,7 +1309,7 @@ func (db *DB) Compact(ctx context.Context) (returnErr error) {
 		}
 	}()
 
-	lastBlockMaxt, firstBlockMint := int64(math.MinInt64), int64(math.MaxInt64)
+	lastBlockMaxt := int64(math.MinInt64)
 	defer func() {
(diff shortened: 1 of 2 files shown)


tsdb/db.go
```diff
@@ -1432,6 +1432,7 @@ func (db *DB) CompactOOOHead(ctx context.Context) error {
 // Callback for testing.
 var compactOOOHeadTestingCallback func()
 
+// firstBlockMint is in milliseconds.
```
chencs, line 1435: Is this comment supposed to be here?

chencs (approved):

tsdb/db.go
```diff
@@ -1480,7 +1480,6 @@ func (db *DB) compactOOOHead(ctx context.Context) error {
 			return fmt.Errorf("truncate ooo wbl: %w", err)
 		}
 	}
-
```
chencs, line 1483: 0% blocking, might want to put this line back

PR #757, opened 2024-11-19
PostingsForLabelMatching: check for waiting writers

colega:
There's still some contention on MemPostings.mtx, I checked goroutines from a profile at a spiking mutex wait time moment, I filtered by `MemPostings` and all I see is:

- 4 goroutines waiting on `p.mtx.RLock()` in `LabelValues`
- 203 goroutines waiting on `p.mtx.RLock()` in `Get`
- 6 goroutines waiting on `p.mtx.Lock()` in `Add`
- 4 goroutines waiting on first `p.mtx.RLock` in `PostingsForLabelMatching`
- 2 goroutines in `for _, v := range vals` loop in `PostingsForLabelMatching`

So, what I undersand is going on here:
- The last two goroutines were able to enter the `PostingsForLabelMatching` second mutex, so they're creating ListPostings as there were no tomorrow (I'll send a separate PR to Prometheus to avoid creating one pointer for each one there)
- The six goroutines waiting on `Add` are blocking the reset 200+ read goroutines.

So, here we have 2 read goroutines blocking 200 goroutines (plus six read ones, if someone cares).

What this change proposes is to check every few iterations (1024, because I expect this loop to go quite fast) whether there's a Lock() waiting. If there is one, we'll stop, unlock, and let it happen, because if we're being selfish, no other reader can get in here.

image

### DIFF
--- a/tsdb/index/postings.go
+++ b/tsdb/index/postings.go
@@ -471,7 +471,17 @@ func (p *MemPostings) PostingsForLabelMatching(ctx context.Context, name string,
 	its := make([]Postings, 0, len(vals))
 	p.mtx.RLock()
 	e := p.m[name]
+	iter := 0
 	for _, v := range vals {
+		iter++
+		if iter%1024 == 0 && p.writersWaiting() {
+			// There are writers waiting, so unlock the mutex, and use this opportunity to check the context.
+			p.mtx.RUnlock()
+			if ctx.Err() != nil {
+				return ErrPostings(ctx.Err())
+			}
+			p.mtx.RLock()
+		}
 		if refs, ok := e[v]; ok {
 			// Some of the values may have been garbage-collected in the meantime this is fine, we'll just skip them.
 			// If we didn't let the mutex go, we'd have these postings here, but they would be pointing nowhere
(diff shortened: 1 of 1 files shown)


colega:
One more time:
image

colega:
And another one:

image

colega:
And another one:

image

tsdb/index/postings.go
```diff
@@ -471,7 +471,17 @@ func (p *MemPostings) PostingsForLabelMatching(ctx context.Context, name string,
 	its := make([]Postings, 0, len(vals))
 	p.mtx.RLock()
 	e := p.m[name]
+	iter := 0
 	for _, v := range vals {
+		iter++
+		if iter%1024 == 0 && p.writersWaiting() {
+			// There are writers waiting, so unlock the mutex, and use this opportunity to check the context.
```
jesusvazquez, line 478: Personally I'd go one step further with this comment and say that the current RWMutex does not offer these custom scheduling capabilities so this is a best effort to be fair. An alternative to this check every 1024 iterations would be to reimplement our own mutex.

jesusvazquez (approved):
LGTM

tsdb/index/postings.go
colega: Well, what I'm doing is adapting to the specific scheduling rules of the current `sync.RWMutex`. I don't want to bring conversation about potential reimplementations here.

colega:
We can solve that one by doing: 
```go
func (p *MemPostings) Symbols() StringIter {
	p.mtx.RLock()
	lvs := maps.Clone(p.lvs)
	p.mtx.RUnlock()
```

I'll send that in a separate PR.

colega:
I didn't find enough evidence of this being helpful.

PR #761, opened 2024-11-21
MemPostings: allocate ListPostings once in PFLM

colega:
Cherry-pick of https://github.com/prometheus/prometheus/pulls/15427

Instead of allocating ListPostings pointers one by one, allocate a slice and take pointers from that. It's faster, and also generates less garbage (NewListPostings is one of the top offenders in number of allocations).


aknuds1 (approved):
Smart!

PR #766, opened 2024-11-28
TSDB: Add 'quiet zero' for Otel start-time handling

bboreham:
When we want to add a zero at the start-time of the series, we don't know if a sample was previously added at that timestamp. So, use a special value which has the correct behaviour when it reaches TSDB.

Part of https://github.com/prometheus/prometheus/issues/14600

### DIFF
--- a/model/value/value.go
+++ b/model/value/value.go
@@ -26,6 +26,9 @@ const (
 	// complicated values in the future. It is 2 rather than 1 to make
 	// it easier to distinguish from the NormalNaN by a human when debugging.
 	StaleNaN uint64 = 0x7ff0000000000002
+
+	// QuietZeroNaN signals TSDB to add a zero, but do nothing if there is already a value at that timestamp.
+	QuietZeroNaN uint64 = 0x7ff0000000000003
 )
 
 // IsStaleNaN returns true when the provided NaN value is a stale marker.
--- a/storage/remote/otlptranslator/prometheusremotewrite/helper.go
+++ b/storage/remote/otlptranslator/prometheusremotewrite/helper.go
@@ -589,9 +589,10 @@ const defaultIntervalForStartTimestamps = int64(300_000)
 // handleStartTime adds a zero sample at startTs only if startTs is within validIntervalForStartTimestamps of the sample timestamp.
 // The reason for doing this is that PRW v1 doesn't support Created Timestamps. After switching to PRW v2's direct CT support,
 // make use of its direct support fort Created Timestamps instead.
+// See https://github.com/prometheus/prometheus/issues/14600 for context.
 // See https://opentelemetry.io/docs/specs/otel/metrics/data-model/#resets-and-gaps to know more about how OTel handles
 // resets for cumulative metrics.
-func (c *PrometheusConverter) handleStartTime(startTs, ts int64, labels []prompb.Label, settings Settings, typ string, value float64, logger *slog.Logger) {
+func (c *PrometheusConverter) handleStartTime(startTs, ts int64, labels []prompb.Label, settings Settings, typ string, val float64, logger *slog.Logger) {
 	if !settings.EnableCreatedTimestampZeroIngestion {
 		return
 	}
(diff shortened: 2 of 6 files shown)


bboreham:
CC @ArthurSens I think this is the reason `AppendCTZeroSample` was added.
In Mimir we have to the addition in a different process to where TSDB lives so I came up with the idea to use a special NaN.
Please comment if you see issues with the idea in this PR.

aknuds1 (approved):

ArthurSens (approved):
I see no obvious scenarios where this would be a problem.

> In Mimir we have to the addition in a different process to where TSDB lives 

Interesting. I was wondering why the OTLP handler doesn't have an appender interface and couldn't just call `AppendCTZeroSample`. Thanks for clarifying! Is there a place where I can see how the Prometheus codebase relates to each Mimir process?

jesusvazquez (approved):
LGTM!

bboreham:
[Mimir Architecture](https://grafana.com/docs/mimir/latest/get-started/about-grafana-mimir-architecture/) 

(it doesn't specifically map Prometheus components, but you can get the general idea)

bboreham:
I guess we need an analogue for `AppendHistogramCTZeroSample` next.

jesusvazquez:
@bboreham no need to hurry for histograms. Until OOO for NH is stable we can skip it.

bboreham:
If this idea works I'd like to remove `AppendCTZero` in Prometheus, because we don't need two ways to do it.

model/value/value.go
```diff
@@ -26,6 +26,9 @@ const (
 	// complicated values in the future. It is 2 rather than 1 to make
 	// it easier to distinguish from the NormalNaN by a human when debugging.
 	StaleNaN uint64 = 0x7ff0000000000002
+
+	// QuietZeroNaN signals TSDB to add a zero, but do nothing if there is already a value at that timestamp.
```
colega, line 30: I think we should be more precise on this comment, as it only checks the last value.

Given this feature is for OTEL, and OTEL sends a lot of OOO samples, it often won't check whether there's really a value at that timestamp or not.

bboreham: Good catch.
However I am now thinking we should also drop the quiet zero if it is OOO.
In the expected flow, we get `startTimeUnixNano` on every sample for a series, and append [quiet zero, real sample] in that order.  We only really want to append the zero the first time it appears.
(we put a limit (5 mins?) to avoid re-sending the zero forever)
Is there any case where we would already have some OOO data, and then want to add the start-time zero ?

bboreham: Separately, I was thinking the docstring could read "do nothing if we detect there is already a value at that timestamp".

ArthurSens: In Prometheus' `AppendCTZeroSample`, we drop OOO always, even when OOO is enabled

jesusvazquez: Not sure if its safe to make the assumption. If samples come unordered we could lose the zero.

For OTel typically all payloads contain the starttimeunixnano but the spec says its optional https://github.com/jmacd/opentelemetry-proto/blob/5951df42ef915e54daa72552a77602783ebf7f17/opentelemetry/proto/metrics/v1/metrics.proto#L109

So if you have sample1(with both timeunixnano + startimeunixnano) and sample2 (with only timeunixnano) arriving unordered (sample2, sample1) we could lose the zero.

bboreham: See https://github.com/grafana/mimir-prometheus/pull/773

bboreham: I think it's ok to move forward while noting that this scheme isn't perfect.

> OTEL sends a lot of OOO samples

This seems worth investigating.  Within the scope of system monitoring, samples for one series will be many seconds apart, so whatever took them out of order must involve a lot of buffering.
(I have seen many cases where out-of-order data is created by misconfiguration, e.g. series that are supposed to be distinct being merged together into one.)

PR #773, opened 2024-11-29
TSDB: Always drop 'quiet zeros' if out-of-order

bboreham:
Follow-up to https://github.com/grafana/mimir-prometheus/pull/766

"Quiet Zeros" are used to insert the starting sample for a counter. If the zero is out-of-order then we already have some data for the series, and we should have already handled the start value, so just drop the quiet zero.

There is one case not handled: if we receive (e.g. over OTLP) the second sample for a series before the first, and the first sample is nonzero and has time == startTime, then we will have inserted a zero erroneously at the start time.  If OOO is enabled we will then insert the correct value, and it is undetermined which value will come out at query time.  This case is not made worse by this PR.

### DIFF
--- a/tsdb/head_append.go
+++ b/tsdb/head_append.go
@@ -505,6 +505,10 @@ func (s *memSeries) appendable(t int64, v float64, headMaxt, minValidTime, oooTi
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


ArthurSens (approved):

jesusvazquez (approved):
LGTM

PR #782, opened 2024-12-03
Merge upstream at 140f4aa

charleskorn:
This PR advances this fork closer to upstream Prometheus `main`, bringing in the changes from https://github.com/prometheus/prometheus/commit/140f4aa9aed6674c582dc603431a4aaa0c702cb7, which is the merge of https://github.com/prometheus/prometheus/pull/13567.

This merge had a number of conflicts. While I _believe_ I've merged this correctly given there were minimal conflicts in the tests and all tests are passing, carefully checking these changes would be appreciated.

Conflicts were in:

* `tsdb/block.go`
* `tsdb/compact.go`
* `tsdb/db.go`
* `tsdb/index/index.go`
* `tsdb/index/index_test.go`
* `tsdb/querier_bench_test.go`

### DIFF
--- a/tsdb/db.go
+++ b/tsdb/db.go
@@ -93,6 +93,7 @@ func DefaultOptions() *Options {
 		EnableDelayedCompaction:     false,
 		CompactionDelayMaxPercent:   DefaultCompactionDelayMaxPercent,
 		CompactionDelay:             time.Duration(0),
+		PostingsDecoderFactory:      DefaultPostingsDecoderFactory,
 
 		HeadChunksEndTimeVariance:             0,
 		HeadPostingsForMatchersCacheTTL:       DefaultPostingsForMatchersCacheTTL,
@@ -283,6 +284,10 @@ type Options struct {
 
 	// BlockChunkQuerierFunc is a function to return storage.ChunkQuerier from a BlockReader.
 	BlockChunkQuerierFunc BlockChunkQuerierFunc
+
+	// PostingsDecoderFactory allows users to customize postings decoders based on BlockMeta.
+	// By default, DefaultPostingsDecoderFactory will be used to create raw posting decoder.
+	PostingsDecoderFactory PostingsDecoderFactory
 }
 
 type NewCompactorFunc func(ctx context.Context, r prometheus.Registerer, l *slog.Logger, ranges []int64, pool chunkenc.Pool, opts *Options) (Compactor, error)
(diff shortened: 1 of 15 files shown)


tsdb/db.go
```diff
@@ -697,7 +702,7 @@ func (db *DBReadOnly) Blocks() ([]BlockReader, error) {
 		return nil, ErrClosed
 	default:
 	}
-	loadable, corrupted, err := openBlocks(db.logger, db.dir, nil, nil, nil, DefaultPostingsForMatchersCacheTTL, DefaultPostingsForMatchersCacheMaxItems, DefaultPostingsForMatchersCacheMaxBytes, DefaultPostingsForMatchersCacheForce)
+	loadable, corrupted, err := openBlocks(db.logger, db.dir, nil, nil, DefaultPostingsDecoderFactory, nil, DefaultPostingsForMatchersCacheTTL, DefaultPostingsForMatchersCacheMaxItems, DefaultPostingsForMatchersCacheMaxBytes, DefaultPostingsForMatchersCacheForce)
```
krajorama, line 705: note: this is weird in upstream as well, why would the default be able to decode all? why isn't this coming from options?

tsdb/index/index.go
```diff
@@ -1165,27 +1167,27 @@ func (b realByteSlice) Sub(start, end int) ByteSlice {
...
 }
 
 // NewFileReader returns a new index reader against the given index file.
-func NewFileReader(path string) (*Reader, error) {
-	return NewFileReaderWithOptions(path, nil)
+func NewFileReader(path string, decoder PostingsDecoder) (*Reader, error) {
+	return NewFileReaderWithOptions(path, decoder, nil)
 }
 
 // NewFileReaderWithOptions is like NewFileReader but allows to pass a cache provider and sharding function.
```
krajorama, line 1184: While we're here. The doc string wasn't true since #411 

```suggestion
// NewFileReaderWithOptions is like NewFileReader but allows to pass a cache provider.
```

krajorama (approved):
Approved with comment (I think I'll just make that a separate PR)
I believe the fact that types changed makes it hard to actually mess this up due to compiler checks. Also index/labelvalues.go seems totally analogous to the upstream changes.

tsdb/index/index.go
krajorama: https://github.com/grafana/mimir-prometheus/pull/783

PR #808, opened 2024-12-18
DO NOT MERGE: Update compactor to fix blocks with quiet zero NaNs

fionaliao:

### DIFF
--- a/tsdb/compact.go
+++ b/tsdb/compact.go
@@ -20,11 +20,14 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"math"
 	"os"
 	"path/filepath"
 	"slices"
 	"sync"
 	"time"
+
+	"github.com/prometheus/prometheus/model/value"
 
 	"github.com/oklog/ulid"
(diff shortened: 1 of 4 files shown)


tsdb/compact.go
```diff
@@ -1167,7 +1185,19 @@ func (c DefaultBlockPopulator) PopulateBlock(ctx context.Context, metrics *Compa
 			// We are not iterating in a streaming way over chunks as
 			// it's more efficient to do bulk write for index and
 			// chunk file purposes.
-			chks = append(chks, chksIter.At())
+			chk := chksIter.At()
+			if shouldRemoveQuietZeroNaNs {
+				updatedChunk, chunkCreated, err := removeQuietZeroNaNs(chk)
```
bboreham, line 1190: Nit: `chunkCreated` is true when the encoding is something other than xor. So it could be `chunkHasData` ?

fionaliao: Done in https://github.com/grafana/mimir-prometheus/pull/808/commits/14c7fb39e5595ab99e55ba8549b2d7aee8202f3f

PR #811, opened 2024-12-24
tsdb: Add MustIndex() function to Head

ying-jeanne:
In the context of cost attribution in this PR [#10269](https://github.com/grafana/mimir/pull/10269), I need to retrieve the index to decrement the active series count when a series goes out of the 20-minute active series window. Since the current Index() function does not return an error, we propose introducing a new function, MustIndex(), which will never return an error. Additionally, I will add tests to ensure that both functions return the same results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [CHANGE] Notifier: Increment the prometheus_notifications_errors_total metric by the number of affected alerts rather than by one per batch of affected alerts. #15428
 * [ENHANCEMENT] OTLP receiver: Convert also metric metadata. #15416
+* [ENHANCEMENT] Introduced MustIndex() as a variant of Index() in tsdb Head that does not return an error. #811
 
 ## 3.0.0 / 2024-11-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 * [CHANGE] Notifier: Increment the prometheus_notifications_errors_total metric by the number of affected alerts rather than by one per batch of affected alerts. #15428
 * [ENHANCEMENT] OTLP receiver: Convert also metric metadata. #15416
-* [ENHANCEMENT] Introduced MustIndex() as a variant of Index() in tsdb Head that does not return an error. #811
 
 ## 3.0.0 / 2024-11-14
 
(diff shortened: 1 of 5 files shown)


tsdb/head_read.go
```diff
@@ -37,6 +37,10 @@ func (h *Head) Index() (IndexReader, error) {
 	return h.indexRange(math.MinInt64, math.MaxInt64), nil
 }
 
+func (h *Head) MustIndex() IndexReader {
+	return h.indexRange(math.MinInt64, math.MaxInt64)
+}
+
```
colega, line 43: Nit, we don't have a policy on this yet, but can you create a new file, called `head_read_mimir.go` and put this method there? This way we'll avoid a bunch of confclicting updates with upstream in the future.

tsdb/head_read_test.go
```diff
@@ -428,6 +428,9 @@ func TestHeadIndexReader_PostingsForLabelMatching(t *testing.T) {
 
 		ir, err := h.Index()
 		require.NoError(t, err)
+
+		irMust := h.MustIndex()
+		require.Equal(t, irMust, ir)
```
colega, line 433: This should go in a separate test (also as noted above, in a separate file)

tsdb/head_read.go
ying-jeanne: addressed in 7a70886a33733ee7d1c790c8da6f957ecdcca833

tsdb/head_read_test.go
ying-jeanne: addressed in 7a70886a33733ee7d1c790c8da6f957ecdcca833

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * [CHANGE] Notifier: Increment the prometheus_notifications_errors_total metric by the number of affected alerts rather than by one per batch of affected alerts. #15428
 * [ENHANCEMENT] OTLP receiver: Convert also metric metadata. #15416
+* [ENHANCEMENT] Introduced MustIndex() as a variant of Index() in tsdb Head that does not return an error. #811
```
colega, line 7: We don't update the changelog in `mimir-prometheus`, please remove this, otherwise it will be just a source of merge conflicts in the future.

colega (approved):

colega:
Approved modulo the changelog nit, please address it before merging.

PR #820, opened 2025-01-15
Include cache entry timestamp in trace events emitted by postings for matchers cache

charleskorn:
This makes it easier to understand the age of a cached entry, which is useful when debugging the behaviour of the cache.


tsdb/postings_for_matchers_cache_test.go
```diff
@@ -270,7 +270,7 @@ func TestPostingsForMatchersCache(t *testing.T) {
 	t.Run("cached value is evicted because cache exceeds max bytes", func(t *testing.T) {
 		const (
 			maxItems         = 100 // Never hit it.
-			maxBytes         = 1250
+			maxBytes         = 1300
```
charleskorn, line 273: Note to reviewers: this change is required because the extra field on `postingsForMatcherPromise` increases its size.

narqo (approved):
🔥 Works for me

PR #822, opened 2025-01-17
Check cached postings TTL before returning from cache + metrics

pracucci:
In this PR I propose two changes to PostingsForMatchers cache:

1. Check if the TTL for cached postings is still valid before returning it from cache, to fix a race condition that could happen between a goroutine running `expire()` and another one skipping the `expire()` execution because it's already in progress.
2. Add metrics to PostingsForMatchers cache, to have better visibility over it. This is something I wanted to do since a long time. The design I picked is to allow to pass the metrics struct as DB options, so that we can use 1 single struct for all per-tenant TSDBs in a Mimir ingester.

The following benchmark shows the difference betweeen:

- `01bb37aae`: the commit before https://github.com/grafana/mimir-prometheus/pull/734
- `6c2603082`: the commit at https://github.com/grafana/mimir-prometheus/pull/734
- `main`
- This PR

```
goos: darwin
goarch: arm64
pkg: github.com/prometheus/prometheus/tsdb
cpu: Apple M3 Pro
                                                          │ 01bb37aae.txt │           6c2603082.txt            │              main.txt              │               pr.txt               │
                                                          │    sec/op     │   sec/op     vs base               │   sec/op     vs base               │   sec/op     vs base               │
PostingsForMatchersCache/no_evictions-11                      594.8n ± 3%   584.2n ± 1%   -1.78% (p=0.015 n=6)   602.9n ± 3%        ~ (p=0.589 n=6)   644.6n ± 1%   +8.38% (p=0.002 n=6)
PostingsForMatchersCache/high_eviction_rate-11                10.52µ ± 5%   10.39µ ± 1%   -1.28% (p=0.002 n=6)   10.85µ ± 1%        ~ (p=0.065 n=6)   10.77µ ± 0%        ~ (p=0.065 n=6)
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11    1411.5n ± 2%   301.1n ± 1%  -78.67% (p=0.002 n=6)   306.2n ± 1%  -78.31% (p=0.002 n=6)   331.2n ± 2%  -76.54% (p=0.002 n=6)
geomean                                                       2.067µ        1.222µ       -40.86%                 1.260µ       -39.03%                 1.320µ       -36.15%

                                                          │ 01bb37aae.txt │             6c2603082.txt             │               main.txt                │               pr.txt                │
                                                          │     B/op      │     B/op      vs base                 │     B/op      vs base                 │     B/op      vs base               │
[...]

PR #825, opened 2025-01-27
Track PostingsForMatchers cached promise completion timestamp in trace

pracucci:
In https://github.com/grafana/mimir-prometheus/pull/822 @charleskorn suggested to log the `PostingsForMatchers` cache promise completion timestamp in the trace. I'm doing it in this PR. However, there's a small performance regression introduced by the extra span event (using epoch timestamp to reduce the regression):

```
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                                                          │ before.txt  │             after.txt              │
                                                          │   sec/op    │   sec/op     vs base               │
PostingsForMatchersCache/no_evictions-11                    637.3n ± 6%   705.8n ± 1%  +10.74% (p=0.002 n=6)
PostingsForMatchersCache/high_eviction_rate-11              10.67µ ± 1%   10.78µ ± 0%   +1.02% (p=0.002 n=6)
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11   327.2n ± 2%   361.6n ± 5%  +10.51% (p=0.002 n=6)
geomean                                                     1.306µ        1.401µ        +7.33%

                                                          │     B/op     │     B/op      vs base               │
PostingsForMatchersCache/no_evictions-11                      974.0 ± 0%    1078.0 ± 0%  +10.68% (p=0.002 n=6)
PostingsForMatchersCache/high_eviction_rate-11              26.32Ki ± 0%   26.42Ki ± 0%   +0.39% (p=0.002 n=6)
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11    1023.5 ± 0%    1131.5 ± 0%  +10.55% (p=0.002 n=6)
geomean                                                     2.925Ki        3.132Ki        +7.09%

                                                          │ before.txt │             after.txt             │
                                                          │ allocs/op  │ allocs/op   vs base               │
PostingsForMatchersCache/no_evictions-11                    20.00 ± 0%   23.00 ± 0%  +15.00% (p=0.002 n=6)
PostingsForMatchersCache/high_eviction_rate-11              46.00 ± 0%   49.00 ± 0%   +6.52% (p=0.002 n=6)
PostingsForMatchersCache_ConcurrencyOnHighEvictionRate-11   20.00 ± 0%   24.00 ± 4%  +20.00% (p=0.002 n=6)
geomean
```

### DIFF
--- a/tsdb/postings_for_matchers_cache.go
+++ b/tsdb/postings_for_matchers_cache.go
@@ -187,6 +187,12 @@ func (p *postingsForMatcherPromise) result(ctx context.Context) (index.Postings,
 		if p.err != nil {
 			return nil, fmt.Errorf("postingsForMatchers promise completed with error: %w", p.err)
 		}
+
+		trace.SpanFromContext(ctx).AddEvent("completed postingsForMatchers promise", trace.WithAttributes(
+			// Do not format the timestamp to introduce a performance regression.
+			attribute.Int64("evaluation completed at (epoch seconds)", p.evaluationCompletedAt.Unix()),
+		))
+
 		return p.cloner.Clone(), nil
 	}
 }
@@ -220,7 +226,6 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Contex
 		promise.callersCtxTracker.close()
 
 		oldPromise := oldPromiseValue.(*postingsForMatcherPromise)
-		oldPromiseEvaluationCompletedAt := time.Time{}
 
 		// Check if the promise already completed the execution and its TTL has not expired yet.
 		// If the TTL has expired, we don't want to return it, so we just recompute the postings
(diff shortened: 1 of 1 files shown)


tsdb/postings_for_matchers_cache.go
```diff
@@ -257,6 +262,7 @@ func (c *PostingsForMatchersCache) postingsForMatchersPromise(ctx context.Contex
 
 		c.metrics.hits.Inc()
 		span.AddEvent("using cached postingsForMatchers promise", trace.WithAttributes(
+			attribute.Stringer("cached promise evaluation completed at", evaluationCompletedAtTime(oldPromiseEvaluationCompletedAt)),
```
charleskorn, line 265: It'd be really nice if we were able to always include the timestamp somewhere, as that will make it much easier to confirm exactly which cache entry is being used.

Perhaps we could log `evaluationCompletedAt` in `postingsForMatchersPromise.result()`?

charleskorn (approved):
LGTM - feedback below is not blocking if you're happy with this as-is.

> However, there's a small performance regression because `attribute.Stringer()` is not lazily evaluated, so the timestamp get stringified for every single call, even if that call is not traced.

What if we passed the timestamp as an integer (in seconds / milliseconds / whatever since the epoch), and used 0 to indicate "not completed"?

tsdb/postings_for_matchers_cache.go
pracucci: Moved to `result()`. There's still a regression using epoch, but a smaller one (I've updated the benchmark in the PR description).

pracucci:
@charleskorn I'm merging this after having addressed your comment. If you have any other (post-merge) comment, I will promptly address it.

PR #833, opened 2025-03-02
Sync upstream Prometheus at 6ede900

charleskorn:
This PR steps us closer to being in sync with the latest upstream changes.

Diff: https://github.com/prometheus/prometheus/compare/2830cba...6ede900

This PR only includes a single upstream PR: https://github.com/prometheus/prometheus/pull/15851.

The main potentially problematic change is the new default value of `model.NameValidationScheme`, which has switched to `UTF8Validation`. I believe this change is safe to bring into Mimir, as we already set `model.NameValidationScheme` to `LegacyValidation` in multiple places ([1](https://github.com/grafana/mimir/blob/71aca2aa6395fe43ed4d79ca7bfec46613035731/pkg/distributor/distributor.go#L68), [2](https://github.com/grafana/mimir/blob/97d684efdc467ed67e245af941106987e7bdb2a2/pkg/frontend/querymiddleware/request_validation.go#L15), [3](https://github.com/grafana/mimir/blob/39dd512ca5168fc7d7522c166713b831d00807ae/pkg/mimir/promexts.go#L14)) that should cover all components that need to validate label names.


jhesketh:
> The main potentially problematic change is the new default value of `model.NameValidationScheme`

Where is the default value being changed?

charleskorn:
> > The main potentially problematic change is the new default value of `model.NameValidationScheme`
> 
> Where is the default value being changed?

As of this PR, Prometheus brings in `prometheus/common` v0.62.0, which includes https://github.com/prometheus/common/pull/724.

jhesketh (approved):
Thanks!

lgtm

PR #838, opened 2025-03-06
TSDB: keep duplicate series records in checkpoints while their samples may still be present

pr00se:
cherry-pick of https://github.com/prometheus/prometheus/pull/16060


charleskorn (approved):
🚀

PR #843, opened 2025-03-18
ruler notifier: make batch size configurable

dimitarvdimitrov:
We have a theory that increasing the batch size will help with throughput to the alertmanager. Pushing this so I can test


dimitarvdimitrov:
closing in favour of https://github.com/prometheus/prometheus/pull/16254

PR #861, opened 2025-03-27
Log instances where WAL replay detect unknown series references, but the series records are found later in the WAL segment

pr00se:
It seems that sometimes we end up with sample records written to the WAL before their corresponding series record. Because the WAL is processed in-order, these leads to the replay reporting unknown series references. It's unclear whether this explains ALL of the instances of unknown series references we see in our cluster, so this PR adds some logging/metrics to let us determine that.

I'm making this change in mimir-prometheus rather than prometheus itself because it's purely a diagnostic thing for now. Impact should be minimal, as this only runs when we already know we have unknown series refs.

### DIFF
--- a/tsdb/head_wal.go
+++ b/tsdb/head_wal.go
@@ -467,6 +467,24 @@ Outer:
 			"tombstones", unknownTombstoneRefs.Load(),
 		)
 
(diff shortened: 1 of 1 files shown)


francoposa (approved):
sure, why not

PR #864, opened 2025-04-03
Log instances where WAL replay detect unknown series references, but the series records are found in a later WAL segment

pr00se:
This builds on https://github.com/grafana/mimir-prometheus/pull/861 to also track if a sample record is written before its series record into a completely different WAL segment. Hopefully this will explain the remaining instances of unknown series refs we see.

I'm keeping this to mimir-prometheus because it is still just diagnostic information.

### DIFF
--- a/tsdb/head_wal.go
+++ b/tsdb/head_wal.go
@@ -76,7 +76,7 @@ func counterAddNonZero(v *prometheus.CounterVec, value float64, lvs ...string) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


tsdb/head_wal.go
```diff
@@ -480,12 +512,21 @@ Outer:
...
+				toDelete = append(toDelete, walRef)
 			}
 		}
+		for _, walRef := range toDelete {
+			delete(unknownSeriesRefs.refs, walRef)
+		}
+		unknownSeriesRefs.mtx.Unlock()
+
+		// Merge missing series refs in this segment into the global list.
+		globalMissingSeriesRefs.merge(unknownSeriesRefs.refs)
```
colega, line 524: I assume this will take the mutex again.

colega (approved):

tsdb/head_wal.go
pr00se: it does [here](https://github.com/grafana/mimir-prometheus/blob/main/tsdb/head_wal.go#L62-L63), yes.

PR #877, opened 2025-05-23
Trigger mimir workflow on branch pushes

dimitarvdimitrov:
_In draft until I have tested this; more detailed description will follow too_


test-file
pracucci, line 1: Remember to remove this test file.

pracucci (approved):
I trust it works. If doesn't, we'll fix it :)

test-file
dimitarvdimitrov: ooops, yes

PR #926, opened 2025-07-03
[TESTS] Labels: Simplify TestByteSize 

bboreham:
Remove lots of duplicated code, in favour of a small table of results for each build.

Cherry-picks one commit from https://github.com/prometheus/prometheus/pull/16807, which is the basis for the simplification.

Also fix `model/labels/labels_slicelabels_test.go` which excluded from compilation.


aknuds1 (approved):
LGTM, thanks!

PR #935, opened 2025-07-10
model: add time tracking to Matcher

dimitarvdimitrov:
> This is an internal version of https://github.com/prometheus/prometheus/pull/16853 so that we can move forward with the work while waiting for the upstream PR to be accepted.

### Background

In grafana/mimir we sometimes see very expensive regular expressions. Unfortunately, apart from profiling and profiling recording rules, there is no insight into how impactful these regular expressions are.

### What this PR does

Adds `NewMatcherWithTimeTracker` which increments an `*atomic.Duration` when the matcher is executed. This helps to put the time of regular expressions into perspective. For example, we can use it to measure the portion of time of serving a query which went to running regular expressions.

### Sampling executions

This PR adds sampling to the execution tracking. The reason is that tracking every execution can lead to significant time spent in `time.Now()` and `time.Since()` and atomic operations. Sampling linearly still has the same effects.

Instead of sampling every 64 or 128 invocation, we sample the 1st, 2nd, 4th, 8th, 16th, 32nd... This helps a lot. We get roughly 20% slowdown with linear sampling compared to exponential sampling.

### Data races

Sampling in a thread-safe way is expensive, so I chose to not support concurrent uses of matchers. In reality we don't use matchers concurrently in a single request in Mimir, so this shouldn't be a problem.

They're not goroutine-safe, but not across cached versions of the matchers: retrieving the same matcher from the cache and using it concurrently is safe. Using the same matcher retrieved only once from the cache concurrently is not safe.


chencs (commented):
Left some comments; approach looks good. Do you have an idea of how expensive it is to measure, e.g., all executions vs every 64th?

dimitarvdimitrov:
> > Do you have an idea of how expensive it is to measure, e.g., all executions vs every 64th?

So I actually decided to run some benchmarks to prove the point of this, that having it hardcoded is better and to show the impact of time tracking. 

I pushed the benchmark and the results 665ecab6d66c8a11ee6fc7c18c200bbfe5c67c90 

### Hardcoded power of 2 constant vs dynamic variable

It looks like having it configurable versus having it hardcoded doesn't have an effect on regular expression evaluations.

```
benchstat -filter '/withDynamicSampling:(false OR true)' -row /sampleRate -col /withDynamicSampling benchmarks.txt  2>/dev/null
goos: darwin
goarch: arm64
pkg: github.com/prometheus/prometheus/model/labels
cpu: Apple M1 Pro
        │     false     │                 true                  │
        │    sec/op     │    sec/op      vs base                │
64        166.7n ± 5% ¹   165.2n ± 5% ¹       ~ (p=0.592 n=378)
128       159.2n ± 6% ¹   157.2n ± 6% ¹       ~ (p=0.183 n=378)
geomean   162.9n          161.2n         -1.02%
¹ benchmarks vary in .fullname
```

I think this just makes it obvious that we should have it configurable.


### Time tracking vs without

It does look like time-tracking is adding 34% overhead to regular expression matching :(


```
benchstat -filter '/withTimeTracking:(false OR true)' -row . -col /sampleRate,/withTimeTracking benchmarks.txt  2>/dev/null
goos: darwin
goarch: arm64
pkg: github.com/prometheus/prometheus/model/labels
cpu: Apple M1 Pro
 │      64       │                  128                  │                                    │
 │     true      │                 true                  │               false                │
 │    sec/op     │    sec/op      vs base                │    sec/op      vs base             │
   166.8n ± 5% ¹   158.5n ± 5% ¹  -4.98% (p=0.000 n=770)   109.4n ± 9% ¹  -34.44% (n=770+378)
¹ benchmarks vary in .fullname
```



The overhead is more than I expected and I think it might be significant in deployments where regular expressions take up a double-digit % of CPU

dimitarvdimitrov:
this is too expensive. A more practical approach is using labels on pprof profiles and profiles recording rules with pyroscope (see grafana/mimir#12404)

PR #936, opened 2025-07-13
Unregister metrics emitted by `remote.WriteStorage` when closed

charleskorn:
This PR modifies the behaviour of `remote.WriteStorage.Close()` to unregister all metrics emitted.

This allows stopping and recreating a `remote.WriteStorage` without causing a `duplicate metrics collector registration attempted` error. This is useful in GEM's remote write functionality for rules, where a rule may have a remote write URL added, removed or changed without restarting the ruler.

### DIFF
--- a/storage/remote/write.go
+++ b/storage/remote/write.go
@@ -268,6 +268,14 @@ func (rws *WriteStorage) Close() error {
 		q.Stop()
 	}
 	close(rws.quit)
+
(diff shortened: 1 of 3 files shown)


Konstantinov-Innokentii:
Just wondering—why do we want to have this as a separate PR instead of waiting for the upstream one to be merged?

PR #940, opened 2025-07-16
labels: Cost & selectivity estimate for Matcher

dimitarvdimitrov:
A very rough sketch of cost estimation for regular expressions. These are being put in a separate file so that they do not conflict with upstream pulls and it's easier to upstream them later.

I've also added tests so that we can track as we make improvements to the model. The values are taken from an environment at GL. These are the most commonly queried matchers in a 24h period by matcher type, plus some exceptionally long matchers.

The TODOs should be addressed later as part of https://github.com/grafana/mimir/issues/11920

### Regular expressions

Taken from https://github.com/grafana/mimir-prometheus/compare/regexp-cost-estimate by @bboreham:

> As part of query planning, estimate the cost of a regexp.
> Very rough and ready, just to get an idea how this might work.
> Add to benchmark so you can compare the estimate against actual
> performance.
>
> Note this is only looking at the Go stdlib implementation; it should
> also make an estimate for those that take the fast path.
>
> Since it looks at `syntax.Regexp` we would either need to do cost
> estimation as part of `newFastRegexMatcher` or retain the pointer so
> we can do it later. Re-parsing every time we want a cost estimate seems
> too expensive.

related to https://github.com/grafana/mimir/issues/11919


model/labels/cost.go
```diff
@@ -0,0 +1,132 @@
...
+func (m *Matcher) EstimateSelectivity(totalLabelValues uint64) float64 {
+	if totalLabelValues == 0 {
+		return 1.0
+	}
+	var selectivity float64
+	// First estimate the selectivity of the operation without taking into account whether it's an inclusive or exclusive matcher.
+	switch m.Type {
+	case MatchEqual, MatchNotEqual:
+		if m.Value == "" {
+			selectivity = 0
```
chencs, line 59: Wouldn't a type of `MatchNotEqual` and `m.Value == ""` result in a selectivity of 1?

model/labels/cost.go
```diff
@@ -0,0 +1,132 @@
...
+			selectivity = 1.0 / float64(totalLabelValues)
+		}
+
+	case MatchRegexp, MatchNotRegexp:
+		// If we have optimized set matches, we know exactly how many values we'll match.
+		// We assume that all of them will be present in the corpus we're testing against.
+		switch setMatchesSize := len(m.re.setMatches); {
+		case setMatchesSize > 0:
+			selectivity = float64(setMatchesSize) / float64(totalLabelValues)
+		case m.Value == "":
```
chencs, line 71: Same as above comment for `MatchNotEqual` applies to `MatchNotRegexp`

chencs: Disregard, I see later that we flip the selectivity for exclusive matchers.

model/labels/cost.go
```diff
@@ -0,0 +1,132 @@
...
+	parsed, err := syntax.Parse(m.reString, syntax.Perl|syntax.DotNL)
+	if err != nil {
+		return 0
+	}
+	return costEstimate(parsed)
+}
+
+// TODO this doesn't account for backtracking, which can come with a large cost.
+func costEstimate(re *syntax.Regexp) float64 {
+	switch re.Op {
```
chencs, line 102: Where did these numbers come from (particularly all the `1`s)? If the cases that return `1` are just so because they're defaults, why not just let them go to the `default` case?

model/labels/cost_test.go
```diff
@@ -0,0 +1,162 @@
...
+		{8, 0.875, "name", MatchNotRegexp, "up"},
+
+		{8, 0.0, "name", MatchRegexp, ""},
+		{8, 1.0, "name", MatchNotRegexp, ""},
+		{8, 0.0, "name", MatchEqual, ""},
+		{8, 1.0, "name", MatchNotEqual, ""},
+
+		{8, 1.0, "name", MatchRegexp, ".+"},
+		{8, 0.0, "name", MatchNotRegexp, ".+"},
+		{8, 0.1, "name", MatchRegexp, ".*(foo|bar).*"},
```
chencs, line 144: Same question here as above -- should we add a test case for `.*foo.*|.*bar.*`?

model/labels/cost_test.go
```diff
@@ -0,0 +1,162 @@
...
+		{1, "__name__", MatchEqual, "adaptive_metrics_canary_agg"},
+		{1, "__name__", MatchEqual, "loki_distributor_bytes_received_total"},
+		{1, "__name__", MatchEqual, "cortex_lifecycler_read_only"},
+		{1, "cluster", MatchEqual, "ops-eu-south-0"},
+		{1, "job", MatchEqual, "integrations/db-o11y"},
+		{1, "__name__", MatchEqual, "up"},
+		{1, "namespace", MatchEqual, "hosted-grafana"},
+		{1, "namespace", MatchEqual, "grafana-com"},
+
+		// =~ matchers
```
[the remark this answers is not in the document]
chencs: Kind of in the same vein, though maybe tangential -- how do we want to handle somewhat nonsensical but still valid regex matchers like `.*.*`? (This is a pattern that is somewhat common despite being redundant)

model/labels/regexp_test.go
```diff
@@ -408,6 +408,7 @@ func BenchmarkFastRegexMatcher(b *testing.B) {
 			require.NoError(b, err)
 
 			b.ResetTimer()
+			b.ReportMetric(m.SingleMatchCost(), "cost")
```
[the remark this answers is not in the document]
dimitarvdimitrov: what do you mean? this is just reported in the output of the benchmark in case we want to adjust the cost model later on

model/labels/cost_test.go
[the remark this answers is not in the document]
chencs: I asked about the `.*value1.*|.*value2.*` because of the backtracking, and while I know there's already a comment above about not accounting for backtracking, it seems like a good thing to at least acknowledge in tests and then if we change estimation to account for backtracking it'll be obvious.

I asked about `.*.*` because it's another pattern I see commonly for some reason, but I realize now it doesn't really exhibit the same backtracking behavior as the first pattern, so maybe it's not worth thinking about for now.

PR #947, opened 2025-07-24
tsdb index: introduce scan matchers

dimitarvdimitrov:
## Summary

This PR integrates the scan matchers feature from prometheus/prometheus#16835 into our mimir-prometheus fork.

### **Background**

In the Prometheus TSDB, matchers are always used to query the index. There are some cases where skipping those matchers from the index and then applying them on the retrieved series from the block is less resource intensive. 

This is the plan we have in Mimir to solve this: https://github.com/grafana/mimir/issues/11916

#### Interfaces

In this PR I'm introducing the concept of **scan matchers**. Unlike to **index matchers** (aka just "matchers" today), scan matchers are used to filter out series after they have been retrieved from the block.

The core of this PR adds a new method to the `IndexReader`, called `IndexLookupPlanner() index.LookupPlanner`. An `index.LookupPlanner` partitions the matchers into matchers which will be used against the inverted index and matchers which will be used in a sequential scan after retrieving the labels of the matching series.

The `LookupPlanner` is an interface so that in downstream projects can implement their own planners based on their storage characteristics and based on the target performance. They can also chain multiple planners via the `index.ChainLookupPlanners`. 

The `LookupPlan` is currently implemented as an interface because later down the road there will be a need for different kind of plans (e.g. for label names and label values; cost-based vs static).

#### `ScanEmptyMatchersLookupPlanner`

There are some matchers, which are expensive to apply on the inverted index and usually don't end up filtering any data. This PR adds one implementation of `LookupPlanner`: `ScanEmptyMatchersLookupPlanner`, which defers those matchers to scan matchers and removes some matchers when they match literally every possible value, even the unset value.

- `{label=""}` - scan matcher
- `{label=~".+"}` - scan matcher
- `{label=~".*"}` - removed matcher

### **Next steps**

The goal now is to have very basic support for this in Prometheus, which would allow experimentation with different planner algorithms. 

We already have some planner implementation which is based on sketches collected from the head block. This is another building block which we can contribute.

### DIFF
--- a/tsdb/index/lookup_test.go
+++ b/tsdb/index/lookup_test.go
@@ -124,7 +124,17 @@ func TestScanEmptyMatchersLookupPlanner(t *testing.T) {
 			expectedScan: []*labels.Matcher{},
 		},
 		{
-			name: "single empty matcher goes to index",
+			name: "single match-non-empty regex goes to index",
+			matchers: []*labels.Matcher{
+				labels.MustNewMatcher(labels.MatchRegexp, "job", ".+"),
+			},
+			expectedIndex: []*labels.Matcher{
+				labels.MustNewMatcher(labels.MatchRegexp, "job", ".+"),
+			},
+			expectedScan: []*labels.Matcher{},
+		},
+		{
+			name: "single empty matcher goes to index; this is an invalid matcher on the API, but some internal code uses it",
 			matchers: []*labels.Matcher{
 				labels.MustNewMatcher(labels.MatchEqual, "", ""),
 			},
--- a/tsdb/index/lookup_test.go
+++ b/tsdb/index/lookup_test.go
@@ -1,0 +1,166 @@
+// Copyright 2025 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package index
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/prometheus/prometheus/model/labels"
+)
+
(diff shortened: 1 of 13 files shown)


tsdb/index/lookup_test.go
```diff
@@ -0,0 +1,166 @@
...
+				labels.MustNewMatcher(labels.MatchEqual, "job", "prometheus"),
+				labels.MustNewMatcher(labels.MatchNotEqual, "status", "down"),
+			},
+			expectedScan: []*labels.Matcher{
+				labels.MustNewMatcher(labels.MatchEqual, "env", ""),
+				labels.MustNewMatcher(labels.MatchRegexp, "region", ".+"),
+			},
+		},
+		{
+			name: "all selective matchers go to index",
```
chencs, line 53: I liked that in the `"splits matchers between index and scan"` test, you labeled what would happen to each of the input matchers in the comments.

tsdb/ooo_head_read.go
```diff
@@ -528,6 +534,12 @@ func (ir *OOOCompactionHeadIndexReader) Close() error {
 	return nil
 }
 
+// IndexLookupPlanner returns the index lookup planner for this reader.
+// OOO compaction head readers use the default index-only planner.
```
chencs, line 538: I find this phrasing a little confusing. Isn't the default planner `ScanEmptyMatchersLookupPlanner` ([code](https://github.com/grafana/mimir-prometheus/blob/f3d5fac5b0f1cc60726c9bf523b2ecd78c86caef/tsdb/head.go#L238))? So wouldn't `PlanIndexLookup()` potentially result in some scan matchers?

tsdb/index/lookup_test.go
```diff
@@ -0,0 +1,166 @@
...
+			matchers: []*labels.Matcher{
+				labels.MustNewMatcher(labels.MatchRegexp, "job", ".*"),
+			},
+			expectedIndex: []*labels.Matcher{
+				labels.MustNewMatcher(labels.MatchRegexp, "job", ".*"),
+			},
+			expectedScan: []*labels.Matcher{},
+		},
+		{
+			name: "single empty matcher goes to index",
```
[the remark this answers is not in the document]
dimitarvdimitrov: 👍  added in https://github.com/grafana/mimir-prometheus/pull/947/commits/8e61309f520915bcecee1251d94462d0b6d79cec

tsdb/ooo_head_read.go
dimitarvdimitrov: oh i think this was just leftover from before it was even configurable

chencs (approved):

PR #962, opened 2025-08-15
Implement periodic head statistics collection

chencs:
Relates to https://github.com/grafana/mimir/issues/11919

This PR implements periodic collection of label value cardinality and other statistics from MemPostings based on `head-statistics-collection-frequency` (0 to disable). These statistics will be used to plan and optimize the execution of Mimir queries.

### DIFF
--- a/tsdb/index/postings_sketches.go
+++ b/tsdb/index/postings_sketches.go
@@ -7,6 +7,8 @@ import (
 	boom "github.com/tylertreat/BoomFilters"
 )
 
+const countMinEpsilon = 0.005
+
 type Statistics interface {
 	// TotalSeries returns the number of series in the TSDB component.
 	TotalSeries() uint64
@@ -72,7 +74,7 @@ func (p *MemPostings) labelValuesSketchForLabelName(name string) *LabelValuesSke
 	}
 
 	sketch := LabelValuesSketch{
-		s:              boom.NewCountMinSketch(0.01, 0.01),
+		s:              boom.NewCountMinSketch(countMinEpsilon, 0.01),
 		distinctValues: uint64(len(labelValuesPostings)),
 	}
 	for value, postings := range labelValuesPostings {
--- a/tsdb/index/postings_sketches.go
+++ b/tsdb/index/postings_sketches.go
@@ -1,0 +1,113 @@
+package index
+
+import (
+	"context"
+	"fmt"
+	"time"
+	"unsafe"
+
+	boom "github.com/tylertreat/BoomFilters"
+)
+
+type Statistics interface {
+	TotalSeries() uint64
+	LabelValuesCount(ctx context.Context, name string) (uint64, error)
+	LabelValuesCardinality(ctx context.Context, name string, values ...string) (uint64, error)
+}
+
+// LabelsValuesSketches contains count-min sketches of the values for each label name.
+// It implements index.Statistics, which can be used to inform query plan generation.
+type LabelsValuesSketches struct {
+	labelNames map[string]*LabelValuesSketch
+}
+
+type LabelValuesSketch struct {
+	s              *boom.CountMinSketch
+	distinctValues uint64
+	lastUpdated    int64 // epoch time of last update
+}
+
+// LabelsValuesSketches builds a LabelsValuesSketches based on the label names and values
+// currently in MemPostings. The postings mutex (p.mtx) is taken independently to generate a list of label names,
+// and then on a per-label-name basis to produce a count-min sketch for each label name,
+// because this process should be interruptable by, e.g., head compaction.
+// If MemPostings is updated between generation of the label name list and sketch generation, that some new label
+// names will be missed.
+func (p *MemPostings) LabelsValuesSketches() LabelsValuesSketches {
+	names := p.LabelNames()
+	labelNames := make(map[string]*LabelValuesSketch, len(names))
+	for _, name := range names {
+		labelNames[name] = p.labelValuesSketchForLabelName(name)
+	}
+	return LabelsValuesSketches{
+		labelNames: labelNames,
+	}
+}
+
+// labelValuesSketchForLabelName returns a count-min sketch of distinct values for a given label name.
+// If name is not in MemPostings, nil is returned.
+func (p *MemPostings) labelValuesSketchForLabelName(name string) *LabelValuesSketch {
+	if name == "" {
+		return nil
+	}
+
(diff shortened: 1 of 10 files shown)


dimitarvdimitrov (commented):
leaving some comments, i think we should agree on all of them before merging.

i did fix up what i think are the most critical ones in this diff. feel free to borrow from there https://github.com/grafana/mimir-prometheus/compare/casie/collect-head-statistics...dimitar/collect-head-statistics

tsdb/index/postingsstats.go
```diff
@@ -74,3 +80,97 @@ func (m *maxHeap) get() []Stat {
 	})
 	return m.Items
 }
+
+// LabelsValuesSketches contains count-min sketches of the values for each label name.
+// It implements index.Statistics, which can be used to inform query plan generation.
+type LabelsValuesSketches struct {
+	labelNames          map[string]*LabelValuesSketch
+	updateLastCompleted int64
```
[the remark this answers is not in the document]
chencs: Good point; I guess we don't need it here. There _is_ also a per-sketch `lastUpdated` that we don't currently use. I thought it might be useful later for whatever reason, but given that we're reporting the time it takes to update stats, I'm not sure it'll actually become useful

tsdb/index/postingsstats.go
```diff
@@ -74,3 +80,97 @@ func (m *maxHeap) get() []Stat {
...
+	return LabelsValuesSketches{
+		labelNames:          labelNames,
+		updateLastCompleted: time.Now().Unix(),
+	}
+}
+
+// labelValuesSketchForLabelName returns a count-min sketch of distinct values for a given label name.
+// If name is not in MemPostings, nil is returned.
+func (p *MemPostings) labelValuesSketchForLabelName(name string) *LabelValuesSketch {
+	if name == "" {
```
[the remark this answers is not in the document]
chencs: Like why an empty `name` could be passed at all? I don't know, but `addFor()` (nor its callers afaict) doesn't really check for label names being empty strings.

I guess you could make an argument that if an empty-string label name exists, we should create a sketch for it, but that seems a bit like wasted work to me (as does adding them to `m` at all).

tsdb/index/postingsstats.go
```diff
@@ -74,3 +80,97 @@ func (m *maxHeap) get() []Stat {
...
+	}
+
+	return &sketch
+}
+
+// LabelValuesCount returns the number of values for a given label name.
+func (lvs *LabelsValuesSketches) LabelValuesCount(_ context.Context, name string) (uint64, error) {
+	sketch, ok := lvs.labelNames[name]
+	if !ok {
+		return 0, fmt.Errorf("no sketch found for label %q", name)
```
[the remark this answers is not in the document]
chencs: It's fair that 0 is a valid return value. I'm not sure if it's possible for a label name to have a 0-length slice in postings; I guess it _probably_ doesn't happen, but if it did, would we want to distinguish between "the number of distinct values is 0" vs. "we don't have a sketch for this at all"?

tsdb/index/postings_sketches.go
dimitarvdimitrov, line 1: sorry i missed this to begin with but we need tests for this code. i can help drafting some

dimitarvdimitrov (commented):
tests look good. agree that it would be nice to have higher volume test cases

a vague idea is to use fuzzing to discover some edge cases, but i can't think of a way to use fuzzing because of the complicated test setup (map of series ID to labels)

tsdb/index/postings_sketches.go
```diff
@@ -72,7 +74,7 @@ func (p *MemPostings) labelValuesSketchForLabelName(name string) *LabelValuesSke
 	}
 
 	sketch := LabelValuesSketch{
-		s:              boom.NewCountMinSketch(0.01, 0.01),
+		s:              boom.NewCountMinSketch(countMinEpsilon, 0.01),
```
dimitarvdimitrov, line 77: this would double the size of the sketch, right? can we somehow isolate the more precise size to tests only?

tsdb/index/postings_sketches_test.go
```diff
@@ -198,3 +196,81 @@ func TestLabelsValuesSketches_LabelValue(t *testing.T) {
...
+	s := p.LabelsValuesSketches()
+
+	require.Equal(t, uint64(numLabelValues), s.LabelValuesCount(ctx, labelName))
+	require.Equal(t, uint64(numSeries), s.LabelValuesCardinality(ctx, labelName))
+
+	for i := 0; i < numLabelValues; i++ {
+		// The cardinality for every label should be within epsilon of the total number of series to the expected cardinality.
+		// Technically, it should be within epsilon of the total increments seen by the count-min sketch,
+		// but that's more opaque to understand. The total increments seen will always be equal or greater than the number of series.
+		require.InDeltaf(t, uint64(numSeries/numLabelValues), s.LabelValuesCardinality(ctx, labelName, strconv.Itoa(i)),
```
dimitarvdimitrov, line 226: nit: doesn't the `require.InEpsilonf` do what you need?

tsdb/index/postings_sketches_test.go
```diff
@@ -198,3 +196,81 @@ func TestLabelsValuesSketches_LabelValue(t *testing.T) {
...
+
+	ctx := context.Background()
+	s := p.LabelsValuesSketches()
+
+	lowValCard := s.LabelValuesCardinality(ctx, labelName, lowCardValue)
+
+	// The cardinality of every other value should be ≥6000. We care about these values being correct in magnitude,
+	// i.e., floor(log(highOccurrenceCardinality) / log(lowOccurrenceCardinality)) should be consistent every time.
+	// We add a little margin since the margin of error (30k) is enough to push us one power up, but not two.
+	for i := 0; i < numHighOccurrenceValues; i++ {
```
dimitarvdimitrov, line 270: here shouldn't we also assert on the cardinalities of `lowCardValue`?

dimitarvdimitrov: no

dimitarvdimitrov (approved):

tsdb/index/postings_sketches_test.go
chencs: Nope. `InEpsilonf` is looking for an error percentage based on the expected value. However, count-min sketch error is calculated based on the size of the _total increments_ -- in our case, the total number of series with `labelName`, so the delta is computed on a completely different basis than the expected value.

PR #973, opened 2025-09-04
Add IndexLookupPlannerFunc to TSDB Options

dimitarvdimitrov:
Refactor index lookup planner to use function-based approach instead of static planners.

### Changes

- Replace `IndexLookupPlanner` with `IndexLookupPlannerFunc` in `Options` and `HeadOptions` 
- `OpenBlockWithOptions` takes function, calls it once when block opens, caches result
- Head calls function periodically during stats updates, stores in atomic pointer

### Use Case

1. Create specialized planners based on block metadata (ULID, time range, etc.)
2. Use different statistics for different blocks when making query planning decisions
3. Optimize query execution on a per-block basis rather than using global statistics

Related to https://github.com/grafana/mimir/issues/12432


chencs (approved):
Looks like https://github.com/grafana/mimir/pull/12603 will need updated vendoring and there's a test that needs fixed, but otherwise/overall, this looks good to me.

PR #987, opened 2025-09-24
tsdb: include PostingsClonerFactory

dimitarvdimitrov:
follow-up of #985. Without this change mimir can't inject its own factory

```release-notes
NONE
```


duricanikolic (approved):
LGTM

PR #1125, opened 2026-04-01
tsdb: add block stat for number of series with __series_hash__

56quarters:
Include the number of series that have a `__series_hash__` label in block stats. This allows consumers (Mimir compactor) to determine if the TSDB block can be used with projections.

Part https://github.com/grafana/mimir/issues/13863

### DIFF
--- a/tsdb/async_block_writer.go
+++ b/tsdb/async_block_writer.go
@@ -90,6 +90,10 @@ func (bw *asyncBlockWriter) loop() (res asyncBlockWriterResult) {
 
 		stats.NumChunks += uint64(len(sw.chks))
 		stats.NumSeries++
+		if sw.lbls.Has("__series_hash__") {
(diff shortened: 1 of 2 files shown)


tsdb/async_block_writer.go
```diff
@@ -90,6 +90,10 @@ func (bw *asyncBlockWriter) loop() (res asyncBlockWriterResult) {
 
 		stats.NumChunks += uint64(len(sw.chks))
 		stats.NumSeries++
+		if sw.lbls.Has("__series_hash__") {
```
56quarters, line 93: An alternative approach to using the `labels.Labels` method would be to expose the cardinality from the `IndexWriter` however it would require more changes.

56quarters: Done in https://github.com/grafana/mimir-prometheus/pull/1126

PR #1142, opened 2026-04-17
tsdb: add head-only querier for projections

56quarters:
Add a method for getting a ChunkQuerier for use with projections that only operates on the TSDB head. Based on configuration it will take `__series_hash__` into account when sorting postings and populating series.

### DIFF
--- a/tsdb/head_read.go
+++ b/tsdb/head_read.go
@@ -35,19 +35,20 @@ func (h *Head) ExemplarQuerier(ctx context.Context) (storage.ExemplarQuerier, er
 
 // Index returns an IndexReader against the block.
 func (h *Head) Index() (IndexReader, error) {
-	return h.indexRange(math.MinInt64, math.MaxInt64), nil
+	return h.indexRange(math.MinInt64, math.MaxInt64, false), nil
 }
 
-func (h *Head) indexRange(mint, maxt int64) *headIndexReader {
+func (h *Head) indexRange(mint, maxt int64, includeSeriesHash bool) *headIndexReader {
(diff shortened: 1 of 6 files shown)


tsdb/head_read.go
```diff
@@ -202,7 +226,14 @@ func (h *headIndexReader) Series(ref storage.SeriesRef, builder *labels.ScratchB
 		h.head.metrics.seriesNotFound.Inc()
 		return storage.ErrNotFound
 	}
-	builder.Assign(s.labels())
+
+	if h.head.opts.EnableSeriesHash && !h.includeSeriesHash {
```
56quarters, line 230: We could potentially pass the `SelectHints` down to `headIndexReader` and only populate the series with the relevant labels to avoid having to remove `__series_hash__` here and then drop more labels in the Mimir ingester.

tsdb/head_read.go
```diff
@@ -151,6 +159,22 @@ func (h *headIndexReader) SortedPostings(p index.Postings) index.Postings {
 	return index.NewListPostings(ep)
 }
 
+func compareWithSeriesHash(a, b *memSeries) int {
+	return labels.Compare(a.labels(), b.labels())
+}
+
+func compareWithoutSeriesHash(a, b *memSeries) int {
```
charleskorn, line 166: Not something for the initial version, but I wonder how expensive it'll be to drop the `__series_hash__` label like this - I wonder if it'll be cheaper to introduce a variant of `Compare` that takes a list of labels to ignore during the comparison.

PR #1154, opened 2026-05-11
Add support for compacting a list of custom series

duricanikolic:
#### Which issue(s) does the PR fix:
Part of https://github.com/grafana/mimir-squad/issues/3596

#### Summary

This PR adds `DB.CompactSelectedSeries([]storage.SeriesRef)`, a new compaction entry point that writes an explicit caller-supplied list of head series into one or more blocks (one per chunk-range window) and then evicts those series from the head — without advancing `HeadMinTime` or touching any other series.

The motivation is to let callers (e.g. Mimir's block-builder) decide which series to flush ahead of a normal compaction cycle, independently of whether those series carry a staleness marker.

 #### What changed

##### New public API

- `DB.CompactSelectedSeries(seriesRefs []storage.SeriesRef) error` — compacts the provided refs into blocks and evicts them from the head.
- `NewSelectedSeriesHead` / `SelectedSeriesHead` — generalised replacement for `StaleHead`; used as the restricted `BlockReader` view during both stale-series and selected-series compaction.
- `BlockMetaCompaction.SetSelectedSeries()` / `FromSelectedSeries()` — new compaction hint from-selected-series so the resulting blocks are identifiable and excluded from `inOrderBlocksMaxTime`.

##### Shared compaction logic

`compactHeadViewLocked` is extracted from `CompactStaleHead` and now drives both flows. It accepts a `headViewFactory` (builds the restricted `BlockReader` for a given time window) and a `headSeriesEvictor` (runs after all blocks are written). This removes the duplicated chunk-range loop.

##### Safety guarantees

Series are skipped — and stay in the head — if, at eviction time:
- they have acquired out-of-order data (evicting them before `CompactOOOHead` flushes their OOO chunks would orphan those chunks),
- their `maxTime` has advanced past `maxt` (a sample landed at a timestamp newer than anything captured in the blocks), or
- they have received a sample whose append-ID is strictly greater than a watermark captured before the blocks were written (a sample committed during the block-write/evict window — possibly at a timestamp inside the block range — and therefore not guaranteed to be in any block).

The third check relies on the per-sample append-IDs maintained by TSDB isolation; see the next section.

##### Isolation requirement

`CompactSelectedSeries` is fully correct **only when TSDB isolation is enabled**, which is the default and only behaviour of stock Prometheus: `tsdb.Options.IsolationDisabled` defaults to `false`, and there is no Prometheus CLI flag or configuration setting to disable it. Embedding callers (Mimir, Cortex, Thanos receive, etc.) can disable it explicitly through that option.

If a caller does disable isolation, the watermark guard above becomes a no-op because per-series append-IDs are no longer tracked. The remaining timestamp guard (`series.maxTime() > maxt`) then only catches late writes whose timestamp is past the head's `MaxTime` at compaction start. A late write committed at an in-range timestamp during the block-write/evict window can therefore be evicted along with the series, and the `{MinInt64, MaxInt64}` tombstone logged by the evictor drops it permanently on the next WAL replay. With isolation disabled the caller is responsible for ensuring no concurrent writes target the chosen refs while the compaction is running.

##### Refactors (no behaviour change)

| Old                                                 | New                                               |
|-----------------------------------------------------|---------------------------------------------------|
| `StaleHead`                                         | `SelectedSeriesHead`                              |
| `headStaleIndexReader`                              | `headSelectedSeriesIndexReader`                   |
| `gcStaleSeries`                                     | `gcSeries` + `shouldEvict` predicate              |
| `filterStaleSeriesAndSortPostings` -> `[]SeriesRef` | -> `seriesRefs{sortedByRef, sortedByLabels}`      |
| `SortedStaleSeriesRefsNoOOOData` (exported)         | `staleSeriesRefsNoOOOData`<br> (unexported)       |

`isStaleSeries` and `isSeriesWithoutOOO` are extracted as named predicates passed into `truncateSeries`/ `gcSeries`, making both the stale-series and selected-series paths share the same machinery.

##### New metrics
- `prometheus_tsdb_selected_series_compactions_triggered_total`
-  `prometheus_tsdb_selected_series_compactions_failed_total`
-  `prometheus_tsdb_selected_series_compaction_duration_seconds`

Release Notes:
```release-notes
 [FEATURE] tsdb: Add `DB.CompactSelectedSeries` to compact and evict an explicit caller-supplied list of head series into blocks without advancing HeadMinTime.
```


tsdb/db.go
```diff
@@ -1872,6 +1894,64 @@ func (db *DB) compactHead(head *RangeHead, truncateMemory bool) error {
 	return nil
 }
 
+// HeadPortionDecorator builds a BlockReader view of a portion of the head, restricted to the
+// closed time range [mint, maxt]. Implementations decide which subset of series the view exposes
+// (for example, by filtering on a list of series refs). The decorator is invoked once per
+// chunk-range slice during compaction.
+type HeadPortionDecorator func(head *Head, mint, maxt int64) BlockReader
```
colega: Nit: I don't really like the name. It suggests me that we're going to decorate a portion of the head, but the key here is that we're going to build a restricted view of the Head, isn't it?

tsdb/compact.go
```diff
@@ -787,6 +787,9 @@ func (c *LeveledCompactor) Write(dest string, b BlockReader, mint, maxt int64, b
 		if base.Compaction.FromStaleSeries() {
 			meta.Compaction.SetStaleSeries()
 		}
+		if base.Compaction.FromSelectedSeries() {
+			meta.Compaction.SetSelectedSeries()
+		}
```
colega: Can we just inherit all hints from meta? Why do we keep adding code for each case?

tsdb/db.go
```diff
@@ -1872,6 +1894,64 @@ func (db *DB) compactHead(head *RangeHead, truncateMemory bool) error {
...
+type HeadPortionEvictor func(maxt int64) error
+
+// CompactHeadPortion writes a block (or sequence of blocks, one per chunk range) covering the
+// portion of the head described by decorator, then runs evictor to remove those series from the
+// head. HeadMinTime is not advanced; series outside the portion are unaffected.
+//
+// For example, this engine powers CompactStaleHead. Callers that maintain their own list of
+// series refs (for example, an external ownership tracker) can invoke it directly with their
+// own decorator/evictor pair.
+func (db *DB) CompactHeadPortion(decorator HeadPortionDecorator, evict HeadPortionEvictor, meta *BlockMeta) error {
```
colega: As discussed offline: if this API doesn't need to be exported, let's unexport it.

colega: This also applies to the interfaces it accepts: let's unexport them until they really need to be exported.

tsdb/db.go
```diff
@@ -1872,6 +1894,64 @@ func (db *DB) compactHead(head *RangeHead, truncateMemory bool) error {
...
+	db.cmtx.Lock()
+	defer db.cmtx.Unlock()
+	return db.compactHeadPortionLocked(decorator, evict, meta)
+}
+
+// compactHeadPortionLocked is the body of CompactHeadPortion, intended for internal callers
+// that already hold db.cmtx.
+func (db *DB) compactHeadPortionLocked(decorator HeadPortionDecorator, evict HeadPortionEvictor, meta *BlockMeta) error {
```
colega: It is unclear why the caller would need to provide `meta *BlockMeta`, but it seems that it's only needed to set an optional compaction hint. Maybe we could accept a `configure func(meta *BlockMeta)` instead of delegating the block meta creation to the caller?

tsdb/db.go
```diff
@@ -1894,36 +1974,88 @@ func (db *DB) CompactStaleHead() (err error) {
...
 	meta := &BlockMeta{}
 	meta.Compaction.SetStaleSeries()
-	mint, maxt := db.head.opts.ChunkRange*(db.head.MinTime()/db.head.opts.ChunkRange), db.head.MaxTime()
-	for ; mint < maxt; mint += db.head.chunkRange.Load() {
-		staleHead := NewStaleHead(db.Head(), mint, mint+db.head.chunkRange.Load()-1, staleSeriesRefs)
 
-		uids, err := db.compactor.Write(db.dir, staleHead, staleHead.MinTime(), staleHead.BlockMaxTime(), meta)
+	if err := db.compactHeadPortionLocked(
+		func(h *Head, mint, maxt int64) BlockReader {
+			return NewSelectedSeriesHead(h, mint, maxt, staleSeriesRefs, db.head.filterStaleSeriesAndSortPostings)
```
colega: `db.head.filterStaleSeriesAndSortPostings` is a quite heavy operation and we already called it to get `staleSeriesRefs`, why do we need to invoke it again?

tsdb/head.go
```diff
@@ -1290,8 +1290,37 @@ func (h *Head) truncateMemory(mint int64) (err error) {
...
 // truncateStaleSeries removes the provided series as long as they are still stale.
 func (h *Head) truncateStaleSeries(seriesRefs []storage.SeriesRef, maxt int64) error {
+	return h.truncateSeries(seriesRefs, maxt, isStaleSeries, true)
+}
+
+// truncateSelectedSeries removes the series identified by the provided refs from the head.
+// Series that received fresh samples since the caller collected the ref list are skipped.
+func (h *Head) truncateSelectedSeries(seriesRefs []storage.SeriesRef, maxt int64) error {
+	return h.truncateSeries(seriesRefs, maxt, alwaysEvictSeries, false)
+}
```
colega, line 1339: What happens if a series received an OOO sample while we were writing the block? Let's say we didn't increment the maxt.

This also applies to the existing `truncateStaleSeries`

tsdb/head.go
```diff
@@ -2227,20 +2259,26 @@ func (s *stripeSeries) gc(mint int64, minOOOMmapRef chunks.ChunkDiskMapperRef) (
...
 	seriesRemoved := len(deleted)
 
 	h.metrics.seriesRemoved.Add(float64(seriesRemoved))
 	h.metrics.chunksRemoved.Add(float64(chunksRemoved))
 	h.metrics.chunks.Sub(float64(chunksRemoved))
 	h.numSeries.Sub(uint64(seriesRemoved))
-	h.numStaleSeries.Sub(uint64(seriesRemoved))
+	if wereStale {
+		h.numStaleSeries.Sub(uint64(seriesRemoved))
+	}
```
colega: This looks quite ugly, maybe we could move this to the caller?

tsdb/head_read.go
```diff
@@ -323,9 +327,52 @@ func (h *Head) filterStaleSeriesAndSortPostings(p index.Postings) ([]storage.Ser
...
+			continue
+		}
+
+		s.Lock()
+		hasOOO := s.ooo != nil
+		s.Unlock()
+		if hasOOO {
+			// Has out-of-order data; skip it because we cannot determine if a series
+			// is stale when it's getting out-of-order data.
+			continue
```
colega: This method is for selected series but this comment is talking about stale series.

tsdb/head_read.go
```diff
@@ -336,6 +383,14 @@ func (h *Head) SortedStaleSeriesRefsNoOOOData(ctx context.Context) ([]storage.Se
 	return h.filterStaleSeriesAndSortPostings(h.postings.Postings(ctx, k, v))
 }
 
+// SortedSelectedSeriesRefNoOOOData partitions the given series refs into those whose series
+// do not currently carry out-of-order data (kept) and those that do (skipped). Kept refs are
+// returned sorted by series labels; refs whose series cannot be found in the head are silently
+// dropped.
+func (h *Head) SortedSelectedSeriesRefNoOOOData(seriesRefs []storage.SeriesRef) ([]storage.SeriesRef, error) {
```
colega: Does this need to be exported? Is this method needed at all?

tsdb/db.go
duricanikolic: I am going to rename both `HeadPortionDecorator` and `HeadPortionEvictor` into `headViewFactory` and `headSeriesEvictor`.

tsdb/db.go
duricanikolic: I am going to:
- get rid of `CompactHeadPortion()`
- rename `compactHeadPortionLocked()` into `compactHeadViewLocked()`
- unexport the interfaces the latter accepts

tsdb/head_read.go
duricanikolic: Thank you for this catch. The comment was just copy-pasted.
I am going to replace it with
```
// Skip series with out-of-order data: evicting them before their OOO chunks
// have been flushed by CompactOOOHead would orphan those chunks.
```

tsdb/head.go
duricanikolic: I agree. I will rework the implementation and put the `numStaleSeries` substraction in `Head.truncateStaleSeries()`, where it belongs to.

tsdb/compact.go
duricanikolic: I kept the original pattern that was already present. But you are right, we can just inherit all the hints from the base `BlockMeta`. I am going to implement that.

tsdb/db.go
duricanikolic: I will replace parameter `meta *BlockMeta` parameter with `config func(meta *BlockMeta)`, and move the block meta creation in `DB.compactHeadViewLocked()` (ex `DB.compactHeadPortionLocked()`).

tsdb/db.go
duricanikolic: Hi @colega 
It's not actually invoked again. The answer is in `headSelectedSeriesIndexReader.Postings()`:  
```
func (h *headSelectedSeriesIndexReader) Postings(ctx context.Context, name string, values ...string) (index.Postings, error) {
  k, v := index.AllPostingsKey()
  if len(h.selectedSeriesRefs) > 0 && name == k && len(values) == 1 && values[0] == v {
    return index.NewListPostings(h.selectedSeriesRefs), nil  // fast path
  }
  seriesRefs, err := h.filter(h.head.postings.Postings(ctx, name, values...))  // filter only reached here
  ...
}
```

Compaction only calls AllPostings, so it always hits the fast path and returns the pre-computed `seriesRefs` directly. There is a [comment in the code explicitly stating that](https://github.com/grafana/mimir-prometheus/pull/1154/changes#diff-73c4d729472ae77b03b5d3a6d05f6a924eb1a53d5f248ae11194e8c8cd2ed284R245-R249) (already present in both `mimir-prometheus` and `prometheus`)
```
Compaction callers only request AllPostings, which is served from the precomputed ref list without re-reading the Head.
```

The filter is stored in the index reader and threaded down for the non-AllPostings methods (`PostingsForLabelMatching`, `PostingsForAllLabelValues`), both of which are explicitly commented as "Unused for compaction" (required by the `IndexReader` interface) — so the filter is dead code in this path.

That means that `filterSelectedSeriesAndSortPostings` doesn't need to be passed to `NewSelectedSeriesHead` at all for the `CompactSelectedSeries` flow, and we can pass `nil`.

Again, this code is already a part of both `mimir-prometheus` and `prometheus`, and this change just makes it more general.

tsdb/head_read.go
duricanikolic: We actually don't need that method, and we can directly call `Head.filterSelectedSeriesAndSortPostings`. I just wanted to follow the same pattern `DB.CompactStaleSeries()` follows.

I will remove it.

tsdb/head.go
duricanikolic: This is a great catch. I silently relied on the fact that `CompactSelectedSeries()` will be called after `CompactOOOHead()`, but this wouldn't help in your usecase. If we receieve an OOO sample while we write the block, that somple will stay orphaned.
So we need to enforce the `s.ooo == nil` check in the `shouldEvict` parameter that is passed to both `truncateSelectedSeries()` and `truncateStaleSeries()`

tsdb/db.go
colega: If the filter isn't used, why do we have it in the argument?

colega: Discussed offline, I've sent https://github.com/prometheus/prometheus/pull/18735 to make this better, since it's taking its time to merge that, I'm not blocking on this part.

duricanikolic: Hi @colega 
I have implemented analogous changes to the one you proposed in https://github.com/prometheus/prometheus/pull/18735.

tsdb/head_read.go
```diff
@@ -309,29 +319,82 @@ func (h *Head) filterStaleSeriesAndSortPostings(p index.Postings) ([]storage.Ser
...
 
-// SortedPostings returns the postings as it is because we expect any postings obtained via
-// headStaleIndexReader to be already sorted.
-func (*headStaleIndexReader) SortedPostings(p index.Postings) index.Postings {
-	// All the postings function above already give the sorted list of postings.
-	return p
+// filterSelectedSeriesAndSortPostings filters the given postings to those whose series exist in
+// the head and carry no out-of-order data, then returns them sorted by series labels. Refs that
+// do not resolve to any series in the head are silently dropped.
+func (h *Head) filterSelectedSeriesAndSortPostings(p index.Postings) (seriesRefs, error) {
```
colega: This method is almost identical to `filterStaleSeriesAndSortPostings` except for the `isStaleSeries` check.

How about we extract one common method from both of them passing in the condition (`s.ooo == nil` in one case, and `s.ooo == nil && isStaleSeries(s)` in the other one)?

colega (dismissed):
LGTM with a nit

tsdb/head_read.go
duricanikolic: Fixed in https://github.com/grafana/mimir-prometheus/pull/1154/commits/96e7d4052f11cd453ce2ee479678f2b6a5ce882e.

tsdb/db.go
duricanikolic: https://github.com/prometheus/prometheus/pull/18735 is merged. I merged the upstream prometheus into mimir-prometheus (https://github.com/grafana/mimir-prometheus/pull/1186), and I merged this PR with the new `main`.

tsdb/head_read.go
```diff
@@ -230,125 +229,142 @@ func (h *headIndexReader) Series(ref storage.SeriesRef, builder *labels.ScratchB
...
+// series refs. Compaction callers only request AllPostings, which is served from the
+// precomputed ref list without re-reading the Head. Other IndexReader methods delegate to
+// the caller-supplied filter, which encodes the compaction flow's notion of "matching series".
+type headSelectedSeriesIndexReader struct {
 	*headIndexReader
-	staleSeriesRefs staleSeriesRefs
+	selectedSeriesRefs seriesRefs
 }
 
 type allStaleSeriesPostings struct{ index.Postings }
```
colega: This should be `allSelectedSeriesPostings` now.

colega (dismissed):
I think this PR is ready to be shipped.

I left a nit, I also understand that the last two commits were upstreamed to Prometheus separately.

tsdb/head_read.go
duricanikolic: Fixed in https://github.com/grafana/mimir-prometheus/pull/1154/commits/ceb7793901c4e275957a6c63d34c09b4293995f8.

colega (approved):

PR #1201, opened 2026-06-09
tsdb: add benchmarks for CompactSelectedSeries and filterSeriesAndSortPostings

duricanikolic:
#### What this PR does

Adds two benchmarks for the `DB.CompactSelectedSeries()` entry point introduced in #1154:

- `BenchmarkCompactSelectedSeries`: measures the end-to-end cost of compacting and evicting a selected subset of head series.
- `BenchmarkFilterSeriesAndSortPostings`: isolates the upfront filter-and-sort step that runs under the compaction lock before block writing begins.

Together, these benchmarks help quantify how the operation scales with the number of selected series and identify where the fixed costs come from.

For a head containing **100k series × 120 samples per series** (single chunk range; benchmark run on an Apple M5 Pro), the cost of compacting N selected series is approximately:

- **Wall time**: `140 ms + 8.7 µs × N`
- **Memory**: `45 MB + 1.9 KB × N`

The model fits well once roughly **10% or more of the head** is selected. Below that point, wall-clock time bottoms out at a fixed cost of about `140 ms`, dominated by `db.reloadBlocks()` and `RebuildSymbolTable()`, which scan the entire head regardless of how many series are compacted. Memory usage, however, continues to decrease roughly in proportion to the number of selected series.

##### Benchmark setup

The benchmark populates a TSDB head with:

- 100,000 series
- 120 samples per series (`DefaultSamplesPerChunk`)
- One chunk range per series

It then compacts different fractions of the head:

| Selected fraction | Selected series |
|---|---:|
| 100% | 100,000 |
| 50% | 50,000 |
| 30% | 30,000 |
| 10% | 10,000 |
| 1% | 1,000 |
| 0.1% | 100 |

##### End-to-end compaction (`BenchmarkCompactSelectedSeries`)

| Selected | Time | Memory |
|---|---:|---:|
| 100,000 | 1,013 ms | 239 MB |
| 50,000 | 571 ms | 140 MB |
| 30,000 | 407 ms | 110 MB |
| 10,000 | 232 ms | 66 MB |
| 1,000 | 152 ms | 47 MB |
| 100 | 140 ms | 45 MB |

The results show two distinct regimes:

- At `30%`-`100%` selectivity, _runtime and memory scale approximately linearly with the number of selected series_.
- Below `10%` selectivity, runtime flattens out around `140 ms`, indicating a fixed-cost floor that is largely independent of selection size.

A simple approximation for the end-to-end operation is:

> Time ≈ 140 ms + 8.7 µs × selected series
>
> Memory ≈ 45 MB + 1.9 KB × selected series

The fit is surprisingly accurate across the high-selectivity range.

##### Filter-and-sort only (`BenchmarkFilterSeriesAndSortPostings`)

| Input refs | Time | Memory |
|---|---:|---:|
| 100,000 | 24.8 ms | 9.8 MB |
| 50,000 | 13.5 ms | 4.7 MB |
| 30,000 | 7.8 ms | 2.8 MB |
| 10,000 | 2.4 ms | 676 KB |
| 1,000 | 0.21 ms | 16 KB |
| 100 | 0.023 ms | 9 KB |

This step scales nearly linearly and remains inexpensive even at the largest input size.

At 100k refs it contributes only ~25 ms out of ~1 second total runtime (~2.4%). At lower selectivities it becomes effectively negligible.

##### Key observations

- `CompactSelectedSeries()` scales predictably with the number of selected series.
- The filter-and-sort phase is not a bottleneck.
- The fixed `~140 ms` floor is dominated by work that is independent of selection size, primarily:
  - `db.reloadBlocks()`
  - `RebuildSymbolTable()`
  - Other whole-head maintenance work
- For callers such as Mimir's block builder, compacting very small batches does not significantly reduce wall-clock cost because the fixed overhead dominates.

##### Caveats

The benchmark is intended to characterize scaling behavior rather than predict exact production latencies.

Notable simplifications:

- The results show here come from a local development machine (Apple M5 Pro).
- No out-of-order data.
- Single chunk range.
- Disk writes benefit from the local filesystem cache.

While absolute timings will vary in production, the observed scaling characteristics should carry over.

### DIFF
--- a/tsdb/compact_test.go
+++ b/tsdb/compact_test.go
@@ -1516,6 +1516,150 @@ func BenchmarkCompactionFromOOOHead(b *testing.B) {
 	}
 }
 
+// setupDBForSelectedSeriesBenchmark opens a fresh DB whose head contains
+// totalSeries series, each with samplesPerSeries in-order samples confined to
+// a single 1s head chunk range. It returns the DB and the head series
+// refs in append order.
+//
+// Auto-compaction is disabled so the benchmark observes exactly the state
+// produced by this helper. The caller owns the returned DB and must close it.
+//
+// The helper is shared by BenchmarkCompactSelectedSeries and the upcoming
+// filterSeriesAndSortPostings benchmark, so it does not assume which entry
+// point will use the setup.
+func setupDBForSelectedSeriesBenchmark(tb testing.TB, totalSeries, samplesPerSeries int) (*DB, []storage.SeriesRef) {
+	require.Positive(tb, samplesPerSeries, "samplesPerSeries must be > 0")
+	require.LessOrEqual(tb, samplesPerSeries, 1000, "samples must fit one head chunk range (1000 ms)")
+
+	opts := DefaultOptions()
+	opts.MinBlockDuration = 1000
+	opts.MaxBlockDuration = 1000
+	db, err := Open(tb.TempDir(), nil, nil, opts, nil)
+	require.NoError(tb, err)
+	db.DisableCompactions()
+
+	const appendBatch = 10_000
+	refs := make([]storage.SeriesRef, 0, totalSeries)
+	app := db.Appender(context.Background())
+	for i := range totalSeries {
+		lbls := labels.FromStrings("__name__", "metric", "instance", strconv.Itoa(i))
+		var ref storage.SeriesRef
+		for ts := int64(0); ts < int64(samplesPerSeries); ts++ {
+			ref, err = app.Append(ref, lbls, ts, float64(ts))
+			require.NoError(tb, err)
+		}
+		refs = append(refs, ref)
+		if (i+1)%appendBatch == 0 {
+			require.NoError(tb, app.Commit())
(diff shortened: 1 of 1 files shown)


tsdb/compact_test.go
```diff
@@ -1516,6 +1516,150 @@ func BenchmarkCompactionFromOOOHead(b *testing.B) {
...
+func pickRefsEvenly(refs []storage.SeriesRef, count int) []storage.SeriesRef {
+	if count >= len(refs) {
+		out := make([]storage.SeriesRef, len(refs))
+		copy(out, refs)
+		return out
+	}
+	out := make([]storage.SeriesRef, 0, count)
+	step := float64(len(refs)) / float64(count)
+	for i := range count {
+		out = append(out, refs[int(float64(i)*step)])
```
colega, line 1574: Really a nit, but I'd select series randomly. This may result in aligning with some of the sharded data structures in the tsdb head.

colega (approved):

PR #1203, opened 2026-06-10
tsdb: add gauge metric for max head chunks m-mapped per series (measurement-only)

krajorama:
My plan is to aggregate these with `max` on Mimir side so it doesn't introduce a new per tenant metric that we'd aggregate away anyway.

## Summary

Cherry-pick of [prometheus/prometheus#18572](https://github.com/prometheus/prometheus/pull/18572) for measurement purposes only. **This will NOT be merged upstream and will be removed once measurements are complete.**

Adds \`prometheus_tsdb_head_chunks_max_mmapped\`, a gauge updated on each \`mmapHeadChunks\` pass to the maximum number of chunks m-mapped in any individual series during the pass.

The primary motivation is to make it possible to detect pathological chunk creation. For example, when OTLP exponential histograms with cumulative temporality are ingested but the sender resets the counter on every sample, each sample causes a chunk cut, causing head chunks to accumulate faster than they are m-mapped. Without this metric such a condition is invisible.

The metric creation has a TODO comment referencing @krajorama so the temporary nature is visible at the source.

## Test plan

- [x] \`go build ./tsdb/...\` passes.
- [x] \`go test -run TestHead_mmapHeadChunks ./tsdb/\` passes (assertions added at all three \`mmapHeadChunks\` call sites verify the gauge value).

```release-notes
NONE
```


aknuds1 (approved):
LGTM

PR #1204, opened 2026-06-10
tsdb: lazily filter ShardedPostings in the Head

colega:

```release-notes
[PERF] TSDB: Filter Head sharded postings lazily to keep series hot in the CPU cache.
```

---

`headIndexReader.ShardedPostings` used to eagerly walk the input postings, look up
every series via `head.series.getByID` and materialize the matching refs into a
slice. Callers then iterate that slice and call `Series()` for each ref, which
performs a second `getByID` for the same series — by then the series is no longer
in the CPU cache.

This returns a lazy `index.Postings` instead, so the shard-membership lookup
happens during `Next()`/`Seek()` and the series stays hot for the `Series()` call
that immediately follows `At()`.

It also splits the shard-membership computation into two implementations:
`lazyPostingsShardPowerOfTwo` replaces the modulo with a bitwise AND when
`shardCount` is a power of two (the common case), and `lazyPostingsShardDecimal`
keeps the modulo for the rest.

The not-found series debug counter and log were removed; the feature is stable and
the log was not actionable.

### DIFF
--- a/tsdb/head_read.go
+++ b/tsdb/head_read.go
@@ -160,36 +160,115 @@ func (h *headIndexReader) SortedPostings(p index.Postings) index.Postings {
 	return index.NewListPostings(ep)
 }
 
-// ShardedPostings implements IndexReader. This function returns an failing postings list if sharding
+// ShardedPostings implements IndexReader. This function returns a failing postings list if sharding
 // has not been enabled in the Head.
+//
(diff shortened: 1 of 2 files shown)


colega:
This approach seems to be better: https://github.com/prometheus/prometheus/pull/18908

PR #1228, opened 2026-06-30
Only return prefix if case-sensitive

chencs:
Mirrors https://github.com/prometheus/prometheus/pull/19045

```release-notes
NONE

```

### DIFF
--- a/model/labels/matcher.go
+++ b/model/labels/matcher.go
@@ -155,7 +155,7 @@ func (m *Matcher) SetMatches() []string {
 // Prefix returns the required prefix of the value to match, if possible.
 // It will be empty if it's an equality matcher or if the prefix can't be determined.
 func (m *Matcher) Prefix() string {
-	if m.re == nil {
+	if m.re == nil || m.re.caseInsensitivePrefix {
 		return ""
 	}
 	return m.re.prefix
--- a/model/labels/matcher_test.go
+++ b/model/labels/matcher_test.go
@@ -190,7 +190,7 @@ func TestPrefix(t *testing.T) {
 		},
 		{
 			matcher:               mustNewMatcher(t, MatchNotRegexp, "(?i)abc.+"),
-			prefix:                "ABC",
+			prefix:                "",
 			caseInsensitivePrefix: true,
 		},
 	} {


francoposa (approved):
Confirmed that it fixes the bug in Mimir as caught by a Claude unit test for it:

```
// TestBucketStore_Series_RegexCharacterClassMatcher documents the store-gateway behaviour
// for exception=~"[nN]one" when the index contains both exception="none" and exception="None".
//
// Current behaviour (BUG): only "None" is returned.
//
// Root cause: Matcher.Prefix() returns "N" (the raw uppercase rune from [nN] compiled as
// OpLiteral('N', FoldCase)). newLazyIntersectingPostingGroup stores that string as the
// LabelValuesOffsets prefix, which is used as a case-sensitive binary-search window.
// The window covers "N..." entries but ends before "n..." entries ('N'=78 < 'n'=110),
// so "none" is never presented to the filter function and the series is silently dropped.
//
// Fix direction: newLazyIntersectingPostingGroup must not forward a case-insensitive
// prefix to LabelValuesOffsets. Either Matcher should expose IsCaseInsensitivePrefix(),
// or the call site should fall back to prefix="" when Prefix() is known to be folded.
func TestBucketStore_Series_RegexCharacterClassMatcher(t *testing.T) {
	bkt := objstore.NewInMemBucket()
	cfg := defaultPrepareStoreConfig(t)
	// prepareTestBlocks splits series[:4] and series[4:] into separate blocks,
	// so we need at least 8 entries. Pad with dummy series.
	cfg.series = []labels.Labels{
		labels.FromStrings("exception", "none"),
		labels.FromStrings("exception", "None"),
		labels.FromStrings("exception", "other"),
		labels.FromStrings("exception", "pad1"),
		labels.FromStrings("exception", "pad2"),
		labels.FromStrings("exception", "pad3"),
		labels.FromStrings("exception", "pad4"),
		labels.FromStrings("exception", "pad5"),
	}
	s := prepareStoreWithTestBlocks(t, bkt, cfg)
	srv := newStoreGatewayTestServer(t, s.store)
	mint, maxt := s.store.TimeRange()

	tests := []struct {
		name    string
		matcher storepb.LabelMatcher
		// wantBug is what the current (buggy) code returns.
		wantBug [][]mimirpb.LabelAdapter
		// wantFixed is what the code should return after the bug is fixed.
		wantFixed [][]mimirpb.LabelAdapter
	}{
		{
			name:    "character class [nN]one should match both none and None",
			matcher: storepb.LabelMatcher{Type: storepb.LabelMatcher_RE, Name: "exception", Value: "[nN]one"},
			// BUG: Prefix()="N" causes a case-sensitive scan; "none" is outside the window.
			wantBug: [][]mimirpb.LabelAdapter{
				{{Name: "exception", Value: "None"}},
			},
			wantFixed: [][]mimirpb.LabelAdapter{
				{{Name: "exception", Value: "None"}},
				{{Name: "exception", Value: "none"}},
			},
		},
		{
			name:    "literal none matches only none",
			matcher: storepb.LabelMatcher{Type: storepb.LabelMatcher_RE, Name: "exception", Value: "none"},
			wantBug: [][]mimirpb.LabelAdapter{
				{{Name: "exception", Value: "none"}},
			},
			wantFixed: [][]mimirpb.LabelAdapter{
				{{Name: "exception", Value: "none"}},
			},
		},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			seriesSet, _, _, _, err := srv.Series(context.Background(), &storepb.SeriesRequest{
				Matchers: []storepb.LabelMatcher{tc.matcher},
				MinTime:  mint,
				MaxTime:  maxt,
			})
			require.NoError(t, err)
			got := make([][]mimirpb.LabelAdapter, len(seriesSet))
			for i, s := range seriesSet {
				got[i] = s.Labels
			}
			assert.Equal(t, tc.wantBug, got, "current (buggy) behaviour")
		})
	}
}
```

PR #1232, opened 2026-07-03
tsdb: advance lastSeriesID past tombstone refs during WAL replay

pr00se:
ref: [#i-2026-07-02-mimir-ops-03-continuous-test-failed](https://grafanalabs.enterprise.slack.com/archives/C0BETLPN56E)

`lastSeriesID` is rebuilt during WAL replay from series records only. A tombstone proves its ref was allocated at some point, even when the series record is no longer in the WAL (full-range tombstones are retained in checkpoints indefinitely, while the deleted series' records are eventually dropped). Without accounting for tombstone refs, the counter can restore below still-tombstoned refs and reissue them to new series — which the retained full-range tombstones then delete on the next replay, dropping their samples as unknown series references. Observed during the above incident in ops as a self-sustaining per-restart data-loss cycle on small tenants.

This advances `lastSeriesID` past any tombstone ref seen during replay, making ref reuse impossible regardless of which records survive checkpointing.

A fuller fix will be implemented upstream, but this will mitigate the issue in the near-term.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tsdb/head_test.go
+++ b/tsdb/head_test.go
@@ -911,6 +911,40 @@ func TestHead_ReadWAL(t *testing.T) {
 	}
 }
 
+func TestHead_ReadWAL_TombstonesAdvanceLastSeriesID(t *testing.T) {
+	entries := []any{
+		[]record.RefSeries{
+			{Ref: 10, Labels: labels.FromStrings("a", "1")},
+		},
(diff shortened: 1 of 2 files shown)


56quarters (approved):
LGTM

PR #1256, opened 2026-08-01
Backport upstream prometheus bug fix for NH append for mimir-release-2.17

dhlaw-canva:
Backport of prometheus/prometheus#16893 to the `mimir-release-2.17` branch.

### What this fixes

The chunk recoder in `tsdb/chunkenc` mis-records a run of two or more consecutive missing buckets.
This is because `expandIntSpansAndBuckets` tags the run with its last index while `adjustForInserts` reads it as the first, so the flushed chunk gains a span with a negative offset. 
The corruption then compounds on each append and then even further escalates through compaction, producing chunks that fail `Validate()` with `span offset is negative`.

This surfaces on native (exponential) histograms ingested via OTLP whose bucket layout flaps without a counter reset, which the Prometheus scrape path avoids because it compacts on parse but the OTLP ingest path does not.

Upstream fix (merged 2025-07-24): prometheus/prometheus#16893. It reached upstream Prometheus v3.6.0 and Mimir picked it up on the 3.0.0 line; this backport carries the same commit onto the 2.17 branch.

### Change

- I've cherry-picked the upstream commit (`abdbed4d6`) containing the above described fix, which merged without conflicts.

### Testing

- I've run `GOTOOLCHAIN=go1.23.0 go test ./tsdb/...` which passes locally post cherry-picking the fix - I'm assuming there's a CI test that will run against this as well

### Context
I've raised an issue for this bug in [grafana/mimir#16273](https://github.com/grafana/mimir/issues/16273) (corrupt native-histogram chunks on the 2.17 LTS line). I thought this would be a good opportunity to learn more about the interrelated grafana reposistories/dependency vending process - so please feel free to let me know if I'm not following the correct process here.

### DIFF
--- a/tsdb/chunkenc/float_histogram.go
+++ b/tsdb/chunkenc/float_histogram.go
@@ -343,7 +343,7 @@ func (a *FloatHistogramAppender) appendable(h *histogram.FloatHistogram) (
 // original deltas to a new set of deltas to match a new span layout that adds
 // buckets, we simply need to generate a list of inserts.
 //
-// Note: Within expandSpansForward we don't have to worry about the changes to the
+// Note: Within expandFloatSpansAndBuckets we don't have to worry about the changes to the
 // spans themselves, thanks to the iterators we get to work with the more useful
 // bucket indices (which of course directly correspond to the buckets we have to
 // adjust).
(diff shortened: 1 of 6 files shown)


dhlaw-canva:
@krajorama if you get a chance sometime this week, could you quickly TAL and let me know if I've missed anything here?

dhlaw-canva:
sorry for the ping again @krajorama - just checking you had a chance to look into this? 

We've raised a related issue against Mimir 3.1.4 ([link](https://github.com/grafana/mimir/issues/16343)) where the resulting series created from this bug cause a go panic in the query path now, which makes this issue even worse for users affected by the above as they have no viable pathway to upgrade to latest right now.

zenador:
Hi, 2.17 is pretty old at this point and we only maintain older releases on a best effort basis, and we encourage users to upgrade to a newer version wherever possible instead. Since this bug is fixed in 3.0.8 which you're on now, do you still need it in 2.17? Separately, this PR isn't a clean backport as it stands, it carries several unrelated commits, likely because the branch was cut from a divergent base. The fix itself cherry-picks cleanly onto the current `mimir-release-2.17` (`git cherry-pick -x abdbed4d6`, touching only the 6 `tsdb/chunkenc` files), so if we had a good reason to proceed we'd first need the PR rebuilt that way.

dhlaw-canva:
@zenador, thanks for following up here!
> Since this bug is fixed in 3.0.8 which you're on now, do you still need it in 2.17? 

We personally don't need it currently, but as it stands if we encountered a bug in 3.0.8 in the near future that requires us to roll back again to a 2.17.x line we'd be stuck - as we need to be on 3.x line to fix the ingest bug, meaning we'd need to choose between:
- rolling back and continue ingesting bad data
- create a fork of 3.0.8 with a fix until such time as the panic issue in the 3.1.x line is fixed on these corrupt histograms.

> this PR isn't a clean backport as it stands

That's my bad, I'll fix that up now

PR #1259, opened 2026-08-11
tsdb: report CommitStats including duplicate samples dropped at commit time

dmatth1:
The TSDB can accept samples at Append and silently drop them as duplicates at Commit, leaving callers unable to account for the drops.

This adds:

- `prometheus_tsdb_duplicate_samples_total{type="float|histogram"}` for duplicate-timestamp drops detected at Append or Commit.
- `CommitStatsReporter` with per-series counts of silent commit-time drops, split into same-value and conflicting-value categories. Append-time errors are excluded from these stats.

This is the fork counterpart of prometheus/prometheus#19711, consumed by grafana/mimir#16347 for discard metrics and cost attribution. Existing sample acceptance and first-value-wins behavior remain unchanged. In-order histogram exact duplicates are counted as duplicates instead of out-of-order rejections.

OOO detection remains limited to the current OOO head chunk. API shape and always-on per-series collection remain open for draft review.

Related to grafana/mimir#15552. Co-authored with @duricanikolic.

Validation: focused V1/V2 duplicate-stat tests and OOO truncation/restart tests.

```release-notes
[ENHANCEMENT] TSDB: Add duplicate-discard metrics and per-series commit statistics.
[BUGFIX] TSDB: Stop counting in-order histogram exact duplicates as out-of-order rejections.
```

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -458,6 +458,7 @@ type headMetrics struct {
 	chunksRemoved             prometheus.Counter
 	gcDuration                prometheus.Summary
 	samplesAppended           *prometheus.CounterVec
+	duplicateSamples          *prometheus.CounterVec
 	outOfOrderSamplesAppended *prometheus.CounterVec
 	outOfBoundSamples         *prometheus.CounterVec
 	outOfOrderSamples         *prometheus.CounterVec
@@ -558,6 +559,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 		samplesAppended: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Name: "prometheus_tsdb_head_samples_appended_total",
 			Help: "Total number of appended samples.",
+		}, []string{"type"}),
+		duplicateSamples: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "prometheus_tsdb_duplicate_samples_total",
+			Help: "Total number of samples dropped because the series already had a sample at the same timestamp, either silently at commit time or with an error at append time. Rejections of synthetic start-timestamp zero samples are not counted.",
 		}, []string{"type"}),
 		outOfOrderSamplesAppended: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Name: "prometheus_tsdb_head_out_of_order_samples_appended_total",
@@ -660,6 +665,7 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			m.walCorruptionsTotal,
 			m.dataTotalReplayDuration,
 			m.samplesAppended,
+			m.duplicateSamples,
 			m.outOfOrderSamplesAppended,
 			m.outOfBoundSamples,
 			m.outOfOrderSamples,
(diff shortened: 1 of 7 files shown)


tsdb/head.go
```diff
@@ -573,6 +574,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
 			Name: "prometheus_tsdb_head_samples_appended_total",
 			Help: "Total number of appended samples.",
 		}, []string{"type"}),
+		duplicateSamples: prometheus.NewCounterVec(prometheus.CounterOpts{
```
duricanikolic, line 577: @dmatth1 Do we trace here the duplicates that used to be handled correctly before this PR? (different value for an already committed sample with the same timestemp)

storage/interface_append.go
```diff
@@ -194,7 +194,8 @@ type AppenderTransaction interface {
 	// Commit submits the collected samples and purges the batch. If Commit
 	// returns a non-nil error, it also rolls back all modifications made in
 	// the appender so far, as Rollback would do. In any case, an Appender
-	// must not be used anymore after Commit has been called.
+	// must not be used anymore after Commit has been called, except for
```
duricanikolic, line 197: Why do we need this change?

tsdb/head.go
```diff
@@ -573,6 +574,10 @@ func newHeadMetrics(h *Head, r prometheus.Registerer) *headMetrics {
...
+			Name: "prometheus_tsdb_duplicate_samples_total",
+			Help: "Total number of samples dropped because the series already had a sample at the same timestamp, either silently at commit time or with an error at append time. Rejections of synthetic start-timestamp zero samples are not counted.",
```
duricanikolic, line 579: I think that this description should be simplified, because it is currently too technical.

```suggestion
			Help: "Total number of samples dropped because the series already had a sample at the same timestamp."
```

tsdb/head_append.go
```diff
@@ -1225,15 +1235,24 @@ type appenderCommitContext struct {
 	// Number of samples out of order but accepted: with ooo enabled and within time window.
 	oooFloatsAccepted    int
 	oooHistogramAccepted int
+	// Number of samples dropped at commit time because the series already had a
```
duricanikolic, line 1238: Is this comment accurate? If I read it correctly, it states that `floatDuplicatesDropped` and `histoDuplicatesDropped` collect only currently non-silent discards, because these discards happen on commit. The other, silent discards, happen on append.

tsdb/head.go
dmatth1: Yep here we do, but in CommitStats no (cause that would be double counting)

tsdb/head.go
dmatth1: Agreed

tsdb/head_append.go
dmatth1: Yeah so these are only tracking commit time discards (in CommitStats) so the comment is technically accurate

storage/interface_append.go
dmatth1: Ah it's because CommitStats needs to be read after a successful Commit. So we are using Appender after Commit now - just read-only

PR #1268, opened 2026-08-13
[mimir-release-2.17] tsdb: harden head-chunk cache lifecycle (#19136)

colega:
Backport of prometheus/prometheus#19136 to `mimir-release-2.17`.

Stacked on the #19134 backport, which is stacked on #1266. Merge those first.

This brings the head-chunk cache introduced by #1266 to the same lifecycle guarantees as `main`: the fingerprint becomes a single comparable key so future fields cannot be forgotten in the comparison, the cache and the OOO reader's fallback buffer are released on `Close`, the `headChunksBufMaxCap` policy is applied when the reader moves to a different series, and the head-and-OOO queriers stop swallowing chunk-reader `Close` errors.

One behaviour change worth noting: `headChunkReader` gains an exported `EnableChunkCache`, so it now satisfies the `chunkCacheToggler` assertion in `selectSeriesSet`/`selectChunkSeriesSet`. Before this, only `HeadAndOOOChunkReader` could ever turn the cache on, so range queries against a head without out-of-order support never used it.

Backport adaptations, since this branch predates two upstream changes:

* `memSeries.headChunkCount` does not exist here (added upstream by prometheus/prometheus#18300), so the replacement collection is pre-sized from `headChunks.len()`. That is an extra list walk, but only on the first collection for a reader or when switching series with an oversized backing array.
* `newTestHeadWithOptions` does not register a cleanup on this branch, so the new tests close the head themselves; `t.Context()` → `context.Background()`.

```release-notes
NONE
```

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -2372,10 +2372,12 @@ func (mc *memChunk) len() (count int) {
 	return count
 }
 
+// collectHeadChunks walks the headChunks linked list once and returns a slice
+// in oldest-first order (matching mmappedChunks ordering).
+// buf must have length 0 but may have non-zero capacity for reuse; the
+// returned slice's tail beyond its length is zeroed, so a reused, shrinking
+// buffer does not pin chunks from a previous, longer collection.
 func collectHeadChunks(head *memChunk, buf []*memChunk) []*memChunk {
-	if head == nil {
-		return buf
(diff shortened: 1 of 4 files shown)


aknuds1 (approved):
LGTM

PR #1282, opened 2026-08-19
fix: correct shared workflow action tag comments

zeitlinger:
## What changed

Update two `send-slack-message` comments to the actual historical tag `send-slack-message-v1.0.0`.

## Why

The bare version comments omit the shared-workflows monorepo component name, preventing Renovate from identifying the correct tag namespace and causing lookup failures. This repository uses the historical hyphenated release tag; the pinned SHA was verified against that exact tag.

## Validation

- Verified `send-slack-message-v1.0.0` resolves to the pinned SHA.
- `actionlint` on changed workflows (pre-existing runner-label/shellcheck/constant-condition diagnostics excluded)
- `git diff --check`

Related to grafana/shared-workflows#2255


zeitlinger:
Could someone take a look at these small workflow action tag comment corrections? Checks are passing.

PR #1288, opened 2026-08-24
ci: bump Go to 1.27

colega:
#### Which issue(s) does the PR fix:

None, but it unblocks CI on every open PR.

`scripts/check-go-mod-version.sh` derives the oldest supported Go version at run time from https://go.dev/dl/?mode=json and asserts that the `test_go_oldest` job pins that same version. Go 1.27.0 was released, so the oldest supported version moved from 1.25 to 1.26, and the check started failing on every pull request:

```
Error: test_go_oldest uses Go 1.25, but should use Go 1.26 (oldest supported version)
make: *** [Makefile:211: check-go-mod-version] Error 1
```

The first commit moves `test_go_oldest` to 1.26 so the check passes. On its own that would leave `test_go_oldest` on the same version as `test_go`, contradicting the "should be N-1 wrt test_go" comment, so the second commit moves the build and test jobs to Go 1.27 (`quay.io/prometheus/golang-builder:1.27-base`, published 2026-08-21) and leaves `test_go_oldest` at 1.26. `.promu.yml` moves with the workflows, as the comment there asks.

Linting on 1.27 then fails inside the standard library, because golangci-lint v2.12.2 bundles a `go/types` older than the generic methods that 1.27 introduced:

```
src/math/rand/v2/rand.go:213:17: method must have no type parameters (typecheck)
```

The third commit bumps golangci-lint to v2.13.1, which added Go 1.27 support and is built with go1.27.0. The `golangci` job reads its version from `make print-golangci-lint-version`, so no workflow change is needed.

Note this puts us ahead of upstream, which is still on Go 1.26 and golangci-lint v2.12.2, so expect conflicts in `merge-upstream-prometheus` until upstream bumps too. Upstream has also since made `test_go_oldest` derive its version dynamically and dropped the assertion from `check-go-mod-version.sh`, which will replace that pin when the merge lands.

`go.mod` is deliberately left at `go 1.25.10`, matching upstream. The check only rejects a `go.mod` version *newer* than the oldest supported one.

```release-notes

NONE

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     container:
       # Whenever the Go version is updated here, .promu.yml
       # should also be updated.
-      image: quay.io/prometheus/golang-builder:1.26-base
+      image: quay.io/prometheus/golang-builder:1.27-base
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -33,7 +33,7 @@ jobs:
     name: More Go tests
     runs-on: ubuntu-latest
     container:
-      image: quay.io/prometheus/golang-builder:1.26-base
+      image: quay.io/prometheus/golang-builder:1.27-base
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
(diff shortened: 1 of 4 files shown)


colega:
Superseded by #1289, which drops the `test_go_oldest` job instead of chasing its Go version pin. mimir-prometheus only needs to test the Go version Mimir builds with, and removing the job also retires the check that broke CI.

PR #1306, opened 2026-09-08
tsdb: retain a series if it received a sample since compaction started

duricanikolic:
#### Description

Both `CompactSelectedSeries` and `CompactStaleHead` write selected series into a block and then evict them from the head. If a sample lands for one of those series after its block is written but before eviction runs, that sample can be silently lost — it isn't in the block, and the series gets removed from the head anyway.

Until now, the only guard against this was an append-ID watermark (`hasAppendIDAbove`), which relies on Prometheus's isolation subsystem tracking per-sample append IDs. When isolation is disabled, that tracking doesn't exist, and the guard silently becomes a no-op — a config-dependent gap with no compile-time or test-time signal that it had disappeared.

This PR replaces it with a fingerprint of each series' in-order head chunk shape (chunk count + current chunk's sample count), snapshotted before the block is written and rechecked right before eviction. Any mismatch means something was appended in between — in flight or already committed — and the series is retained instead of evicted. This works identically whether isolation is enabled or not, since it never reads any isolation-related state. Two other transitions that also need to block eviction — a series turning out-of-order, and (for stale-series compaction specifically) a series becoming un-stale — are already caught for free, because `isSeriesWithoutOOO`/`isStaleSeries` are re-evaluated live at eviction time rather than from a stale snapshot.

With the fingerprint covering everything the watermark did and more, the watermark and `hasAppendIDAbove` are removed from both compaction paths entirely, along with the now-unused `appendIDWatermark` plumbing through `headSeriesEvictor`/`compactHeadViewLocked`.

```release-notes
[BUGFIX] TSDB: Fix a race in selected-series and stale-series compaction where a sample committed after the block was written but before the series was evicted from the head could be silently lost. The fix no longer depends on isolation being enabled.
```

### DIFF
--- a/tsdb/head.go
+++ b/tsdb/head.go
@@ -1380,20 +1380,21 @@ func isStaleSeries(s *memSeries) bool {
 
 // truncateStaleSeries removes the provided series as long as they are still stale and
 // carry no out-of-order data.
-// appendIDWatermark is the lastAppendID captured before the upstream block write. Series that
-// have received samples with greater appendIDs are skipped, because those samples may not be
-// present in the generated block.
-func (h *Head) truncateStaleSeries(seriesRefs []storage.SeriesRef, maxt int64, appendIDWatermark uint64) error {
+// Series that received a sample since their fingerprint was snapshotted are skipped via
+// fingerprintChangedForRef.
+// This covers both fresh samples, which the live isStaleSeries check also catches, and
+// new stale markers, which only the fingerprint detects.
+func (h *Head) truncateStaleSeries(seriesRefs []storage.SeriesRef, maxt int64, fingerprints map[storage.SeriesRef]seriesFingerprint) error {
 	_, err := h.truncateSeries(seriesRefs, maxt, func(s *memSeries) bool {
-		return isSeriesWithoutOOO(s) && isStaleSeries(s) && !hasAppendIDAbove(s, appendIDWatermark)
+		return isSeriesWithoutOOO(s) && isStaleSeries(s) && !fingerprintChangedForRef(s, fingerprints)
 	})
 	return err
 }
 
 // truncateSelectedSeries removes the series identified by the provided refs from the head.
-// Series that received fresh samples or acquired OOO data after the caller collected the ref
-// list are skipped, via fingerprints -- see fingerprintChangedForRef. The latter must be
-// flushed by CompactOOOHead before they can be evicted.
+// Series that received new samples or acquired OOO data after the refs were collected are
+// skipped via fingerprints; see fingerprintChangedForRef. OOO data must first be flushed by
+// CompactOOOHead before the series can be evicted.
 func (h *Head) truncateSelectedSeries(seriesRefs []storage.SeriesRef, maxt int64, fingerprints map[storage.SeriesRef]seriesFingerprint) error {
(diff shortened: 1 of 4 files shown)


tsdb/head.go
```diff
@@ -1380,47 +1380,92 @@ func isStaleSeries(s *memSeries) bool {
...
+// No field tracks OOO transitions: isSeriesWithoutOOO is re-evaluated live at eviction-check
+// time, so a series that turns OOO after being selected is already caught there, for free.
+// Likewise, no field tracks staleness: isStaleSeries is re-evaluated live too.
+type seriesFingerprint struct {
+	headChunkCount   uint32
+	lastChunkSamples int
+}
+
+// snapshotFingerprint captures s's current fingerprint. Must be called with s.Lock held.
+func snapshotFingerprint(s *memSeries) seriesFingerprint {
```
colega, line 1418: Opinionated nit: I'd make this function a receiver of *memSeries, so we would call `s.fingerprint()` which would return a `seriesFingerprint`.

tsdb/head.go
```diff
@@ -1380,47 +1380,92 @@ func isStaleSeries(s *memSeries) bool {
...
+	}
+	if s.headChunks != nil {
+		fp.lastChunkSamples = s.headChunks.chunk.NumSamples()
+	}
+	return fp
+}
+
+// fingerprintChanged reports whether s's current fingerprint no longer matches fp. Must be
+// called with s.Lock held.
+func fingerprintChanged(s *memSeries, fp seriesFingerprint) bool {
```
colega, line 1430: Opinionated nit: if we want to keep this function, I'd make this function a receiver of `memSeries` with a positive name like `memSeries.sameFingerprint(seriesFingerprint) bool`

Then the usage in `shouldEvict` would be just `s.sameFingerprint(fp)`.

However, I'd just get rid of this function and do `s.fingerprint() == fp`, as we can just compare plain structs. It is true that it would be less performant if the number of samples in the last chunk has changed, but that is the corner case we want to protect from, not the happy case, so in the happy case we'll have to retrieve both numbers anyway.

tsdb/head.go
```diff
@@ -1380,47 +1380,92 @@ func isStaleSeries(s *memSeries) bool {
...
-// When isolation is disabled (s.txs == nil), it always returns false;  in that mode,
-// CompactSelectedSeries and CompactStaleHead rely on their existing requirement that
-// no concurrent writes target the affected series.
+// seriesFingerprint is an isolation-independent snapshot of a series' in-order head chunk
+// shape. Comparing two snapshots detects an in-order append in between, whether still in
+// flight or already committed.
+//
+// No field tracks OOO transitions: isSeriesWithoutOOO is re-evaluated live at eviction-check
+// time, so a series that turns OOO after being selected is already caught there, for free.
+// Likewise, no field tracks staleness: isStaleSeries is re-evaluated live too.
```
colega, line 1411: The comment is ok, but comments are fragile, I would consider adding a non-comparable mark here (a float64 NaN) that wouldn't be equal if any of the compared fingerprints has OOO, see this example: https://go.dev/play/p/h4lVrRSiwAC

colega (dismissed):
LGTM, please check my nitpicks, they're not blocking though.

duricanikolic:
Closed in favour of https://github.com/prometheus/prometheus/pull/19666 that has already been merged.

